PR #132, opened 2021-06-28
{#design-sender-query-get_completion_scheduler}: Fix mistakes in `get_completion_scheduler` example

### LINKED ISSUE #122: {#design-sender-query-get_completion_scheduler}: `// completion_sch3 is equivalent to gpu_sched`
In the example in 4.5.1 execution::get_completion_scheduler

// completion_sch3 is equivalent to cpu_sched

I think it should say gpu_sched instead of cpu_sched.

meastp:
I believe the example has a few typos, but could be wrong, of course. :)

fixes #122


brycelelbach (approved):
Looks good. The commit message is a bit too vague, but I'll fix that.

PR #143, opened 2021-07-02
Update the get_stop_token wording.

### LINKED ISSUE #139: {#spec-execution.receiver.queries.get_stop_token}: The wording for `get_stop_token` is incorrectly copy/pasted from the wording for `get_scheduler`
One part of the wording for `get_stop_token` is incorrectly copy/pasted from the wording for `get_scheduler`:

> 1. `execution::get_stop_token` is used to ask a receiver object for an associated stop token to be used by a sender it is connected to when it needs to launch additional work.

The "when it needs to launch additional work" part is wrong, as this has nothing to do with launching work. That exact wording appears in {#spec-execution.receiver.queries.get_scheduler}.

griwes:
Resolves #139

### DIFF
--- a/std_execution.bs
+++ b/std_execution.bs
@@ -1854,7 +1854,8 @@ namespace std::execution {
 
 #### `execution::get_stop_token` <b>[execution.receivers.queries.get_stop_token]</b> #### {#spec-execution.receivers.queries.get_stop_token}
 
-1. `execution::get_stop_token` is used to ask a receiver object for an <i>associated stop token</i> to be used to send stop requests by a sender it is connected to.
+1. `execution::get_stop_token` is used to ask a receiver object for an <i>associated stop token</i> of that receiver. A sender connected with that receiver can use this stop token to check whether a stop request has been made. [<i>Note</i>: such
+    a stop token being signalled does not bind the sender to actually cancel any work. --<i>end note</i>]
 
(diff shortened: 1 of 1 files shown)


std_execution.bs
```diff
@@ -1854,7 +1854,8 @@ namespace std::execution {
 
 #### `execution::get_stop_token` <b>[execution.receivers.queries.get_stop_token]</b> #### {#spec-execution.receivers.queries.get_stop_token}
 
-1. `execution::get_stop_token` is used to ask a receiver object for an <i>associated stop token</i> to be used to send stop requests by a sender it is connected to.
+1. `execution::get_stop_token` is used to ask a receiver object for an <i>associated stop token</i> of that receiver. A sender connected with that receiver can use this stop token to check whether a request to cancel its execution has been raised. [<i>Note</i>: such
```
lewissbaker, line 1857: The wording here is ok, although I'd probably use the term "stop request has been made" instead of "request to cancel its execution has been raised" to be consistent with the wording used for std::stop_token.

The main bit I think is missing here from P2175 is requirements placed on usages of a stop-token returned from `get_stop_token()` with regards to the lifetime of the operation-state produce when connecting that receiver to a sender.
i.e. that the stop-token returned is not guaranteed to remain valid after the call to signal completion of the operation, and also that any stop-callbacks registered using the stop-token must be destroyed/deregistered before signalling completion of that operation.

lewissbaker: I think also, we need to require that customisations provide at least the minimum guarantee of the lifetime of validity of the stop-tokens returned from `get_stop_token` and we need to mention (perhaps in a note) the implication of these minimal guarantees on safe usage of stop-tokens obtained from this CPO.

griwes: I believe that the paragraph 3 that I've just added resolves this.

PR #144, opened 2021-07-02
Add design section of cancellation

lewissbaker:


std_execution.bs
```diff
@@ -553,6 +553,293 @@ merged into a single one in an asynchronous fashion.
...
+
+TODO: Finish describing this
+* stop-requests can be issued concurrently from another thread 
+* usually need to either start the operation or attach a stop-callback
+  * if stop-callback runs before launching the op then may need synchronisation to avoid cancelling the operation before it's started.
+  * if starting the operation befor launching then may need synchronisation to signal when the stop-callback has been subsequently registered so that if the operation completes on another thread it doesn't try to deregister the stop-callback before it's finished registering
+* For example of how this is solved see following async_read() example.
+
+
+### Cancellation Example: Socket async_read() ###
```
griwes, line 698: This example is a big blob of code, and while I think it is helpful, I wonder if it should go into an appendix instead of being right here inline with the rest of the text.

std_execution.bs
```diff
@@ -553,6 +553,293 @@ merged into a single one in an asynchronous fashion.
...
+    timeout(
+      when_all(
+        first_successful(
+          query_server_a(query),
+          query_server_b(query)),
+        load_file("some_file.jpg")),
+      5s),
+    cancelButton.on_click());
+}
+</pre>
```
griwes, line 608: This example creates a precedent for the rest of the paper: I believe that other than the lines that obtain schedulers from some handwaved sources, the rest of the examples in this paper only actually use other facilities in this paper. The only one that does something else is the `async_read` example, but it treats it as a basis operation of sorts, presumably from another library.

This, however, uses at least three things that certainly should be a part of *this* library if we standardize them: `stop_when`, `timeout`, and `first_successful`. I'm somewhat worried about this; I believe that this makes it no longer possible for us to claim that everything in this paper is self-contained, because now we have an example that clearly requires additional sender machinery to even explain what it tries to achieve.

std_execution.bs
```diff
@@ -553,6 +553,293 @@ merged into a single one in an asynchronous fashion.
...
+to ask for an already-started operation to complete early.
+
+The ability to be able to cancel in-flight operations is fundamental to supporting some kinds
+of generic concurrency algorithms.
+
+For example:
+* a generic `timeout(src, duration)` algorithm needs to be able to cancel the `src` operation after the timeout duration has elapsed.
+* a `first_successful(ops...)` algorithm should cancel the other operations as soon as one operation completes successfuly
+* a `when_all(ops...)` algorithm should cancel other operations as soon as one operation fails
+* a `stop_when(src, trigger)` algorithm should cancel `src` if `trigger` completes first and cancel `trigger` if `src` completes first
```
griwes, line 583: See the comment below; I feel that suddenly introducing a number of algorithms that exist solely for the purpose of explaining why another piece of machinery needs to exist goes contrary to what the rest of the paper has tried to do; I'm especially troubled by the fact that `when_all` is buried in the middle of this list.

std_execution.bs
lewissbaker: Yes, that was my intent - to move this to an appendix.

Although it's possible I could trim the example down to some abstract hypothetical OS call.
e.g. `void begin_os_call(int fd, void* buf, size_t len, void(*cb)(std::error_code, size_t, void*), void* data)`
and `void cancel_os_call(int fd, void* data)`, which would make it easier to explain the essentials of the synchronisation needed to coordinate cancellation.

std_execution.bs
lewissbaker: Yes, I think for cancellation to be usable and useful in general, that algorithms like `timeout()`, `first_successful` and `stop_when` should be included in this paper. Although for `timeout()`, this also implies the need for a `time_scheduler` concept that provides the ability to create time-based schedule-senders.

It's not particularly compelling for including cancellation support if the only operation that makes use of cancellation is `when_all`.

std_execution.bs
lewissbaker: These algorithms are not solely for the purpose of explaining why cancellation exists - the intent is to provide algorithms like these in the standard library. That they are not included in this paper is mainly a scoping exercise.

There are few algorithms in this paper that actually have an interaction with cancellation - `when_all()` being the only algorithm that actually introduces a new cancellation-scope. There are a large number of other algorithms not yet included in this paper that do provide more facilities for dealing with cancellation.

std_execution.bs
```diff
@@ -553,6 +553,446 @@ merged into a single one in an asynchronous fashion.
...
+    This is the strategy that `std::jthread` uses for its destructor.
+    It reduces the risk of deadlock but does not eliminate it.
+
+
+Another implication of the use of eager operations with regards to cancellation is that the
+eagerly executing operation will not have access to the caller's stoppable-token until the
+sender is connected to a receiver. If we still want to be able to have the eager operation
+able to be cancelled then the eager operation will need to create a new stop-source and pass
+its associated stop-token down to child operations. Then when the returned sender is eventually
+connected it will
```
ericniebler, line 723: Incomplete.

std_execution.bs
```diff
@@ -553,6 +553,446 @@ merged into a single one in an asynchronous fashion.
...
+P2175R0 contains more details on the background motivation and prior-art and design rationale of this design.
+
+It is important to note, however, that initial review of this design in the SG1 concurrency subgroup raised some concerns
+related to runtime overhead of the design in single-threaded scenarios and these concerns are still being investigated.
+
+The design of P2175R0 has been included in this paper for now, despite its potential to change, as we believe that
+support for cancellation is a fundamental requirement for an async model and is required in some form to be able to
+talk about the semantics of some of the algorithms proposed in this paper.
+
+This paper will be updated in future with any changes that arise from the investigations into P2175R0.
```
ericniebler, line 994:
```suggestion
This paper will be updated in the future with any changes that arise from the investigations into P2175R0.
```

ericniebler, line 994:
```suggestion
This paper will be updated in the future with any changes that arise from the investigations into P2175R0.
```

std_execution.bs
cor3ntin: Put `when_all` first :)

std_execution.bs
cor3ntin: I'm not sure `stop_when` adds to the example. `when_all` and `timeout` seem sufficient

std_execution.bs
```diff
@@ -553,6 +553,446 @@ merged into a single one in an asynchronous fashion.
...
+The design of cancellation described in this paper is built on top of and extends the `std::stop_token`-based
+cancellation facilities added in C++20.
+
+At a high-level, the facilities proposed by this paper for supporting cancellation include:
+* Add `std::stoppable_token` and `std::stoppable_token_for` concepts that generalise the interface of `std::stop_token` type to allow other types with different implementation strategies.
+* Add `std::unstoppable_token` concept for detecting whether a `stoppable_token` can never receive a stop-request.
+* Add `std::in_place_stop_token`, `std::in_place_stop_source` and `std::in_place_stop_callback<CB>` types that provide a more efficient implementation of a stop-token for use in structured concurrency situations.
+* Add `std::never_stop_token` for use in places where you never want to issue a stop-request
+* Add `std::execution::get_stop_token()` CPO for querying the stop-token to use for an operation from its receiver.
+* Add `std::execution::stop_token_type_t<T>` for querying the type of a stop-token returned from `get_stop_token()`
```
cor3ntin, line 639: This being just `decltype(get_stop_token())` right?

std_execution.bs
```diff
@@ -553,6 +553,446 @@ merged into a single one in an asynchronous fashion.
...
+1. Make this undefined-behaviour - the caller must ensure that any eagerly-executing sender is always joined by connecting and starting that sender.
+   This approach is generally pretty hostile to programmers, particularly in the presence of exceptions, and is thus not recommended.
+2. Detach from the computation - let the operation continue in the background - like an implicit call to `std::thread::detach()`.
+   While this approach can work in some circumstances for some kinds of applications, in general it is also pretty user-hostile
+   as it makes it difficult to reason about the safe destruction of resources used by these eager operations. In general, you'll
+   need some kind of garbage collection, e.g. `std::shared_ptr`, to ensure resources are kept alive until the operations complete.
+3. Block in the destructor until the operation completes.
+   This approach is probably the safest to use as it preserves the structured nature of the concurrent operations,
+   but also introduces the potential for deadlocking the application if the completion of the operation depends on
+   the current thread making forward progress.
```
cor3ntin, line 698: Mentioned the history of `jthread` maybe? It seems to be justified the same way. And "we went on that path before" is a good argument imo

std_execution.bs
lewissbaker: It's `std::remove_cvref_t<decltype(get_stop_token())>`

PR #148, opened 2021-07-03
Add missing dots

danra:


std_execution.bs
```diff
@@ -717,7 +717,7 @@ execution::sender auto vals = execution::transfer_just(
     1, 2, 3
 );
 execution::sender auto snd = execution::then(pred, [](auto... args) {
-    std::print(args..);
+    std::print(args...);
```
ericniebler, line 720: What is `std::print` ayway? @griwes?

danra: I imagine a reference to to-be-standardized `fmt::print`

Issue #150, opened 2021-07-06
Add section to compare/contrast senders and coroutines


ericniebler:
... and also discuss under what conditions senders can be made awaitable in coroutines, and the hope we get a `std::task` type that makes senders awaitable.

cor3ntin:
Haha, I was going to create this issue.

So the plan is to propose `operator co_await(awaitable_sender)` in a separate paper?
Folks seem to be under the impression that it's two separate things/models so anything that can show that coroutines can be sugar over senders would be great.

Despite the imperfect fit

LeeHowes:
It seemed hard to add it in here without any of the coroutine library types.

The task type should hook into the same scheduler and cancellation propagation machinery that we have here, so co_await in a task should work on any sender. That seems the right order to approach it in.

ericniebler:
>The task type should hook into the same scheduler and cancellation propagation machinery that we have here, so co_await in a task should work on any sender. That seems the right order to approach it in.

I disagree. For any real solution, we would want that ability to automatically propagate stop tokens and schedulers, but that's not necessary for some edifying code to help people understand the basic idea. _Very_ simple `sender_as_awaitable` and `awaitable_as_sender` that does nothing fancy would go a long way to clearing up the fear and uncertainty. We have to dispel the impression that a vote for sender/receiver is a vote against coroutines.

lewissbaker:
> _Very_ simple `sender_as_awaitable` and `awaitable_as_sender` that does nothing fancy would go a long way to clearing up the fear and uncertainty. We have to dispel the impression that a vote for sender/receiver is a vote against coroutines.

I think any examples we show involving coroutines should just have coroutines magically interoperating with senders. We should not see the identifiers “awaitable_as_sender” or “sender_as_awaitable” in the example code.

ericniebler:
Hrm, maybe, but some might find it instructive to see _how_ these things interoperate.

I've had luck teaching someone about senders but showing them how to implement the `as_future` algorithm.

cor3ntin:
@ericniebler The thing is, that does not illustrate a big difference from the networking TS where you can get it to return an awaitable using `use_awaitable` as a completion token.

Certainly, explaining how it works is super interesting but not *the main point*.

I think it would be useful to present, in order of importance

- This model works with coroutines, out of the box, without having to use a different interface. Senders (a subset thereof) are awaitable. and because algorithms are senders, (some) algorithms are awaitable, etc

[...]

PR #167, opened 2021-08-18
Flesh out the section about why coroutines make a poor basis for generic asynchrony

ericniebler:
In a separate PR, I will add a section discussing the interactions between coroutines and sender/receiver.

refs #150


std_execution.bs
```diff
@@ -292,13 +292,33 @@ essential when writing user code using standard execution concepts; we have also
 
 ## Prior art ## {#intro-prior-art}
 
-This proposal builds upon and learns from years of prior art with asynchronous and parallel programming frameworks in C++.
+This proposal builds upon and learns from years of prior art with asynchronous and parallel programming frameworks in C++. In this section, we discuss async abstractions that have previously been suggested as a possible basis for asynchronous algorithms and why they fall short.
+
+### Futures ### {#intro-prior-art-futures}
+
+A future is a handle to work that has already been scheduled for execution. It is the receiving end of a communication channel. The other end is a promise, used by the concurrent operation to set the result.
```
griwes, line 299: I wonder if it's possible to reword this to not use the word "receiving", because here it has a meaning that's a reverse of the meaning in "receiver".

std_execution.bs
```diff
@@ -292,13 +292,33 @@ essential when writing user code using standard execution concepts; we have also
...
+
+Those advantages come at a cost, though. Because of the dynamic allocation of coroutine frames, coroutines in embedded or heterogeneous environments, which often lack support for dynamic allocation, require great attention to detail. And the allocations and indirections tend to complicate the job of the inliner, often resulting in sub-optimal codegen.
+
+The coroutine language feature mitigates these shortcomings somewhat with the HALO optimization [[P0981R0]], which leverages existing compiler optimizations such as allocation elision and devirtualization to inline the coroutine, completely eliminating the runtime overhead.
+
+However, HALO requires a sophisiticated compiler, and a fair number of stars need to align for the optimization to kick in. In our experience, more often than not in real-world code the compiler is not able to inline the coroutine, resulting in allocations and indirections in the generated code.
```
griwes, line 313: Fold this paragraph into the previous one maybe? It's possible it works better in rendered form, but the paragraph break feels distracting here.

griwes (approved):
Other than the two comments below, LGTM.

PR #169, opened 2021-08-18
remove superfluous tag_invoke poison-pill overload

### LINKED ISSUE #168: tag_invoke poison pill overload is not necessary
It only exists in libunifex because of a bug in ADL in MSVC.

ericniebler:
fixes #168

### DIFF
--- a/std_execution.bs
+++ b/std_execution.bs
@@ -1404,13 +1404,7 @@ Insert this section as a new subclause, between Searchers <b>[func.search]</b> a
 <blockquote>
 
 1. The name `std::tag_invoke` denotes a customization point object. For some subexpressions `tag` and `args...`, `tag_invoke(tag, args...)` is expression-equivalent to an unqualified call to <code>tag_invoke(<i>decay-copy</i>(tag), args...)</code> with overload
-    resolution performed in a context that includes the declaration:
-
-    <pre>
-    void tag_invoke();
-    </pre>
-
-    and that does not include the the `std::tag_invoke` name.
+    resolution performed in a context that does not include the the `std::tag_invoke` name.
 
 </blockquote>
 </ins>


std_execution.bs
```diff
@@ -1404,13 +1404,7 @@ Insert this section as a new subclause, between Searchers <b>[func.search]</b> a
...
 
 1. The name `std::tag_invoke` denotes a customization point object. For some subexpressions `tag` and `args...`, `tag_invoke(tag, args...)` is expression-equivalent to an unqualified call to <code>tag_invoke(<i>decay-copy</i>(tag), args...)</code> with overload
-    resolution performed in a context that includes the declaration:
-
-    <pre>
-    void tag_invoke();
-    </pre>
-
-    and that does not include the the `std::tag_invoke` name.
+    resolution performed in a context that does not include the the `std::tag_invoke` name.
```
lewissbaker, line 1407: Can we instead use some wording along the lines of:
> ... to an unqualified call to <code>tag_invoke(<i>decay-copy</i>(tag), args...)</code> with overload resolution performed only on overloads found by argument-dependent lookup \[\[Note: Overloads of <code>tag_invoke</code> declared in <code>std::</code> or global namespace are not considered unless one or more of the argument types is defined in that respective namespace.\]\]

ericniebler: I don't think that's implementable in standard C++. A tag_invoke overload at global scope can be found by phase 1 lookup if it comes lexically before the unqualified call, regardless of the associated namespaces of the arguments.

lewissbaker: But not if we insert a poison-pill overload, right?

ericniebler: No, all the overloads from phase 1 and phase 2 get thrown into the same bucket, and then overload resolution is performed. If the poison pill overload doesn't get selected (which it won't because it takes 0 arguments), then it's like it's not even there.

lewissbaker: Interesting. I had always thought that name lookup from phase 1 stopped looking for overloads in parent namespaces once it found a name in an inner namespace. i.e. that the poison pill effectively shadowed all overloads of the same name in any ancestor namespaces of the namespace the poison pill was defined in.

ericniebler: Ah, yes that's right. I forgot! There's just too much to remember about ADL. :-D Thanks.

PR #170, opened 2021-08-19
start of a reference implementation and the hello_world example from P2300

ericniebler:
This gets all the CPOs and most of the concepts in place, as well as the `execute`, `lazy_then`, and `sync_wait` algorithms. More to come in future PRs. This should be enough to get started writing examples.


ericniebler:
What sort of license/ copyright notice do we want on these files?

PR #176, opened 2021-08-24
sender_adaptor_closure for defining third-party sender adaptors

### LINKED ISSUE #173: Make it possible for third party sender adaptors to compose with std ones portably
We want `sender | Adapt1 | Adapt2` to be equivalent to `sender | (Adapt1 | Adapt2)`, regardless of whether the two adaptors are standard-provided ones or from a third party.

ericniebler:
fixes #173


griwes (approved):
This is surprisingly concise; I like how it came together.

Separately, we should add something in Design - implementer side that talks about this and provides a trivial example of how to write your own pipeable sender adaptor object.

PR #178, opened 2021-08-24
change specification of sender_traits and connect to make awaitables senders

ericniebler:
refs #150


griwes (approved):
This looks reasonable to me, though we should definitely have Lewis read through this as well.

PR #180, opened 2021-08-26
Add utilities as_awaitable() and with_awaitable_senders<>

ericniebler:
This makes it trivial for coroutine type authors to make senders awaitable within their coroutines. It also provides a default implementation of `unhandled_done()` that makes a done signal behave like an uncatchable exception.


std_execution.bs
```diff
@@ -3022,3 +3039,152 @@ are all well-formed.
...
+        struct <i>awaitable-receiver</i> { // exposition only
+          variant&lt;monostate, result_t, exception_ptr>* result_ptr_;
+          coroutine_handle&lt;P> continuation_;
+          // ... <i>see below</i>
+        };
+        </pre>
+
+        Let `r` be an rvalue expression of type <code><i>awaitable-receiver</i></code>, let `v` be an expression of type `result_t`, let `e` be an expression of type `exception_ptr`, let `p` be an lvalue reference to the coroutine promise type, let `c` be a customization point object, and let `as...` be a pack of arguments. Then:
+
+          - If `value_t` is `void`, then `execution::set_value(r)` is expression-equivalent to  `(result_ptr_->emplace<1>(v), continuation_.resume())`; otherwise, `execution::set_value(r, v)` is expression-equivalent to `(result_ptr_->emplace<1>(v), continuation_.resume())`.
```
ericniebler, line 3112: Lol, the "if" and the "otherwise" cases are the same. 😳

std_execution.bs
```diff
@@ -3022,3 +3039,152 @@ are all well-formed.
...
+
+          - `execution::set_error(r, e)` is expression-equivalent to `(result_ptr_->emplace<2>(e), continuation_.resume())`.
+
+          - `execution::set_done(r)` is expression-equivalent to `continuation_.promise().unhandled_done().resume()`.
+
+          - `std::tag_invoke(c, r, as...)` is expression-equivalent to `c(p, as...)`.
```
ericniebler, line 3118: ... when c is not a receiver set_* cpo.

std_execution.bs
```diff
@@ -2180,7 +2195,9 @@ enum class forward_progress_guarantee {
 
           - <code><i>operation-state-task</i></code> is a type that models `operation_state`. Its `execution::start` resumes the <code><i>connect-awaitable</i></code> coroutine, advancing it past the initial suspend point.
 
-          - Let `p` be an lvalue reference to the promise of the <code><i>connect-awaitable</i></code> coroutine, let `b` be an lvalue reference to the receiver `r`, and let `c` be any customization point object excluding those of type `set_value_t`, `set_error_t` and `set_done_t`. Then `std::tag_invoke(c, p, as...)` is expression-equivalent to `std::tag_invoke(c, b, as...)` for any set of arguments `as...`.
+          - Let `p` be an lvalue reference to the promise of the <code><i>connect-awaitable</i></code> coroutine, let `b` be an lvalue reference to the receiver `r`, and let `c` be any customization point object excluding those of type `set_value_t`, `set_error_t` and `set_done_t`. Then `std::tag_invoke(c, p, as...)` is expression-equivalent to `c(b, as...)` for any set of arguments `as...`.
```
lewissbaker, line 2198: Do we want to limit this to CPOs called on const-reference-to-receiver?

std_execution.bs
```diff
@@ -2180,7 +2195,9 @@ enum class forward_progress_guarantee {
...
+
+          - The expression `p.unhandled_done()` is expression-equivalent to `(execution::set_done((R&&) r), noop_coroutine())`.
```
lewissbaker, line 2200: `set_done()` doesn't take any extra args.
Why is the `noop_coroutine()` being passed here?

lewissbaker: Never mind. I just noticed it's a comma-operator, not an extra arg....

std_execution.bs
```diff
@@ -3022,3 +3039,152 @@ are all well-formed.
...
+
+## Coroutine utilities <b>[execution.coro_utils]</b> ## {#spec-execution.coro_utils}
+
+### `execution::as_awaitable` <b>[execution.coro_utils.as_awaitable]</b> ### {#spec-execution.coro_utils.as_awaitable}
+
+1. `as_awaitable` is used to wrap a sender into an object that is awaitable within a particular coroutine. A sender `s` of type `S` can be made awaitable if and only if:
+
+    - `S` models the concept `typed_sender`, and
+
+    - It passes zero or one values of a unique type to `execution::set_value`, and
```
lewissbaker, line 3051: Wording here is a little unclear. How do I pass zero values of a unique type? What is a unique type? What makes it unique?

Maybe change to something like:
- If the operation completes with `set_value()` then it calls at most one overload of `set_value()` with either
  zero or one value arguments.

Even then, is the fact that an operation completes with 1 type already covered by the `single-typed-sender` requirement below? Can we just say that `S` models `single-typed-sender` instead?

std_execution.bs
```diff
@@ -3022,3 +3039,152 @@ are all well-formed.
...
+
+    - It passes only objects of type `exception_ptr` to `exception::set_error`, and
+
+    - The promise type of the awaiting coroutine implements a function called `unhandled_done` that takes no arguments and returns a `coroutine_handle<>`.
```
lewissbaker, line 3055: Does it _have_ to return a `coroutine_handle<void>` or does it just need to be convertible to `coroutine_handle<void>`?

lewissbaker: Also - do we need to specify that this is a _member_ function?
Or just that the expression `promise.unhandled_done()` is well-formed and returns `coroutine_handle<P>` for some type `P`?
Do we need to say anything about whether such an expression is required to be noexcept?

std_execution.bs
```diff
@@ -3022,3 +3039,152 @@ are all well-formed.
...
+
+    template&lt;<i>single-typed-sender</i> Sender, class Promise>
+      <i>unspecified</i> as_awaitable(Sender&& s, Promise& p);
+    </pre>
+
+2. Alias template <i>single-sender-value-type</i> is defined as follows:
+
+    - If `sender_traits<remove_cvref_t<S>>::value_types<Tuple, Variant>` would have the form `Variant<Tuple<T>>`, then <code><i>single-sender-value-type</i>&lt;S></code> is an alias for type `T`.
+
+    - Otherwise, if `sender_traits<remove_cvref_t<S>>::value_types<Tuple, Variant>` would have the form `Variant<Tuple<>>` or `Variant<>`, then <code><i>single-sender-value-type</i>&lt;S></code> is an alias for type `void`.
```
lewissbaker, line 3076: I wonder if the result is `Variant<>` whether we should specify that the `await_resume()` method of the returned awaitable will be tagged as `[[noreturn]]`?

std_execution.bs
```diff
@@ -3022,3 +3039,152 @@ are all well-formed.
...
+      struct <i>awaitable-receiver</i>; // exposition only
+
+      variant&lt;monostate, result_t, exception_ptr> result_{}; // exposition only
+      connect_result_t&lt;S, <i>awaitable-receiver</i>> state_;  // exposition only
+
+     public:
+      <i>sender-awaitable</i>(S&& s, P& p);
+      bool await_ready() const noexcept { return false; }
+      void await_suspend(coro::coroutine_handle&lt;>) noexcept { start(state_); }
+      value_t await_resume();
```
lewissbaker, line 3096: If `value_types<Tuple, Variant>` has form `Variant<>` then maybe this should be marked `[[nodiscard]]`?

And if `error_types<Variant>` has form `Variant<>` then maybe this should be marked `noexcept`?

std_execution.bs
```diff
@@ -3022,3 +3039,152 @@ are all well-formed.
...
+      using result_t = conditional_t&lt;is_void_v&lt;value_t>, unit, value_t>;
+      struct <i>awaitable-receiver</i>; // exposition only
+
+      variant&lt;monostate, result_t, exception_ptr> result_{}; // exposition only
+      connect_result_t&lt;S, <i>awaitable-receiver</i>> state_;  // exposition only
+
+     public:
+      <i>sender-awaitable</i>(S&& s, P& p);
+      bool await_ready() const noexcept { return false; }
+      void await_suspend(coro::coroutine_handle&lt;>) noexcept { start(state_); }
```
lewissbaker, line 3095: I am still a little concerned about the potential for unbounded recursion and stack-overflow if someone awaits a sender that completes synchronously in a loop with this formulation.

Maybe this should conditionally schedule itself onto a trampoline_scheduler if the operation may potentially complete synchronously (e.g. if the `get_blocking()` CPO from P2257 was adopted and calling it returned something other than `asynchronous_completion_t`).

std_execution.bs
lewissbaker: I think you need to drop the `v` parameter to `emplace<1>(v)` in the "If `value_t` is `void`" case.

std_execution.bs
lewissbaker: This method should take a `coroutine_handle<P>` to force that the coroutine-type it is being awaited in has exactly type `P`.

std_execution.bs
```diff
@@ -3022,3 +3039,152 @@ are all well-formed.
...
+        <i>see-below</i> await_transform(Value&& value);
+
+       private:
+        coroutine_handle&lt;> continuation_{}; // exposition only
+        coroutine_handle&lt;> (*done_handler_)(void*) noexcept = // exposition only
+          [](void*) noexcept -> coroutine_handler&lt;> { terminate(); };
+      };
+    </pre>
+
+  2. <b>`void set_continuation(coroutine_handle<OtherPromise> h) noexcept`</b>
```
lewissbaker, line 3165: Do we need to specify that this method is only allowed to be called once?
Otherwise it may be possible to call it once with a coroutine_handle<P1> where P1::unhandled_done() is defined and a subsequent time with coroutine_handle<P2> where P2::unhandled_done() is not defined.
In this case you'd be left with the previous coroutine's done_handler_ as defined here.
Maybe it should always be assigned back to a "default_unhandled_done" function if the promise doesn't support it?

std_execution.bs
```diff
@@ -3022,3 +3039,152 @@ are all well-formed.
...
+
+      - <i>Effects:</i> equivalent to:
+
+        <pre highlight="c++">
+        continuation_ = h;
+        if constexpr ( requires(OtherPromise& other) { other.unhandled_done(); } ) {
```
lewissbaker, line 3171: Requires also that this `other.unhandled_done()` method returns something that is `convertible_to<coroutine_handle<>>`?

std_execution.bs
```diff
@@ -3022,3 +3039,152 @@ are all well-formed.
...
+          };
+        }
+        </pre>
+
+  3. <b>`decltype(auto) await_transform(Value&& value)`</b>
+
+      - <i>Effects:</i> equivalent to:
+
+        <pre highlight="c++">
+        if constexpr (<i>is-awaitable</i>&lt;Value>)
```
lewissbaker, line 3184: Normally we'd expect to have a call to the `await_transform()` customisation point here.
Is it guaranteed that we'll be able to add such a customisation in here later without breaking anything?
What's the rationale for not adding in the `await_transform()` CPO at this point?

std_execution.bs
```diff
@@ -3022,3 +3039,152 @@ are all well-formed.
...
+          return (Value&&) value;
+        else if constexpr (<i>single-typed-sender</i>&lt;Value>)
+          return as_awaitable(static_cast&lt;Value&&>(value), static_cast&lt;Promise&>(*this));
+        else
+          return (Value&&) value;
```
lewissbaker, line 3189: This case would never expect to be reached in a well-formed program, right?
The `is-awaitable` concept check should in theory catch and cover all cases where the type should be awaitable.
If we get to the point here where we are returning value unchanged but it is not awaitable then the program is going to be ill-formed anyway...

Having said that, I wonder if we should be doing a stricter check than `is-awaitable` above - we do know the promise_type at this point so we can correctly construct the appropriate `coroutine_handle<P>` argument to `await_suspend()`.

std_execution.bs
ericniebler: Actually, no. If `value_t` is `void`, then `result_t` is `unit`, and you _do_ want to emplace a `unit` object in the variant.

lewissbaker: Yes, but `unit` is default constructible so `emplace<1>()` should be just fine.
Also, there is no definition of `v` for the `void` branch of this logic.

std_execution.bs
ericniebler: I'm not going to sweat this for now.

std_execution.bs
ericniebler: Deferring this.

std_execution.bs
ericniebler: > I am still a little concerned about the potential for unbounded recursion and stack-overflow if someone awaits a sender that completes synchronously in a loop with this formulation.

This has to be future work. I can't open that can of worms here.

std_execution.bs
ericniebler: Unsure. If someone messes up when defining `unhandled_done()`, if I add that requirement then their code compiles but terminates instead of propagating a done signal. If I don't add the requirement, they get a hard error. I prefer a hard error in this case.

std_execution.bs
ericniebler: The rationale is that an `await_transform` CPO should really be its own paper. I worry that putting it in P2300 will raise eyebrows. This is the smallest change that works, and at this point I think the Committee will be scared off by anything other than small changes.

std_execution.bs
ericniebler: > The `is-awaitable` concept check should in theory catch and cover all cases where the type should be awaitable.

True. The question is one of diagnostics. By returning the non-awaitable thing unchanged, the user will get a compiler error telling them that they are trying to await a thingy that isn't awaitable. That's a much better usability story.

> we do know the promise_type at this point so we can correctly construct the appropriate `coroutine_handle<P>` argument to `await_suspend()`.

That's a fair point.

std_execution.bs
ericniebler: > If `value_types<Tuple, Variant>` has form `Variant<>` then maybe this should be marked `[[nodiscard]]`?

@lewissbaker, wouldn't that mean that this awaitable will never produce a value, and so `await_resume` will never be called? What would be the point of declaring it `[[nodiscard]]`?

> And if `error_types<Variant>` has form `Variant<>` then maybe this should be marked `noexcept`?

Filed as #213

std_execution.bs
ericniebler: > I wonder if we should be doing a stricter check than `is-awaitable above` - we do know the promise_type at this point so we can correctly construct the appropriate `coroutine_handle<P>` argument to `await_suspend()`.

Filed as #214

ericniebler (commented):
Following up on outstanding issues here, and a question for @lewissbaker

PR #181, opened 2021-08-27
Add subsections to the "Design - User side" section describing sender/awaitable interop

ericniebler:
fixes #167


std_execution.bs
```diff
@@ -669,6 +669,58 @@ Sender consumers could be made pipeable, but we have chosen to not do so.
 However, since these are terminal nodes in a pipeline and nothing can be piped after them, we believe a pipe syntax may be confusing as well as unnecessary, as consumers cannot be chained.
 We believe sender consumers read better with function call syntax.
 
+## All awaitables are senders ## {#design-awaitables-are-senders}
+
+Since C++20 added coroutines to the standard, we expect that coroutines and awaitables will be how a great many will choose to express their asynchronous code. However, in this paper, we are proposing to add a suite of asynchronous algorithms that accept senders, not awaitables. One might wonder whether and how these algorithms will be accessible to those who choose coroutines instead of senders.
+
+In truth there will be no problem because all awaitable types automatically model the `typed_sender` concept. The adaptation is transparent and happens in the sender customization points, which are aware of awaitables.
+
```
lewissbaker, line 677: > "all awaitable types"

You need to be careful and precise what you mean when you say "awaitable" here.
I found there was much confusion about the subtle nuances in what we mean when we say "awaitable".

There can be awaitable types that are only awaitable within certain contexts.
e.g. many of the folly::coro awaitable types extract some information from the awaiting coroutine's promise via the coroutine_handle passed to await_suspend(). You can't just pass any old coroutine_handle to such an awaitable and expect it to work here.

In the [P1288](https://wg21.link/P1288) paper the `awaitable` concept required that `await_suspend()` was invocable with `coroutine_handle<void>` as a proxy for accepting any `coroutine_handle` and thus being awaitable within any natural coroutine-context (i.e. without `await_transform`).

std_execution.bs
```diff
@@ -669,6 +669,58 @@ Sender consumers could be made pipeable, but we have chosen to not do so.
...
+```c++
+task<int> doSomeAsyncWork();
+
+int main() {
+  // OK, awaitable types satisfy the requirements for typed senders:
+  auto o = this_thread::sync_wait(doSomeAsyncWork());
+}
+```
+
+Since awaitables are senders, writing a sender-based asynchronous algorithm is trivial if you have a coroutine task type: implement the algorithm as a coroutine. If you are not bothered by the possibility of allocations and indirections as a result of using coroutines, then there is no need to ever write a sender, a receiver, or an operation state.
```
lewissbaker, line 689: > there is no need to ever write a sender

This comes with the proviso that there are sufficient concurrency building-blocks available that you can use.
Not all algorithms can be implemented purely with coroutines. Having said that, most algorithms can be written using coroutines once you have something like `async_scope` that can spawn operations that execute concurrently - allowing structured concurrency algorithms to be built out of the ad-hoc concurrency mechanisms.

std_execution.bs
```diff
@@ -669,6 +669,58 @@ Sender consumers could be made pipeable, but we have chosen to not do so.
...
+## Many senders can be trivially made awaitable ## {#design-senders-are-awaitable}
+
+If you choose to implement your sender-based algorithms as coroutines, you'll run into the issue of how to retrieve results from a passed-in sender. This is not a problem. If the coroutine type opts in to sender support -- trivial with the `execution::with_awaitable_senders` utility -- then a large class of senders are transparently awaitable from within the coroutine.
+
+For example, consider the following trivial implementation of the sender-based `retry` algorithm:
+
+<pre highlight="c++">
+template &lt;class S>
+  requires <i>single-typed-sender</i>&lt;S&> // See <a href="#spec-execution.coro_utils.as_awaitable">[execution.coro_utils.as_awaitable]</a>
+task&lt;<i>single-sender-value-type</i>&lt;S>> retry(S& s) {
```
lewissbaker, line 700: Should this be pass-by-value to avoid the dangling-reference problems?

This algorithm would be difficult to use in a sender composition expression as it won't accept temporary senders as arguments.

std_execution.bs
```diff
@@ -669,6 +669,58 @@ Sender consumers could be made pipeable, but we have chosen to not do so.
...
+
+When looking at the sender-based `retry` algorithm in the previous section, we can see that the value and error cases are correctly handled. But what about cancellation? What happens to a coroutine that is suspended awaiting a sender that completes by calling `execution::set_done`?
+
+When your task type's promise inherits from `with_awaitable_senders`, what happens is this: the coroutine behaves as if an *uncatchable exception* had been thrown from the `co_await` expression. (It is not really an exception, but it's helpful to think of it that way.) Provided that the promise types of the calling coroutines also inherit from `with_awaitable_senders`, or more generally implement a member function called `unhandled_done`, the exception unwinds the chain of coroutines as if an exception were thrown except that it bypasses `catch(...)` clauses.
+
+In order to "catch" this uncatchable done exception, one of the calling coroutines in the stack would have to await a sender that maps the done channel into either a value or an error. That is achievable with either the `execution::let_done` or `execution::upon_done` sender adaptors.
```
lewissbaker, line 718: I'm not sure that we should be recommending `let_done` here.
It's use within a coroutine is complicated as having the `let_done()` return a `just(std::nullopt_t)` will result in the overall sender having multiple result-types and then needing to be piped through `into_variant`.

It would be nicer if we could describe an `done_as_optional` algorithm and show an example of its use...
```c++
std::optional<T> result = co_await done_as_optional(some_operation(args...));
if (!result) {
  // completed with set_done()
} else {
  // completed with set_value()
  auto& value = *result;
}
```

Whereas doing the same thing with `let_done()` is more cumbersome:
```c++
std::optional<T> result = co_await transform(
  let_done(some_operation(args...), []() noexcept { return just(std::nullopt); }),
  [](auto&& value) -> std::optional<T> { return static_cast<decltype(value)>(value); });
```

Maybe also throw in an example of using `done_as_error<E>()`?
```c++
try {
  T result = co_await done_as_error<operation_cancelled>(some_operation(args...));
} catch (operation_cancelled) {
  ...
}
```
Or alternatively show that you can just let this exception propagate up through the call-stack of coroutines...

PR #185, opened 2021-08-31
add stop token bits, receiver queries, a manual event loop, and a single-thread context

ericniebler:


examples/schedulers/manual_event_loop.hpp
```diff
@@ -0,0 +1,163 @@
...
+#include <mutex>
+#include <type_traits>
+
+namespace example {
+  struct manual_event_loop;
+
+  namespace detail {
+    struct task_base {
+      void execute() noexcept {
+        this->execute_(this);
```
kirkshoop, line 32: `execute_` does not match the spelling `_execute` on line 41

examples/schedulers/manual_event_loop.hpp
```diff
@@ -0,0 +1,163 @@
...
+      if (head_ == nullptr) {
+        tail_ = nullptr;
+      }
+      lock.unlock();
+      task->execute();
+      lock.lock();
+    }
+  }
+
+  inline void manual_event_loop::stop() {
```
kirkshoop, line 146: probably no change from libunifex, but stop doesn't appear to do the right thing. I don't see the code in run that loops through all the tasks invoking set_done()

kirkshoop (approved):

examples/schedulers/manual_event_loop.hpp
ericniebler: It's not referring to that function. It's referring to the `task_base::execute_` data member on line 36.

examples/schedulers/manual_event_loop.hpp
ericniebler: It doesn't. It only stops when the queue has drained.

kirkshoop: It is a bit confusing because stop means cancel elsewhere and it means exit run when empty here.

ericniebler: Perhaps I should call it `request_stop()`?

PR #188, opened 2021-09-02
Clarify invocable constraints in pseudo-signatures in the _Design - user side_ section

ericniebler:


ericniebler:
Closing stale, unnecessary PR.

PR #198, opened 2021-09-30
Fix underspec'd op state lifetimes, remove spurios exception catching.

### LINKED ISSUE #196: schedule_from algorithm specification leaves dangling operation state
The specification of the `schedule_from` algorithm reads as follows:

> When a receiver completion-signal `Signal(r, args...)` is called, it constructs a receiver `r2` [...]
> [...]
> It then calls `execution::schedule(sch)`, resulting in a sender `s3`. It then calls `execution::connect(s3, r2)`, resulting in an operation state `op_state3`. It then calls `execution::start(op_state3)`.

Nowhere does it say what becomes of `op_state3`. It needs to be stored somewhere until the operation completes.

griwes:
Fixes #196

### DIFF
--- a/std_execution.bs
+++ b/std_execution.bs
@@ -2637,8 +2637,9 @@ are all well-formed.
 
             4. When `execution::set_done(r)` is called, it calls `execution::set_done(out_r)`.
 
-        2. Calls `execution::schedule(sch)`, which results in `s3`. It then calls `execution::connect(s3, r)`, resulting in `op_state3`, and then it calls `execution::start(op_state3)`. If any of these throws an exception, it catches it and calls
-            `execution::set_error(out_r, current_exception())`.
+        2. Calls `execution::schedule(sch)`, which results in `s3`. It then calls `execution::connect(s3, r)`, resulting in `op_state3`, and then it calls `execution::start(op_state3)`.
+
+        3. The lifetimes of `op_state2` and `op_state3`, once constructed, last until `op_state` is destroyed.
 
 4. Any receiver `r` created by an implementation of `on` shall implement the `get_scheduler` receiver query. The scheduler returned from the query for all such receivers should be equivalent to the `sch` argument passed into the `on` call.
 
@@ -2674,7 +2675,7 @@ are all well-formed.
 
         1. Constructs a receiver `r`.
 
-        2. Calls `execution::connect(s, r)`, which results in an operation state `op_state2`. If any of these throws an exception, calls `execution::set_error` on `out_r`, passing `current_exception()` as the second argument.
+        2. Calls `execution::connect(s, r)`, which results in an operation state `op_state2`.
 
         3. When a receiver completion-signal <code><i>Signal</i>(r, args...)</code> is called, it constructs a receiver `r2`:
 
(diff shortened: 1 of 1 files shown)


std_execution.bs
```diff
@@ -3033,7 +3032,7 @@ are all well-formed.
 
         1. Constructs a receiver `r`.
 
-        2. Calls `execution::connect(s, r)`, resulting in operation state `op_state`, and then calls `execution::start(op_state)`. If any of these throws an exception, it catches it and calls `execution::set_error(r, current_exception())`.
+        2. Calls `execution::connect(s, r)`, resulting in operation state `op_state`, and then calls `execution::start(op_state)`.
 
         3. Constructs a sender `s2`. When `s2` is connected with some receiver `out_r`, it results in an operation state `op_state2`. Once both `execution::start(op_state2)` and one of the receiver completion-signals has been called on `r`:
```
ericniebler, line 3037: I'm a bit confused by this. What is `s2`? What is `out_r`? Why are they needed? As an implementer, I wouldn't know how to implement this.

std_execution.bs
```diff
@@ -3043,6 +3042,8 @@ are all well-formed.
 
             3. If `execution::set_done(r)` has been called, calls `execution::set_done(out_r)`.
 
+            The lifetime of `op_state` ends when both `s2` and `op_state2` are destroyed.
```
ericniebler, line 3045: You typically do not need to keep a sender around after it has been connected to a receiver. You only need to keep the op state at that point. The op state should maintain all the state that is needed for the operation to complete.

griwes: Yes, but in this case the shared state is also owned by the sender.

This isn't saying "you need to keep both to keep it alive"; it is saying (or at least trying to say) "either of these will keep it alive.

Do you have an idea how to reword this to make it clearer?

std_execution.bs
```diff
@@ -2636,8 +2636,7 @@ are all well-formed.
 
             4. When `execution::set_done(r)` is called, it calls `execution::set_done(out_r)`.
 
-        2. Calls `execution::schedule(sch)`, which results in `s3`. It then calls `execution::connect(s3, r)`, resulting in `op_state3`, and then it calls `execution::start(op_state3)`. If any of these throws an exception, it catches it and calls
-            `execution::set_error(out_r, current_exception())`.
+        2. Calls `execution::schedule(sch)`, which results in `s3`. It then calls `execution::connect(s3, r)`, resulting in `op_state3`, and then it calls `execution::start(op_state3)`.
```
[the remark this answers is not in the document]
griwes: Oh, I fixed the other bug here, but not the main one, didn't I...

std_execution.bs
griwes: `s2` is the sender that is returned from `ensure_started`; `out_r` is the receiver that a user connects to it.

std_execution.bs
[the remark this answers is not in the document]
griwes: See the new line of diff below, hopefully that's sufficient.

std_execution.bs
griwes: Actually, this is still insufficient...

std_execution.bs
ericniebler: Instead of "Constructs a sender `s2`..." does it make sense to say "Returns a sender `s2`..."? I think that might have avoided my confusion.

std_execution.bs
griwes: @ericniebler please see the latest update around this line, I think this is what's actually needed.

std_execution.bs
griwes: I am unsure how well that meshes with the "expression-equivalent" phrasing we use throughout the paper. We also use "constructs a sender" for the other algorithms, though it's possible that they are less confusing, because it occurs immediately after the "otherwise," in the top level list of expression to try?

PR #204, opened 2021-10-01
add just_error, just_done, done_as_optional and done_as_exception

ericniebler:


std_execution.bs
```diff
@@ -999,18 +1001,16 @@ When looking at the sender-based `retry` algorithm in the previous section, we c
...
 
-In order to "catch" this uncatchable done exception, one of the calling coroutines in the stack would have to await a sender that maps the done channel into either a value or an error. That is achievable with either the `execution::let_done` or `execution::upon_done` sender adaptors, or with a custom `done_as_optional` algorithm. `done_as_optional()` would map a sender of `T` into a sender of `optional<T>`, where the done signal is mapped to a value of `std::nullopt`, which could be used in a coroutine as follows to handle a sender that is cancelled:
+In order to "catch" this uncatchable done exception, one of the calling coroutines in the stack would have to await a sender that maps the done channel into either a value or an error. That is achievable with the `execution::let_done`, `execution::upon_done`, `execution::done_as_optional`, or `execution::done_as_exception` sender adaptors. For instance, we can use `execution::done_as_optional` as below to "catch" the done signal and map it into an empty optional as shown below:
```
griwes, line 1004: "as below" is repeated in this final sentence here.

std_execution.bs
```diff
@@ -1091,13 +1091,31 @@ execution::sender auto vals = execution::transfer_just(
     1, 2, 3
 );
 execution::sender auto snd = execution::then(vals, [](auto... args) {
-    std::print(args...);
+    }    std::print(args...);
```
griwes: typo?

std_execution.bs
```diff
@@ -1209,13 +1227,34 @@ Returns a sender which, when started, will start the provided sender on an execu
...
 
+### `execution::done_as_optional` ### {#design-sender-adaptor-done_as_optional}
+
+<pre highlight=c++>
+execution::sender auto done_as_optional(
+    execution::sender auto snd
+);
+</pre>
+
+Returns a sender that maps the value channel from a `T` to an `optional<T>`, and maps the done channel to a value of an empty `optional<T>`.
```
griwes, line 1245: Should we mention that this requires a single-type sender here?

griwes (approved):
Minor nitpicks, otherwise good to go.

std_execution.bs
```diff
@@ -1098,6 +1098,24 @@ execution::sender auto snd = execution::then(vals, [](auto... args) {
...
 
+### `execution::just_error` ### {#design-sender-factory-just_error}
+
+<pre highlight="c++">
+execution::sender auto just_error(
+    auto && error
+);
+</pre>
+
+Returns a sender with no [=completion scheduler|completion schedulers=], which completes with the specified error. If the provided error is an lvalue reference, a copy is made inside the returned sender and a non-const lvalue reference to the copy is sent to the receiver's `set_error`. If the provided value is an rvalue reference, it is moved into the returned sender and an rvalue reference to it is sent to the receiver's `set_error`.
```
lewissbaker, line 1109: I don't understand why we'd pass an lvalue-reference if we have already copied the value.

Could this be simplified to just say that `just_error(e)` decay-copies `e` and sends an rvalue-reference to the copied value when started?

std_execution.bs
```diff
@@ -1098,6 +1098,24 @@ execution::sender auto snd = execution::then(vals, [](auto... args) {
...
+
+### `execution::just_done` ### {#design-sender-factory-just_done}
+
+<pre highlight="c++">
+execution::sender auto just_done();
```
lewissbaker, line 1114: Do we want to define this as a type named `just_done` or a function named `just_done()`?

kirkshoop:
This gives me pause. But don't wait for me to figure out why.

std_execution.bs
lewissbaker: Can this also elaborate on what it does when `T` is a reference-type?
Does it decay-copy `T` into an `optional<remove_cvref_t<T>>` or does it wrap references into `optional<reference_wrapper<T>>`?

std_execution.bs
```diff
@@ -2563,6 +2611,100 @@ enum class forward_progress_guarantee {
...
+    template&lt;class T>
+    struct <i>just-error-sender</i> // exposition only
+    {
+      T err_;
+
+      template&lt;template&lt;class...> class Tuple, template&lt;class...> class Variant>
+      using value_types = Variant&lt;>;
+
+      template&lt;template&lt;class...> class Variant>
+      using error_types = Variant&lt;T>;
```
lewissbaker, line 2628: Shouldn't this also include `std::exception_ptr` since you are handling a throwing `execution::set_value()` call below by sending `std::current_exception()`?

std_execution.bs
```diff
@@ -2563,6 +2611,100 @@ enum class forward_progress_guarantee {
...
+      }
+
+      template&lt;receiver&lt;T> R>
+      friend auto tag_invoke(execution::connect_t, <i>just-error-sender</i>&& j, R && r) {
+        return operation_state&lt;R>{ std::move(j.err_), std::forward&lt;R>(r) };
+      }
+    };
+
+    template&lt;<i>moveable-value</i> T>
+      <i>just-error-sender</i>&lt;remove_cvref_t&lt;T> just_error(T && t) noexcept(<i>see-below</i>);
```
lewissbaker, line 2661: Do we want to require that the error type is at least `nothow_move_constructible_v<remove_cvref_t<T>>`?

This may be a wider question for what we require/assume with regards to error types sent via `execution::set_error()`.

std_execution.bs
```diff
@@ -2563,6 +2611,100 @@ enum class forward_progress_guarantee {
...
+
+      template&lt;receiver&lt;T> R>
+        requires copyable&ltT>
+      friend auto tag_invoke(execution::connect_t, const <i>just-error-sender</i>& j, R && r) {
+        return operation_state&lt;R>{ j.err_, std::forward&lt;R>(r) };
+      }
+
+      template&lt;receiver&lt;T> R>
+      friend auto tag_invoke(execution::connect_t, <i>just-error-sender</i>&& j, R && r) {
+        return operation_state&lt;R>{ std::move(j.err_), std::forward&lt;R>(r) };
```
lewissbaker, line 2656: Return type should be `operation_state<remove_cvref_t<R>>`

std_execution.bs
```diff
@@ -2563,6 +2611,100 @@ enum class forward_progress_guarantee {
...
+          catch (...) {
+            execution::set_error(std::move(s.r_), current_exception());
+          }
+        }
+      };
+
+      template&lt;receiver&lt;T> R>
+        requires copyable&ltT>
+      friend auto tag_invoke(execution::connect_t, const <i>just-error-sender</i>& j, R && r) {
+        return operation_state&lt;R>{ j.err_, std::forward&lt;R>(r) };
```
lewissbaker, line 2651: Return type should be `operation_state<remove_cvref_t<R>>`

std_execution.bs
```diff
@@ -3032,6 +3174,48 @@ are all well-formed.
...
+
+    <pre highlight=c++>
+        execution::let_done(
+          execution::then(s, [](auto&& t) {
+              return optional&lt;<i>single-sender-value-type</i>&lt;S>>{
+                static_cast&lt;decltype(t)>(t)
+              };
+            }
+          ),
+          [] {
```
lewissbaker, line 3191: `[]() noexcept {`

std_execution.bs
```diff
@@ -3032,6 +3174,48 @@ are all well-formed.
 
     3. Returns an operation state `op_state` that contains `op_state2`. When `execution::start(op_state)` is called, calls `execution::start(op_state2)`.
 
+#### `execution::done_as_optional` <b>[execution.senders.adaptors.done_as_optional]</b> #### {#spec-execution.senders.adaptors.done_as_optional}
+
+1. `execution::done_as_optional` is used to handle a done signal by mapping it into the value channel as an empty optional. The value channel is also converted into an optional. The result is a sender that never completes with done, reporting cancellation by completing with an empty optional.
+
+2. The name `execution::done_as_optional` denotes a customization point object. For some subexpression `s.`, let `S` be `decltype((s))`. If the type `S` does not satisfy <code><i>single-typed-sender</i></code>, `execution::done_as_optional(s)` is ill-formed. Otherwise, the expression `execution::done_as_optional(s)` is expression-equivalent to:
```
lewissbaker, line 3181: You mention that it is a customisation-point-object but do not actually describe the mechanism by which it can be customised. Is it not actually intended to be customisable?

std_execution.bs
```diff
@@ -3032,6 +3174,48 @@ are all well-formed.
...
+
+    <pre highlight=c++>
+        execution::let_done(
+          execution::then(s, [](auto&& t) {
+              return optional&lt;<i>single-sender-value-type</i>&lt;S>>{
```
lewissbaker, line 3186: This would be ill-formed if the single-sender-value-type is `T&`.

std_execution.bs
```diff
@@ -3032,6 +3174,48 @@ are all well-formed.
...
+              };
+            }
+          ),
+          [] {
+            return execution::just(optional&lt;<i>single-sender-value-type</i>&lt;S>>{});
+          }
+        )
+    </pre>
+
+#### `execution::done_as_exception` <b>[execution.senders.adaptors.done_as_exception]</b> #### {#spec-execution.senders.adaptors.done_as_exception}
```
lewissbaker, line 3197: If it's not actually throwing an exception (e.g. as a specific awaitable type) then perhaps this should be called `done_as_error()`?

std_execution.bs
```diff
@@ -3032,6 +3174,48 @@ are all well-formed.
...
+          <i>see-below</i> done_as_exception(S && s);
+    </pre>
+
+3. <i>Effects:</i> Equivalent to:
+
+    <pre highlight=c++>
+        return execution::let_done(
+          s,
+          [] {
+            return execution::just_error(make_exception_ptr(Exception{}));
```
lewissbaker, line 3214: Can we change this so that instead of always constructing a `std::exception_ptr` here, that it just sends `Exception{}` as the error?
The as-awaitable adapter can then turn that into an exception if needed when it throws it.

I'm concerned about the overhead of calling `make_exception_ptr` for cheap error types.

std_execution.bs
```diff
@@ -3032,6 +3174,48 @@ are all well-formed.
...
+
+    <pre highlight=c++>
+        template&lt;default_initializable Exception, sender S>
+          <i>see-below</i> done_as_exception(S && s);
+    </pre>
+
+3. <i>Effects:</i> Equivalent to:
+
+    <pre highlight=c++>
+        return execution::let_done(
```
lewissbaker, line 3211: The implementation in terms of `let_done()`, while functional is not optimal as it forces the operation-state to store a copy of the value. Whereas if it were hand-implemented then we could just have the implementation of `set_done` just call `set_error(std::move(self.r), Error{})` and not have to store the error.

std_execution.bs
ericniebler: I would say `reference_wrapper`, except that doesn't work with rvalue references. Probably decay-copy for the time being.

std_execution.bs
ericniebler: This is for consistency with the specification of `just`. I don't particularly care for this behavior either, and IMO we should revisit this for both `just` and `just_error` in R3.

std_execution.bs
ericniebler: I _think_ I prefer it as a function, but I could go either way, really.

std_execution.bs
ericniebler: Yes, thanks.

std_execution.bs
ericniebler: I'd rather we revisit that post-R2.

std_execution.bs
ericniebler: Thanks.

std_execution.bs
ericniebler: This is a quirk of the term. Not all CPOs in the stdlib are customizable. That's just what we call them. Blame me and Casey for that.

std_execution.bs
ericniebler: Ugh. I'll think about that.

std_execution.bs
ericniebler: yeah, perhaps.

std_execution.bs
ericniebler: Yeah, I think I like that better.

std_execution.bs
ericniebler: We can revisit that in R3.

std_execution.bs
```diff
@@ -2563,6 +2611,100 @@ enum class forward_progress_guarantee {
...
+
+      template&lt;class R>
+      struct operation_state {
+        T err_;
+        R r_;
+
+        friend void tag_invoke(execution::start_t, operation_state& s)
+          noexcept {
+          try {
+            execution::set_error(std::move(s.r_), std::move(err_));
```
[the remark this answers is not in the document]
ericniebler: Yes, but moving the error into the call is possibly not.

PR #234, opened 2021-11-02
remove circularity in definition of the scheduler concept; fixes #156

### LINKED ISSUE #156: Cyclic definition of schedule and scheduler
In 9.3 paragraph 1 the concept `scheduler` is defined to require `execution::schedule((S&&s))`. The definition of `execution::schedule` in 9.6.4.2 paragraph 2 states that "If `S` does not satisfy `execution::scheduler`, `execution::schedule` is ill-formed".  That is, the concept `scheduler` is defined in terms of the customization point object `schedule` which is defined in terms of the concept `scheduler`, i.e., there is a cyclic definition.

I believe the problem can be fixed by explicitly requiring the `scheduler` concept properties in the definition of `schedule` which avoids the cyclic dependency.

ericniebler:


griwes (approved):

kirkshoop (approved):

miscco:
Should we also fix the `operation_state` circularity with `execution::start`?

ericniebler:
> Should we also fix the `operation_state` circularity with `execution::start`?

What circularity are you referring to?

miscco:
tl;dr There is a cyclic dependency but I am not sure whether we should care

The `opertation_state` concept requires `execution::start`
```cpp
template<class O>
concept operation_state = destructible<O> && is_object_v<O> && requires (O& o) {
      { execution::start(o) } noexcept;
};
```

At the same time `execution::start` is defined as (abbreviated):
```
execution::start is used to start work represented by an **operation state object**

....

tag_invoke(execution::start, O), if that expression is valid. If the function selected by tag_invoke does not start the work represented by the **operation state** O,

...

The caller of execution::start(O) must guarantee that the lifetime of the **operation state object O** ...
```

So from my reading `operation_state` requires `execution::start` to be valid and `execution::start` requires its argument to be an `operation_state`, which forms a cyclic dependency.

I would like to note that an implementation can easily work around this:
```cpp
namespace _Operation_state_start {
        struct _Cpo {
            // clang-format off
            template <class _Ty>
                requires destructible<_Ty> && is_object_v<_Ty> && tag_invocable<_Cpo, _Ty&>
            _NODISCARD constexpr auto operator()(_Ty& _State) const noexcept(nothrow_tag_invocable<_Cpo, _Ty&>) {
                // clang-format on
                return tag_invoke(*this, _State);
            }
        };
    } // namespace _Operation_state_start
    using start_t = _Operation_state_start::_Cpo;

    inline namespace _Cpos {
        inline constexpr _Operation_state_start::_Cpo start;
    }

    template <class _Ty>
    concept operation_state = destructible<_Ty> && is_object_v<_Ty> && requires(_Ty& __t) {
        {execution::start(__t)};
    };
```

Essentially any CPO constraint the "basic" type constraints and then add the desired `tag_invoke` on top.

On the other hand for simple CPOs we could also write:
```cpp
template<class O>
concept operation_state = destructible<O> && is_object_v<O> && tag_invocable<start_t, O&>
```

But this will become brittle for more complex CPOs that e.g. consider coroutines like `execution::connect`.

PR #235, opened 2021-11-04
Simplify examples

miscco:
Currently the examples are really hard to read. We have many lambdas that capture by reference and do complicated things. However, the actual intent for the examples is to show how to build an asynchronous pipeline.

Therefore, pull the lambdas out, give them (hopefully) reasonable names and lets the reader focus on what is important

For obvious reasons I have not yet adopted any of the wording / explanation. If everyone is on board that this is a good idea, I am happy to put in the work to adopt the wording

### DIFF
--- a/std_execution.bs
+++ b/std_execution.bs
@@ -172,7 +172,7 @@ sender auto async_inclusive_scan(scheduler auto sch,                          //
   vector<double> partials(tile_count + 1);                                    // 4
   partials[0] = init;                                                         // 4
 
-  auto chuncked_inclusive_scan = [=](size_t i, vector<double>& partials) {    // 7
+  auto chunked_inclusive_scan = [=](size_t i, vector<double>& partials) {     // 7
     auto start = i * tile_size;                                               // 8
     auto end   = min(input.size(), (i + 1) * tile_size);                      // 8
     partials[i + 1] = *--inclusive_scan(begin(input) + start,                 // 9
@@ -185,7 +185,7 @@ sender auto async_inclusive_scan(scheduler auto sch,                          //
     return move(partials);                                                    // 13
   };
 
-  auto chuncked_update_output = [=](size_t i, vector<double>&& partials) {    // 14
+  auto chunked_update_output = [=](size_t i, vector<double>&& partials) {     // 14
     auto start = i * tile_size;                                               // 14
     auto end   = min(input.size(), (i + 1) * tile_size);                      // 14
     for_each(begin(output) + start, begin(output) + end,                      // 14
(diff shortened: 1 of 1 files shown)


std_execution.bs
```diff
@@ -168,33 +168,39 @@ sender auto async_inclusive_scan(scheduler auto sch,                          //
   std::vector<double> partials(tile_count + 1);                               // 4
   partials[0] = init;                                                         // 4
 
+  auto chuncked_inclusive_scan =
```
villevoutilainen: "chunked", not "chuncked". Repeats.

std_execution.bs
```diff
@@ -233,20 +239,22 @@ struct dynamic_buffer {                                                       //
...
-                      buf.data = std::make_unique(new std::byte[buf.size]);   // 11
-                      return std::span(buf.data.get(), buf.size);             // 12
-                    }
-                | async_read(handle)                                          // 13
-                | then(
-                    [&] (std::size_t bytes_read) {
-                      assert(bytes_read == buf.size);                         // 14
-                      return std::move(buf);                                  // 15
-                    });
+          auto valiate_and_create_storage = [&buf] (std::size_t bytes_read) { // 9
```
villevoutilainen: "validate", not "valiate".

std_execution.bs
```diff
@@ -233,20 +239,22 @@ struct dynamic_buffer {                                                       //
...
+            assert(bytes_read == sizeof(buf.size));                           // 10
```
villevoutilainen, line 239: ..and here we notice that this function doesn't validate anything, it just asserts it.

villevoutilainen (commented):
I like the idea in general. This sort of named-little-pieces code is what I expect users will have to write anyway, for anything even moderately complex. The alternative is not very readable, and reminds me of the huge switch-cases I've seen throughout the years, where the logic of each case is so complex that you forget that you're in a switch.

miscco:
So my reasoning with removing `std::` is that it introduces some line breaks in between. I would never do this in prod, but slideware usually benefits from less

msimberg:
I like moving the lambdas into named functions, and I also like `using namespace std;` in this case as it's unnecessary noise. However, what about _only_ using `using namespace std;`, i.e. no `using namespace std::execution`? It would make some lines longer again, but I don't consider it noise in this case since it makes it immediately obvious which parts come from P2300 and which don't.

ericniebler:
@miscco I just updated this PR (and force pushed to your branch). What is your intention for it?

ericniebler:
Closing this old PR. @miscco, if you'd still like to make this change to P2300, pls submit it to https://github.com/brycelelbach/wg21_p2300_execution.

PR #236, opened 2021-11-04
add receiver_adaptor utility to P2300R3

ericniebler:
refs #215

### DIFF
--- a/examples/algorithms/retry.hpp
+++ b/examples/algorithms/retry.hpp
@@ -42,18 +42,17 @@ struct _conv {
 
 // pass through all customizations except set_error, which retries the operation.
 template<class O, class R>
-struct _retry_receiver {
+struct _retry_receiver
+  : std::execution::receiver_adaptor<_retry_receiver<O, R>> {
   O* o_;
+
+  R&& base() && noexcept { return (R&&) o_->r_; }
+  const R& base() const & noexcept { return o_->r_; }
+
   explicit _retry_receiver(O* o) : o_(o) {}
-  friend void tag_invoke(std::execution::set_error_t, _retry_receiver&& self, auto&&) noexcept {
-    self.o_->_retry(); // This causes the op to be retried
-  }
-  // Forward all other tag_invoke-based CPOs (_copy_cvref_t from P1450):
-  template <_decays_to<_retry_receiver> Self, class... As, class Tag>
-  friend auto tag_invoke(Tag tag, Self&& self, As&&... as)
-    noexcept(std::is_nothrow_invocable_v<Tag, _copy_cvref_t<Self, R>, As...>)
-    -> std::invoke_result_t<Tag, _copy_cvref_t<Self, R>, As...> {
-    return ((Tag&&) tag)((_copy_cvref_t<Self, R>&&) self.o_->r_, (As&&) as...);
+
+  void set_error(auto&&) && noexcept {
+    o_->_retry(); // This causes the op to be retried
   }
 };
 
--- a/examples/algorithms/retry.hpp
+++ b/examples/algorithms/retry.hpp
@@ -1,0 +1,108 @@
+/*
+ * Copyright (c) NVIDIA
+ *
+ * Licensed under the Apache License Version 2.0 with LLVM Exceptions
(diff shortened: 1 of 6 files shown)


kirkshoop (commented):
I noticed that retry doe not use the new receiver base class. Should it?

ericniebler:
> I noticed that retry doe not use the new receiver base class. Should it?

For `retry`, the inner receiver needs to live in the operation state. So a `receiver_adaptor` that aggregates an inner receiver isn't very helpful. This could indicate that there is a more general form for a receiver adaptor than what is proposed.

kirkshoop:
> This could indicate that there is a more general form for a receiver adaptor than what is proposed.

Yeah. I think many receivers must refer to a receiver stored in the operation state. Perhaps the general for is for the adapter to require the derived receiver to have a get_receiver() method

ericniebler:
> Perhaps the general form is for the adapter to require the derived receiver to have a get_receiver() method

This is an interesting suggestion. I'll noodle on that, thanks.

examples/algorithms/retry.hpp
```diff
@@ -0,0 +1,107 @@
...
+};
+
+// pass through all customizations except set_error, which retries the operation.
+template<class O, class R>
+struct _retry_receiver
+  : std::execution::receiver_adaptor<_retry_receiver<O, R>> {
+  O* o_;
+
+  R&& base() && noexcept { return (R&&) o_->r_; }
+  const R& base() const & noexcept { return o_->r_; }
```
kirkshoop, line 50: Yeah. That looks good. I did confuse base with base class for a bit. If you want to try another name. The function of that receiver is destination or completion or result or output.

examples/retry.cpp
```diff
@@ -0,0 +1,62 @@
...
+  }
+};
+
+int main() {
+  auto x = retry(fail_some{});
+  // prints:
+  //   fail!
+  //   fail!
+  //   success!
+  auto [a] = std::this_thread::sync_wait(std::move(x)).value();
```
kirkshoop, line 61: Should this validate or print`a`?

kirkshoop (commented):
The complexity of supporting both storage in the adaptor and outside the adaptor seems a bit excessive when the derived receiver can store it, yes?

It looks like you might be updating the docs before switching this out of draft. So I'll just leave comments.

examples/algorithms/retry.hpp
ericniebler: All the iterator and view adaptors provide accessors named `base()` that return the inner thing.

ericniebler:
> The complexity of supporting both storage in the adaptor and outside the adaptor seems a bit excessive when the derived receiver can store it, yes?

It's true, the derived type can always just store the receiver. But much of the complexity will be needed anyway for the other adaptors, which will almost certainly want the adaptor to store the inner object. All sender adaptors aggregate the inner sender, for instance.

Given that much of the machinery needs to exist anyway, I figured why not put it to good use here? It does simplify the `then` receiver somewhat.

kirkshoop:
I don't consider it blocking. 

This design works, but is more complicated to specify and to discuss in LEWG. 

Derived senders and op states can also provide the storage so I fall on the other side of the complexity convenience trade off on this one.

kirkshoop:
More as a laugh. It is also possible to take a lambda in the adaptor constructor which would return the base.  if that would seem lower impact than adding the base() functions to the derived.

examples/algorithms/retry.hpp
kirkshoop: Yeah. It looks weirder when base() is implemented in the derived class.

ericniebler:
> More as a laugh. It is also possible to take a lambda in the adaptor constructor which would return the base. if that would seem lower impact than adding the base() functions to the derived.

Passing it to the constructor isn't enough. The rest of the adaptor needs to know the type of the base, too. Before settling on this design, I played with one where a lambda was specified as a template parameter. I didn't find it very satisfying, and it was a little cryptic.

ericniebler:
> more complicated to specify and to discuss in LEWG

I have an idea for how this can be specified in a fairly straightforward manner.

kirkshoop (approved):

PR #239, opened 2021-11-05
add schedule_from algorithm

ericniebler:


griwes (approved):

kirkshoop (approved):

include/execution.hpp
```diff
@@ -1468,6 +1465,184 @@ namespace std::execution {
...
+              // information is saved off in a variant so that when this receiver
+              // completes, it can read the completion out of the variant and forward
+              // it to the output receiver after transitioning to the scheduler's
+              // context.
+              struct __receiver2 {
+                __op1* __op_;
+
+                // If the work is successfully scheduled on the new execution context
+                // and is ready to run, forward the completion signal in the variant
+                friend void tag_invoke(set_value_t, __receiver2&& self) noexcept try {
```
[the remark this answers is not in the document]
ericniebler: No, this is a [function try block] (https://en.cppreference.com/w/cpp/language/function-try-block).

include/execution.hpp
```diff
@@ -1468,6 +1465,184 @@ namespace std::execution {
...
+                      tag((Receiver&&) op->__rec_, (decltype(args)&&) args...);
+                    }, (decltype(tup)&&) tup);
+                  }, std::move(self.__op_->__data_));
+                } catch(...) {
+                  set_error((Receiver&&) self.__op_->__rec_, current_exception());
+                }
+
+                template <__one_of<set_error_t, set_done_t> Tag, class... Args>
+                  requires invocable<Tag, Receiver, Args...>
+                friend void tag_invoke(Tag, __receiver2&& self, Args&&... args) noexcept {
```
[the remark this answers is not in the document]
ericniebler: No, `set_error` and `set_done` are required to be `noexcept`.

include/execution.hpp
```diff
@@ -1468,6 +1465,184 @@ namespace std::execution {
...
+              //   >
+              template <class... Ts>
+                using __bind_tuples = __bind_front<variant, tuple<set_done_t>, Ts...>;
+
+              using __bound_values_t = typename sender_traits<Sender>::template value_types<
+                  __transform<
+                      __bind_front<tuple, set_value_t>::template __f,
+                      decay_t
+                  >::template __f,
+                  __bind_tuples>;
```
ericniebler, line 1516: @lewissbaker This causes the source sender's values to be decay-copied into the operation state. I think this is correct, but the current wording in P2300 doesn't reflect the fact that the values are decay-copied before being propagated to the completion sender.

include/execution.hpp
```diff
@@ -1468,6 +1465,184 @@ namespace std::execution {
...
+                __op1* __op_;
+
+                // If the work is successfully scheduled on the new execution context
+                // and is ready to run, forward the completion signal in the variant
+                friend void tag_invoke(set_value_t, __receiver2&& self) noexcept try {
+                  std::visit([op = self.__op_](auto&& tup) -> void {
+                    std::apply([op](auto tag, auto&&... args) -> void {
+                      tag((Receiver&&) op->__rec_, (decltype(args)&&) args...);
+                    }, (decltype(tup)&&) tup);
+                  }, std::move(self.__op_->__data_));
```
ericniebler, line 1543: This causes references to objects stored in the operation state to be passed through the value channel. I notice that in `unifex::finally`, we first copy the objects out of the operation state and into locals before propagating them through the value channel. Should we be doing that here? cc @lewissbaker @kirkshoop

kirkshoop: The comment in finally indicates that the copy is made to simplify the operation destructor. 

It seems like an implementation decision not a correctness issue

PR #241, opened 2021-11-07
Fix `just_error::operation_state`

miscco:
we require that the receiver connected with `just_error` is a `receiver_of<R, T>`

This implies that `set_error` is noexcept, so the whole try catch block is unnecessary

### DIFF
--- a/std_execution.bs
+++ b/std_execution.bs
@@ -3126,14 +3126,8 @@ enum class forward_progress_guarantee {
         T err_;
         R r_;
 
-        friend void tag_invoke(execution::start_t, operation_state& s)
-          noexcept {
-          try {
+        friend void tag_invoke(execution::start_t, operation_state& s) noexcept {
             execution::set_error(std::move(s.r_), std::move(err_));
-          }
-          catch (...) {
-            execution::set_error(std::move(s.r_), current_exception());
-          }
         }
       };
 


ericniebler:
We don't know that converting a `T` to whatever type the receiver's `set_error` accepts doesn't throw, so the `try` block is necessary.

miscco:
> We don't know that converting a `T` to whatever type the receiver's `set_error` accepts doesn't throw, so the `try` block is necessary.

That is not how it is currently specified, the `connect` method is given as:
```cpp
      template<receiver R>
        requires receiver<R, T> && copyable<T>
      friend auto tag_invoke(execution::connect_t, const just-error-sender& j, R && r) {
        return operation_state<remove_cvref_t<R>>{ j.err_, std::forward<R>(r) };
      }
      template<receiver R>
        requires receiver<R, T>
      friend auto tag_invoke(execution::connect_t, just-error-sender&& j, R && r) {
        return operation_state<remove_cvref_t<R>>{ std::move(j.err_), std::forward<R>(r) };
      }
```

Note that we constraint it as `receiver<R,T>` which requires
```cpp
    template <class _Ty, class _Exception = exception_ptr>
    concept receiver = _Moveable_value<_Ty> && requires (remove_cvref_t<_Ty>&& __t, _Exception&& __e) {
        { _EXEC set_done(_STD move(__t)) } noexcept;
        { _EXEC set_error(_STD move(__t), _STD forward<_Exception>(__e)) } noexcept;
    };
```

So either the receiver has a matching `set_error` without any conversion or we cannot connect.

ericniebler:
You're right. Sorry.

miscco:
I mean we *could* change the definition of `connect` to take a `receiver<R, U>` and additionally constraint it with `convertible_to<T, U>`

I cannot really tell if it is desirable though

PR #242, opened 2021-11-07
Move `transfer_just` below `just_error` and `just_done`

miscco:
The three just algorithms belong together. Despite that `transfer_just` is a whole other beast. Move it down


griwes:
This is moot now that the specification of the just family has been unified.

PR #247, opened 2021-11-11
add reference implementation of the execution::on algorithm

ericniebler:


include/execution.hpp
```diff
@@ -1692,6 +1676,111 @@ namespace std::execution {
...
+              using R = __t<R_>;
+
+              friend void tag_invoke(start_t, __operation& self) noexcept {
+                start(std::get<0>(self.__data_));
+              }
+
+              template <class S, class R>
+              __operation(Sch sch, S&& s, R&& r)
+                : __data_{in_place_index<0>, __conv{[&, this]{
+                    return connect(schedule(sch), __receiver<R_>{{}, this});
```
miscco, line 1735: This is not according to spec:
`Otherwise, constructs a sender s2. When s2 is connected with some receiver out_r, it results in an operation state op_state. When execution::start is called on op_state, it:`

Here, you are immediately  calling `execution::schedule` on the passed scheduler during construction of the operation state. However, it should only be called when `execution::start` is called, as that is the point where the receiver `r` is created.

miscco (commented):
So this does work and it some quite clever stuff.

At the same time I would never merge this is in any codebase, especially not if it is intended as a reference implementation. The reason is that it is barely decipherable what is going on.

Within the implementation we have `S_`, `S`, `R_`, `R` and there is also `Sch_` and `Sch`. It makes it so much harder to follow what is happening. 

Citing Kate Gregory, can we maybe use names like `InputSender` and `OutputReceiver` or anything that has more than a single letter? I doubt that it will slow us down to type some few characters.

include/execution.hpp
```diff
@@ -1692,6 +1676,111 @@ namespace std::execution {
...
+      template <class R_>
+        struct __receiver_ref : receiver_adaptor<__receiver_ref<R_>> {
+          using R = __t<R_>;
+          R& __r_;
+          R&& base() && noexcept { return (R&&) __r_; }
+          const R& base() const & noexcept { return __r_; }
+        };
+
+      template <class Sch_, class S_>
+        struct __sender : sender_base {
```
miscco, line 1691: In my current implementation I am specializing `sender_traits` for the inner sender
```cpp
    template <scheduler _Scheduler, sender _InnerSender>
    struct sender_traits<_On::_OuterSender<_Scheduler, _InnerSender>>
        : decltype(_Get_sender_traits_base<_InnerSender>()) {};
```

Would that be a valid alternative

include/execution.hpp
ericniebler: Yes, I think the spec is wrong. I'm planning to suggest this change. The spec is also wrong about the lifetime of op_state3.

include/execution.hpp
ericniebler: Your sender_traits specialization must dispatch to sender_traits on the inner receiver. There may also be a specialization of sender_traits for that sender type, and your suggested implementation would miss it.

miscco: Thanks for pointing that out! Totally missed it

include/execution.hpp
miscco: I am not sure whether it would be simpler to just store `sched` until we actually call `start` 

But yeah, both should work

include/execution.hpp
```diff
@@ -1692,6 +1676,111 @@ namespace std::execution {
...
+          Sch __sch_;
+          S __s_;
+
+          template <class R_>
+            struct __operation;
+
+          template <class R_>
+            struct __receiver : receiver_adaptor<__receiver<R_>> {
+              using R = __t<R_>;
+              __operation<R_>* __op_;
```
miscco, line 1704: So the inner receiver needs essentially all the state. And we never give it out to the user. 

So what I thought an simpler setup is to store `sch`, `__r_` and `__s_` inside the receiver and make that a member of the operation_state. 

Yes we would construct `r` when `connect` is called and not when `start` is called but that should be opaque. That way  it is much simpler to reason about lifetimes

include/execution.hpp
miscco: Another issue that comes from this is that you need the variant and the `__conv` dance.

If you store `sched` as a member and only call `execution::schedule` during `start` then you can immediately move the result from that operation to `execution::start` without having to store it.

Without the need to store two distinct types in the operation state you can simply go with a poor mans optional where you store the result from the `execution::connect` call

miscco: Argh scratch that, I missed that we need to hold it..

miscco: > Yes, I think the spec is wrong. I'm planning to suggest this change. The spec is also wrong about the lifetime of op_state3.

Are you talking about op_state2 which you construct here?  I would agree that it would be valid to drop it on the floor as you did once the `op_state3` comes along

ericniebler: > Are you talking about `op_state2` which you construct here? I would agree that it would be valid to drop it on the floor as you did once the `op_state3` comes along

No, I mean `op_state3` which (somewhat confusingly) is constructed before `op_state2`. I should probably switch the names of the operation states and possibly restructure this section to avoid confusion.

include/execution.hpp
ericniebler: I don't think this works. Once `set_value` is called on the inner receiver, its lifetime effectively ends, which will be true if you decide, like I did, to reuse the storage for the intermediate operation state. You don't want that teardown to destroy _all_ your state; then you can't launch the inner sender.

Also be mindful of the fact that receivers are copied/moved when calling connect. You must take care that the copy/move isn't potentially throwing, or else that if it does throw, you still have a valid receiver on which you can call `set_error`.

miscco (approved):
Thanks a lot, it is much better to follow now

kirkshoop (approved):

PR #251, opened 2021-11-12
Add a rationale for why potentially eagerly execution algorithms were removed

ericniebler:


std_execution.bs
```diff
@@ -1243,6 +1244,133 @@ talk about the semantics of some of the algorithms proposed in this paper.
...
+    on the operation state. That puts the onus of dealing with the race
+    condition on the sender author, which is arguably where it belongs. But it
+    has the negative effect of delaying the completion of the operation. Even
+    when the work is done and the continuation available, the sender has to
+    withold the completion, possibly indefinitely.
+
+Of these options, (3) seems the least bad but would seem to run contrary to the
+<i>raison d'être</i> of eager algorithms, which ostensibly is to make the
+results available sooner.
+
```
kirkshoop, line 1349: in the GPU space the argument seems to be that eager lets them start faster, not complete faster, yes?

std_execution.bs
```diff
@@ -1243,6 +1244,133 @@ talk about the semantics of some of the algorithms proposed in this paper.
...
+    problem.
+
+3.  Push the problem back onto eager senders by requiring them to never call a
+    receiver's `set_*` completion until after `execution::start` has been called
+    on the operation state. That puts the onus of dealing with the race
+    condition on the sender author, which is arguably where it belongs. But it
+    has the negative effect of delaying the completion of the operation. Even
+    when the work is done and the continuation available, the sender has to
+    withold the completion, possibly indefinitely.
+
```
kirkshoop, line 1345: In the call last week I asked how a scheduler could produce an eager sender at all. the whole point is to complete on a particular context ands the only way top achieve that is to queue the receiver after it has been connected.

std_execution.bs
```diff
@@ -1243,6 +1244,133 @@ talk about the semantics of some of the algorithms proposed in this paper.
...
+callback with the receiver's stop token that will request stop on the eager
+sender's stop source.
+
+As the eager operation does not know at the time that it is launched what the
+type of the receiver is going to be, and thus whether or not the stop token
+returned from `execution::get_stop_token` is an `std::unstoppable_token` or not,
+the eager operation is going to need to assume it might be later connected to a
+receiver with a stop token that might actually issue a stop request. Thus it
+needs to incur the runtime overhead of supporting cancellation, even if
+cancellation will never be requested by the caller.
```
kirkshoop, line 1367: part of the cost of this is that the stop_token_callback must be type-erased with potential allocation.

kirkshoop (approved):
Very cool, thanks!

std_execution.bs
ericniebler: I can only imagine they want the work to start faster so that it finishes faster, no?

std_execution.bs
ericniebler: Of course that's true, and I don't think the "possibly eager" crowd is asking for _every_ sender to start eagerly. I think they would like some algorithms to start their work eagerly. `schedule` is certainly not one of them for the reason you mention.

ericniebler:
Merging now, we can continue the conversation about this PR and I can make any tweaks we deem necessary.

std_execution.bs
kirkshoop: What I heard was that they had limits in getting the data to the GPU and needed that to start before the work was actually started.

std_execution.bs
kirkshoop: I just thought that this should be part of the justification for removing eager.

std_execution.bs
```diff
@@ -1247,6 +1248,133 @@ talk about the semantics of some of the algorithms proposed in this paper.
...
+completed executing by the time that the destructor is run we need to decide
+what the semantics of the destructor is.
+
+There are three main strategies that can be adopted here, none of which is
+particularly satisfactory:
+
+1.  Make this undefined-behaviour - the caller must ensure that any
+    eagerly-executing sender is always joined by connecting and starting that
+    sender. This approach is generally pretty hostile to programmers,
+    particularly in the presence of exceptions, and is thus not recommended.
```
lewissbaker, line 1278: I would probably expand on this a little and say that it is hostile to the ability to compose these operations.
If I pass ownership of an eager sender to another algorithm's sender then I need to make sure both that I connect/join that sender and that the algorithm I passed the sender to also guarantees to connect/join the eager sender if I connect/join that algorithm's sender.
This can be problematic if composing an eager sender with senders that might fail on `connect()` (e.g. because they allocate resources in `connect()` which might fail) or if composing with other eager senders.
e.g. `when_all(eager_op_1(), eager_op_2())`

As eager operations typically need to allocate resources on the heap in order to start the operations, calling an eager algorithm can potentially throw an exception, so in this case calling `eager_op_1()` might start an eager op and then `eager_op_2()` throws an exception. This was the basic problem that I described in my CppCon talk.

std_execution.bs
```diff
@@ -1247,6 +1248,133 @@ talk about the semantics of some of the algorithms proposed in this paper.
...
+    destructor. It reduces the risk of deadlock but does not eliminate it.
+
+### Eager senders complicate algorithm implementations ### {#design-lazy-algorithms-complexity}
+
+Algorithms that can assume they are operating on senders with strictly lazy
+semantics are able to make certain optimizations that are not available if
+senders can be potentially eager. With lazy senders, an algorithm can safely
+assume that a call to `execution::start` on an operation state strictly happens
+before any receiver methods are called. This frees the algorithm from needing to
+resolve potential race conditions between the start and completion of the work.
```
lewissbaker, line 1327: The guarantee that completion-signalling methods are not called on the receiver until after `start()` is called is guaranteed for all senders, so I don't think that's a distinguishing feature of eager vs lazy.

Instead, the main difference is whether or not the logic of a particular algorithm is guaranteed not to run until after the sender has been connected and started.

e.g. we might write something like:
```
sender auto example(sender src) {
  return sequence(do_a(), then(src, [] { do_b(); }), do_c());
}
```
An algorithm like this might expect that do_a(), do_b() and do_c() are sequenced and therefore do not need synchronisation. If the `then()` algorithm might actually execute the lambda eagerly then this might break the algorithm's assumption as it could be possible for `do_b()` to run before `do_a()` has completed.

std_execution.bs
```diff
@@ -1247,6 +1248,133 @@ talk about the semantics of some of the algorithms proposed in this paper.
...
+
+2.  Require senders to declare whether they are eager or not with a query.
+    Algorithms can then implement two different implementation strategies, one
+    for strictly lazy senders and one for potentially eager senders. This
+    addresses the performance problem of (1) while compounding the complexity
+    problem.
+
+3.  Push the problem back onto eager senders by requiring them to never call a
+    receiver's `set_*` completion until after `execution::start` has been called
+    on the operation state. That puts the onus of dealing with the race
```
lewissbaker, line 1344: Again, this is something that is already required as part of the sender/receiver contract.

The limitation on algorithms is that it "should not do the work" until the operation is started, not that it "should not complete" until it is started.

std_execution.bs
lewissbaker: The type erasure is required (and indeed is implicit in the design of stop-tokens/stop-callbacks).
However, you shouldn't need a separate heap allocate storage for the stop-token callback - that should be stored inside the operation-state returned from connecting the eager sender.

std_execution.bs
```diff
@@ -1247,6 +1248,133 @@ talk about the semantics of some of the algorithms proposed in this paper.
...
+the eager operation is going to need to assume it might be later connected to a
+receiver with a stop token that might actually issue a stop request. Thus it
+needs to incur the runtime overhead of supporting cancellation, even if
+cancellation will never be requested by the caller.
+
+The eager operation will also need to do this to support sending a stop request
+to the eager operation in the case that the sender representing the eager work
+is destroyed before it has been joined (assuming strategy (5) or (6) listed
+above is chosen).
+
```
lewissbaker, line 1377: Other topics you didn't mention here:
* Extra synchronisation/atomic-ops required to resolve race between concurrent connect/start and completion of the operation.
* Extra heap-allocation required to hold state of the operation-state of the eager operation while it's executing before the returned sender has been connected/started.
* Lack of the ability to implicitly propagate context from the receiver to the eager operation. As the operation may run to completion before a receiver is connected/started, we won't necessarily have a receiver available which can obtain implicitly propagated context information from (like the current scheduler, stop token, priority, deadline, etc.).

std_execution.bs
ericniebler: I don't understand. If you don't know the type of the receiver, you don't know the type of the stop callback for that receiver. It needs to be type erased, and since you don't know how big it will be, that could incur a dynamic allocation. What am I missing?

PR #253, opened 2021-11-12
add a static_thread_pool execution context

ericniebler:


mjgarland:
What's the plan for implementing the launch of multi-thread tasks via `bulk` on this thread pool where I need to create a group of `k` activities that are going to cooperatively perform some computation like `inclusive_scan`?

ericniebler:
@mjgarland wrote:
> What's the plan for implementing the launch of multi-thread tasks via `bulk` on this thread pool where I need to create a group of `k` activities that are going to cooperatively perform some computation like `inclusive_scan`?

You mean, you want to guarantee that `k` activities are all launched together and concurrently? There really isn't a plan at the moment. I mean, I could write an algorithm that keeps scheduling work onto the threadpool until it has `k` in-flight units of work with different thread-ids, but that's pretty clown-y.

Do you need this example execution context to provide an API for that? I'm curious about your requirements.

ericniebler:
>  I could write an algorithm that keeps scheduling work onto the threadpool until it has k in-flight units of work with different thread-ids, but that's pretty clown-y.

Actually, this isn't so clowny, I don't think. You declare a barrier initialized with `k`, schedule `k` items on the threadpool that arrive_and_waits the barrier, and then does the work of the `inclusive_scan`.

Blocking threads in a threadpool is bad, but is there a better way?

PR #259, opened 2021-11-15
Update `run_loop` wording to introduce synchronizes with edges.

brycelelbach:


std_execution.bs
```diff
@@ -4130,16 +4130,14 @@ from the operation before the operation completes.
...
     <pre highlight="c++">
     void run_loop::push_back(<i>run-loop-opstate-base</i>* item);
     </pre>
 
     1. <i>Effects:</i> Adds `item` to the back of the queue and increments <i>count</i> by `1`.
 
-    2. <i>Synchronization:</i> This function is thread-safe with respect to `pop_front()` and `finish()`. Adding a queue item and incrementing the <i>count</i> are executed atomically.
-
+    2. <i>Synchronization:</i> This operation synchronizes with the `pop_front` operation that obtains `item`.
+
```
ericniebler, line 4140: I confess I don't understand how this guarantees that multiple threads can call `push_back` concurrently, or that `push_back` can run concurrently with a call to `pop_front` that returns an object other than `item`.

brycelelbach: Ah, I forgot an edit.

std_execution.bs
```diff
@@ -4128,18 +4128,16 @@ from the operation before the operation completes.
 
         * <i>count</i> is `0` and <i>state</i> is <i>finishing</i>, in which case `pop_front()` returns `nullptr`; or
 
-        * <i>count</i> is greater than `0`, in which case `pop_front()` removes an item from the front of the queue, decrements <i>count</i> by `1`, and returns the item.
-
-    2. <i>Synchronization:</i> This function is thread-safe with respect to `push_back()` and `finish()`. Removing a queue item and decrementing <i>count</i> are executed atomically.
+        * <i>count</i> is greater than `0`, in which case an item is atomically removed from the front of the queue, <i>count</i> is atomically decremented by `1`, and the removed item is returned.
```
ericniebler, line 4131: Can another thread observe the queue state after the item has been removed but before the count is decremented?

std_execution.bs
ericniebler: Does this wording imply that the two operations (add item, increment count) are executed as a single atomic operation?

include/execution.hpp
```diff
@@ -172,6 +172,14 @@ namespace std::execution {
...
       __has_sender_types<sender_traits<remove_cvref_t<S>>>;
 
+  template<class S, class... Ts>
+    concept sender_of =
+      typed_sender<S> &&
+      same_as<
+        __types<Ts...>,
+        typename sender_traits<S>::template
+            value_types<__types, type_identity_t>>;
+
```
ericniebler, line 182: This comes from a different commit on main, I think.

std_execution.bs
```diff
@@ -4126,19 +4128,17 @@ from the operation before the operation completes.
...
-
-    2. <i>Synchronization:</i> This function is thread-safe with respect to `push_back()` and `finish()`. Removing a queue item and decrementing <i>count</i> are executed atomically.
+        * <i>count</i> is greater than `0`, in which case an item is removed from the front of the queue, <i>count</i> is decremented by `1`, and the removed item is returned.
 
     <pre highlight="c++">
     void run_loop::push_back(<i>run-loop-opstate-base</i>* item);
     </pre>
 
-    1. <i>Effects:</i> Adds `item` to the back of the queue and increments <i>count</i> by `1`.
+    1. <i>Effects:</i> Atomically adds `item` to the back of the queue and increments <i>count</i> by `1`.
```
ericniebler, line 4139: This `atomically` not needed since it's covered above.

std_execution.bs
```diff
@@ -4126,19 +4128,17 @@ from the operation before the operation completes.
...
 
-    2. <i>Synchronization:</i> This function is thread-safe with respect to `pop_front()` and `finish()`. Adding a queue item and incrementing the <i>count</i> are executed atomically.
+    2. <i>Synchronization:</i> This operation synchronizes with the `pop_front` operation that obtains `item`.
```
ericniebler, line 4141: Do you need to say something about not introducing data races?

std_execution.bs
brycelelbach: Addressed by moving to barrier style wording (https://eel.is/c++draft/thread.barrier#class-4), as per Olivier's suggestion

std_execution.bs
brycelelbach: Addressed by moving to barrier style wording (https://eel.is/c++draft/thread.barrier#class-4), as per Olivier's suggestion.

include/execution.hpp
brycelelbach: Yah not sure how this got in here?

std_execution.bs
brycelelbach: Nope - that's covered by some wording in the front matter, similar to what we do for barrier (https://eel.is/c++draft/thread.barrier#class-4).

include/execution.hpp
brycelelbach: Fixed.

std_execution.bs
```diff
@@ -4014,7 +4014,9 @@ from the operation before the operation completes.
 
 2. A `run_loop` instance has an associated <i>count</i> that corresponds to the number of work items that are in its queue. Additionally, a `run_loop` has an associated <i>state</i> that can be either <i>finishing</i> or <i>not-finishing</i>.
 
-3. [<i>Note:</i> Implementations are encouraged to use an intrusive queue of operation states to hold the work units to make scheduling allocation-free. -- <i>end note</i>]
+3. Concurrent invocations of the member functions of `run_loop`, other than its destructor, do not introduce data races. The member functions `pop_front`, `push_back`, and `finish` execute atomically.
```
ericniebler, line 4017: Maybe we should exclude the `run()` member from the comment about data races since we don't mean for it to be called concurrently.

brycelelbach: Good point.

ericniebler (approved):

PR #264, opened 2021-11-17
Implement the when_all() sender adaptor

ericniebler:


include/execution.hpp
```diff
@@ -1873,6 +1860,283 @@ namespace std::execution {
...
+          typed_sender<tag_invoke_result_t<transfer_just_t, Scheduler, Values...>>
+      auto operator()(Scheduler&& sched, Values&&... values) const
+        noexcept(nothrow_tag_invocable<transfer_just_t, Scheduler, Values...>)
+        -> tag_invoke_result_t<transfer_just_t, Scheduler, Values...> {
+        return tag_invoke(*this, (Scheduler&&) sched, (Values&&) values...);
+      }
+      template <scheduler Scheduler, __movable_value... Values>
+      auto operator()(Scheduler&& sched, Values&&... values) const
+        -> decltype(transfer((Scheduler&&) sched, just((Values&&) values...))) {
+        return transfer((Scheduler&&) sched, just((Values&&) values...));
```
griwes, line 1878: The arguments to transfer here are backwards.

griwes (approved):
Approve modulo the one comment below.

include/execution.hpp
```diff
@@ -1873,6 +1860,251 @@ namespace std::execution {
...
+                    // Transition to the "done" state if and only if we're in the
+                    // "started" state. (If this fails, it's because we're in an
+                    // error state, which trumps cancellation.)
+                    if (__op_->__state_.compare_exchange_strong(expected, __done)) {
+                      __op_->__stop_source_.request_stop();
+                    }
+                    (void) __op_->__barrier_.arrive(); // potentially throwing
+                  } catch(...) {
+                    // Try to put this operation into the error state
+                    ((__receiver&&) *this).set_error(current_exception(), __done);
```
kirkshoop, line 1951: I think that if 'done' won the race, then errors are irrelevant and should be discarded. the request to discard results 'happens-before' the error

include/execution.hpp
```diff
@@ -1873,6 +1860,251 @@ namespace std::execution {
...
+                        std::get<Index>(__op_->__values_).emplace(
+                            (Values&&) vals...);
+                      }
+                      (void) __op_->__barrier_.arrive();
+                    } catch(...) {
+                      ((__receiver&&) *this).set_error(current_exception());
+                    }
+                  template <class Error>
+                      requires receiver<Receiver, Error>
+                    void set_error(Error&& error, __state_t expected = __started) && noexcept {
```
kirkshoop, line 1926: I would prefer an internal method that set_value, set_error and set_done can call with an explicit expected state.

kirkshoop (commented):
Very cool! pretty code :) I had a couple of comments..

include/execution.hpp
ericniebler: I'm not sure what you mean. One method for all three signals? But the code is different for each.

include/execution.hpp
ericniebler: Right, but we only get here if trying to transition to the "done" state has failed. Maybe when `std::barrier::arrive` throws, I should just consider the error fatal?

include/execution.hpp
```diff
@@ -1873,6 +1860,251 @@ namespace std::execution {
...
+              }
+
+              child_optstates_tuple_t __child_states_;
+              Receiver __recvr_;
+              barrier<__completion_fn> __barrier_;
+              // Could be non-atomic here and atomic_ref everywhere except __completion_fn
+              atomic<__state_t> __state_{__started};
+              error_types<variant> __errors_{};
+              tuple<typename sender_traits<__t<SenderIds>>::template
+                  value_types<tuple, optional>...> __values_{};
```
ericniebler, line 2055: I could save some space here by putting the values and errors into one variant, but I would need to figure out how to "lock" the variant while writing values into it so that if another thread errors, it doesn't try to concurrently access it.

include/execution.hpp
kirkshoop: The idea was to make an internal __set_error(Error, __state_t) and call that from set_value and set_done instead of this->set_error() with a defaulted arg. set_error() could forward to __set_error as well.

include/execution.hpp
kirkshoop: if the compare exchange fails then the error would own the call to arrive(), so done should not call arrive.

If done() wins the compare exchange race then it should own the call to arrive.

if arrive throws, then attempting to shift to an error state will only call arrive again. might as well loop on arrive in done.. I think terminate is a better option than trying to move from the done state to the error state.

ericniebler: I'll be replacing the `barrier` with a simple atomic ref count. That failure mode will simply go away, and this code gets simpler.

include/execution.hpp
```diff
@@ -172,32 +174,27 @@ namespace std::execution {
...
-  template <class... As>
-    requires (sizeof...(As) <= 1)
-    using __single_or_void_t = __t<__front<As..., void>>;
+      same_as<__types<Ts...>, value_types_of_t<S, __types, __single_t>>;
 
   template<class S>
     using __single_sender_value_t =
-      typename sender_traits<remove_cvref_t<S>>
-        ::template value_types<__single_or_void_t, __single_or_void_t>;
+      value_types_of_t<S, __single_or_void_t, __single_t>;
```
miscco, line 197: I am a bit unsure why we need the duplication of `__single_or_void_t` and `__single_t` Shouldnt the latter be subsumed by the former?

kirkshoop (approved):

PR #267, opened 2021-11-19
more rigorous specification for execution::when_all with cancellation

### LINKED ISSUE #256: when_all specification says nothing about cancellation
... nor does it say that `when_all`'s child receivers are stop token providers.

### LINKED ISSUE #261: when_all wording requiring all work to finish even when one finishes with set_done
The way we're currently wording this:

> If execution::set_value(ri, ti...) is called for every ri, execution::set_value(out_r, t0..., t1..., ..., tn...) is called, where n is sizeof...(s) - 1.
> 
> Otherwise, if execution::set_error(ri, e) is called for any ri, execution::set_error(out_r, e) is called.
> 
> Otherwise, if execution::set_done(ri) is called for any ri, execution::set_done(out_r) is called.

In other words, `set_error` trumps `set_done`. To implement these semantics, a `set_done` signal from a child must wait in case one of the other children completes with `set_error`. I don't think this is what we want.

I think we want to treat `set_error` and `set_done` the same: request stop on the stop source and propagate the signal (error or done) when all children complete.

### LINKED ISSUE #263: What does when_all with no arguments mean?
Should we constrain it to require at least one argument?

### LINKED ISSUE #266: when_all needs a special case to handle senders that can never complete with values
If any input senders have `value_types` of `Variant<>` then `when_all` should have a `value_types` of `Variant<>`

ericniebler:
fixes #256, fixes #261, fixes #263, fixes #266

### DIFF
--- a/std_execution.bs
+++ b/std_execution.bs
@@ -2439,6 +2439,9 @@ namespace std::execution {
     struct get_stop_token_t;
     inline constexpr get_stop_token_t get_stop_token{};
   }
+  template <class R>
+    using stop_token_type_t =
+      remove_cvref_t<decltype(get_stop_token(declval<R>()))>;
 
   // [execution.op_state], operation states
   template&lt;class O>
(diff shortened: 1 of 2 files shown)


kirkshoop (approved):

std_execution.bs
```diff
@@ -3582,29 +3585,71 @@ are all well-formed.
...
+        template &lt;class... Ts>
+            requires (sizeof...(Ts) <= 1)
+          using <i>zero-or-one</i> = void;
+        </pre>
+
+    Otherwise, the expression <code>execution::when_all(s<i><sub>i</sub></i>...)</code> is expression-equivalent to:
+
+    1. <code>tag_invoke(execution::when_all, s<i><sub>i</sub></i>...)</code>, if that expression is valid and its type satisfies `execution::sender`. If the function selected by `tag_invoke` does not return a sender that sends a concatenation of values sent by <code>s<i><sub>i</sub></i>...</code> when they all complete with `set_value`, the program is ill-formed with no diagnostic required.
+
+    2. Otherwise, constructs a sender `w` of type `W`. When `w` is connected with some receiver `out_r` of type `OutR`, it returns an operation state `op_state` specified as below:
```
griwes, line 3605: Drop the part about op_state here, since it's spelled out below?

griwes (approved):

std_execution.bs
ericniebler: I added it here because I mention `op_state` in the specification of the inner receivers' semantics, before I introduce `op_state` below.

Issue #271, opened 2021-11-20
Should receiver queries apply to more than just receivers?


ericniebler:
Currently `get_stop_token` and friends are constrained to only work on receivers. In libunifex, we also use them with coroutine promise types. Consider loosening the constraints.

kirkshoop:
Yes. There is more detail in #280, but I expect get_scheduler/get_allocator/get_stop_token and many others to be available from more than just receivers. One of many is that an execution_policy should be able to provide all those things to the existing parallel algos.

ericniebler:
This is already fixed in the current draft of R4.

PR #280, opened 2021-11-25
Add customization points for controlling the forwarding of sender/receiver queries

### LINKED ISSUE #279: Query forwarding is underspecified
[execution.senders.adaptors.general]/p5 reads:

> Unless otherwise specified, all sender adaptors which accept a single sender argument return sender objects that propagate sender queries to that single sender argument. 

What does "sender queries" mean? Those CPOs listed in [execution.senders.queries]? What about user-defined sender queries? Or sender queries added to [execution.senders.queries] in future standards? How am I, as the author of a sender adaptor, to write this forwarding so it forwards all present and future sender queries, and only sender queries?

Note: simply forwarding all `tag_invoke` customizations is not an option here. That will _also_ forward algorithm implementations since they too use the `tag_invoke` mechanism.

Receiver queries have the same problem, but they can skate by by simply forwarding everything that isn't a receiver `set_*` CPO. But that too fails should we ever decide in the future to add another receiver channel (e.g., `set_next`).

I suggest the addition of a `tag_category_t<Tag>` trait that names the type `typename Tag::tag_category` if that exists, and `void` otherwise. Sender query CPOs can all identify using some `sender_query_tag`, which can be used when forwarding sender queries (and only sender queries).

ericniebler:
fixes #279 

This PR does the following:
1. Removes the `receiver` constraint on the arguments to `get_scheduler`, `get_allocator`, and `get_stop_token`.
2. Adds CPOs for controlling what queries get forwarded through adaptors: `forwarding_[scheduler|receiver|sender]_query`.
3. Precisely specify what queries get forwarded through sender and receiver adaptors.

### DIFF
--- a/include/execution.hpp
+++ b/include/execution.hpp
@@ -522,9 +522,8 @@ namespace std::execution {
           // Pass through receiver queries
           template <__receiver_query _CPO, class... _As>
             requires invocable<_CPO, const _Receiver&, _As...>
-          friend auto tag_invoke(_CPO cpo, const __promise& __self, _As&&... __as)
-            noexcept(is_nothrow_invocable_v<_CPO, const _Receiver&, _As...>)
-            -> invoke_result_t<_CPO, const _Receiver&, _As...> {
+          friend decltype(auto) tag_invoke(_CPO cpo, const __promise& __self, _As&&... __as)
+            noexcept(is_nothrow_invocable_v<_CPO, const _Receiver&, _As...>) {
             return ((_CPO&&) cpo)(as_const(__self.__rcvr_), (_As&&) __as...);
           }
 
@@ -1242,12 +1241,11 @@ namespace std::execution {
             execution::connect(((_Self&&) __self).base(), (_Receiver&&) __rcvr);
           }
 
-          template <__sender_query _Tag>
-            requires invocable<_Tag, const _Base&>
-          friend auto tag_invoke(_Tag __tag, const _Derived& __self)
-            noexcept(is_nothrow_invocable_v<_Tag, const _Base&>)
-            -> invoke_result_t<_Tag, const _Base&> {
-            return ((_Tag&&) __tag)(__self.base());
+          template <__sender_query _Tag, class... _As>
+            requires invocable<_Tag, const _Base&, _As...>
+          friend decltype(auto) tag_invoke(_Tag __tag, const _Derived& __self, _As&&... __as)
+            noexcept(is_nothrow_invocable_v<_Tag, const _Base&, _As...>) {
+            return ((_Tag&&) __tag)(__self.base(), (_As&&) __as...);
           }
 
          protected:
@@ -1352,9 +1350,8 @@ namespace std::execution {
 
           template <__receiver_query _Tag, class _D = _Derived, class... _As>
             requires invocable<_Tag, __base_t<const _D&>, _As...>
-          friend auto tag_invoke(_Tag tag, const _Derived& __self, _As&&... __as)
-            noexcept(is_nothrow_invocable_v<_Tag, __base_t<const _D&>, _As...>)
-            -> invoke_result_t<_Tag, __base_t<const _D&>, _As...> {
+          friend decltype(auto) tag_invoke(_Tag tag, const _Derived& __self, _As&&... __as)
+            noexcept(is_nothrow_invocable_v<_Tag, __base_t<const _D&>, _As...>) {
             return ((_Tag&&) tag)(__get_base(__self), (_As&&) __as...);
           }
 
@@ -1762,9 +1759,8 @@ _PRAGMA_POP()
 
           template <__receiver_query _Tag, class... _As>
               requires invocable<_Tag, const _Receiver&, _As...>
-            friend auto tag_invoke(_Tag tag, const __receiver& __self, _As&&... __as)
-                noexcept(is_nothrow_invocable_v<_Tag, const _Receiver&, _As...>)
-                -> invoke_result_t<_Tag, const _Receiver&, _As...> {
+            friend decltype(auto) tag_invoke(_Tag tag, const __receiver& __self, _As&&... __as)
+                noexcept(is_nothrow_invocable_v<_Tag, const _Receiver&, _As...>) {
               ((_Tag&&) tag)(__self.base(), (_As&&) __as...);
             }
 
@@ -2312,9 +2308,9 @@ _PRAGMA_POP()
                 }
 
                 template <__receiver_query _Tag, class... _Args>
-                friend auto tag_invoke(_Tag tag, const __receiver2& __self, _Args&&... __args)
-                  noexcept(is_nothrow_invocable_v<_Tag, const _Receiver&, _Args...>)
-                  -> invoke_result_t<_Tag, const _Receiver&, _Args...> {
+                  requires invocable<_Tag, const _Receiver&, _Args...>
+                friend decltype(auto) tag_invoke(_Tag tag, const __receiver2& __self, _Args&&... __args)
+                  noexcept(is_nothrow_invocable_v<_Tag, const _Receiver&, _Args...>) {
                   return ((_Tag&&) tag)(as_const(__self.__op_state_->__rcvr_), (_Args&&) __args...);
                 }
               };
@@ -2349,9 +2345,9 @@ _PRAGMA_POP()
                 }
 
                 template <__receiver_query _Tag, class... _Args>
-                friend auto tag_invoke(_Tag tag, const __receiver1& __self, _Args&&... __args)
-                  noexcept(is_nothrow_invocable_v<_Tag, const _Receiver&, _Args...>)
-                  -> invoke_result_t<_Tag, const _Receiver&, _Args...> {
+                  requires invocable<_Tag, const _Receiver&, _Args...>
+                friend decltype(auto) tag_invoke(_Tag tag, const __receiver1& __self, _Args&&... __args)
+                  noexcept(is_nothrow_invocable_v<_Tag, const _Receiver&, _Args...>) {
                   return ((_Tag&&) tag)(as_const(__self.__op_state_->__rcvr_), (_Args&&) __args...);
                 }
               };
--- a/include/execution.hpp
+++ b/include/execution.hpp
@@ -1844,6 +1844,13 @@ _PRAGMA_POP()
               };
             }
 
+          template <__sender_query _Tag, class... _As>
+            requires invocable<_Tag, const _Sender&, _As...>
+          friend decltype(auto) tag_invoke(_Tag __tag, const __sender& __self, _As&&... __as)
+            noexcept(is_nothrow_invocable_v<_Tag, const _Sender&, _As...>) {
+            return ((_Tag&&) __tag)(__self.__sndr_, (_As&&) __as...);
+          }
+
           _Sender __sndr_;
           _Fun __fun_;
         };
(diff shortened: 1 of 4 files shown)


include/functional.hpp
```diff
@@ -38,13 +38,27 @@ namespace std {
...
 
   template<auto& Tag>
   using tag_t = decay_t<decltype(Tag)>;
 
+  template<class Tag>
+  using tag_category_t = decltype(__impl::__tag_category<Tag>());
+
+  template<class Tag, class... Categories>
+  concept tag_category_one_of =
+    (derived_from<tag_category_t<Tag>, Categories> ||...);
```
miscco, line 60: I really do not like this

We should have a `receiver_querry` and a `sender_querry` concept.

They should simply inspect whether the alias is there and if so whether it is the right one. Do you ha e examples where we would need this dance?

miscco: Edit, it seems one needs to pull the `same_as` out but yeah...
```cpp
template<class CPO>
concept receiver_query = requires {
    typename CPO::tag_category;
} && std::same_as<typename CPO::tag_category, query_tag>;
```

https://godbolt.org/z/raar7nsxG

griwes: I like this suggestion better; it makes constraining the forwarding tag_invoke overloads easier than what you have in the PR right now.

miscco: Surprisingly just `same_as` seems to also work even of there is no `tag_category`
https://godbolt.org/z/s3f4xnzoh

ericniebler:
@kirkshoop has strong feelings about how best to approach this, and a categorization for tags ain't it. Making this a draft until we can do a proper think. It will miss R3.

kirkshoop:
To leave a record in the issue: here is the summary. 

A requirement I have is that I can create my own concept and have a subset of std library cpos forwarded through type-erasure/expressions/etc.. of that new concept.

A Cpo member is the wrong scope for determining forwarding by concept. The Cpo does not have an exhaustive list of concepts. 

Categories of CPOs have nothing to do with a forwarding query. If a cpo is in a type-erasure category, but not a receiver category, how would a generic type-erasure object, provided a list of cpos that satisfies the receiver concept, decide to forward that cpo. 

This example is too convoluted. 

It is the underlying conceptual mapping that I am trying to examine. Conceptually, outside this context, a category is too general to make a specific decision about how things in the category are treated. This is why role-based security systems categorize people and assets and then apply rules to control access. The rules apply different meaning for assets in a category depending on the categories that the person belongs to. 

forward_cpo_through_reciever() is a rule that applies forwarding from the category of a single cpo to a category for receiver implementations that forward. 

If there were tags for concepts, then a more general rule system could be applied such as forward(concept_tag, cpo)

kirkshoop:
See also #271

ericniebler:
@kirkshoop wrote:

> A requirement I have is that I can create my own concept and have a subset of std library cpos forwarded through type-erasure/expressions/etc.. of that new concept.

I've been thinking about this. If I were writing a generic library that had its own concepts and adapters then I would specify a concept or a trait for specifying what CPOs to forward through the adapters. If I want standard CPOs to be part of that list, I agree that tag categories don't cut it.

But I'm not trying to dictate how third parties should controlling forwarding. They can use traits or a separate CPO to control which CPOs get forwarded for which concepts. That's fine. In this PR I'm saying that for the scheduler/sender/receiver concepts, tags are how you opt into the "query that should forward" concept.

> If a cpo is in a type-erasure category, but not a receiver category, how would a generic type-erasure object, provided a list of cpos that satisfies the receiver concept, decide to forward that cpo.

I'm having trouble parsing this. What is a type-erasure category?

So, I hear you saying that a CPO for opting in (or out) of the "should forward by default for <concept>" concept is more flexible and more general. I agree. But I'm not (yet) sure I agree it's the right solution for, say, queries that should be forwarded through receiver adaptors.

I can't explain it, but I feel we need to be judicious with CPOs. In R3 I'm already adding one: `as_awaitable`. I strongly resisted that -- ask @lewissbaker -- but ultimately felt the return was too great, and there's no other way I can see. I'm afraid if we say "you control what CPOs are forwarded by defining a new CPO," that it might land poorly.

@miscco wrote:

> More explicitly what is wrong with
> 
> ```c++
> template<class CPO>
> concept receiver_query = requires {
>     typename CPO::tag_category;
> } && std::same_as<typename CPO::tag_category, query_tag>;
> ```

We could certainly add a `receiver_query` concept, but I would like it to be expressed in terms of `tag_category_one_of`, or at least `std::derived_from` instead of `std::same_as`.

I am foreseeing a need for CPOs that belong to more than one category. Such a CPO might define:

```c++
using tag_category = inherit< receiver_query_category, some_other_category >;
```

Maybe inheritance isn't the right mechanism for this, I don't know. And maybe Kirk ultimately convinces me that we really need another CPO for this, but I'm not there yet.

miscco:
> I am foreseeing a need for CPOs that belong to more than one category. Such a CPO might define:
> 
> ```c++
> using tag_category = inherit< receiver_query_category, some_other_category >;
> ```
> 
> Maybe inheritance isn't the right mechanism for this, I don't know. And maybe Kirk ultimately convinces me that we really need another CPO for this, but I'm not there yet.

I feel that in this case we are back to inline variables:

```cpp
template<class T>
static inline constexpr bool forwards_receiver_querries = false;

template<class T>
static inline constexpr bool forwards_sender_querries = false;
```

Those you can easily compose as you wish but with types I believe the additional complexity of making is somehow work is not worth it.

ericniebler:
> I feel that in this case we are back to inline variables:

That's reasonable and would meet my needs. I _think_ it also meets @kirkshoop's. Kirk?

kirkshoop:
My objection to those is mostly aesthetic. 

I never felt that opening a namespace, that I do not own, to specialize a type (or var or alias), that I do not own, was ok. It violates my understanding of scoping/partitioning/layering for isolated component design. 

I have used the return type of function declarations (that have no definition) to replace traits in my code for years. 

To me tag_invoke is just a better way to declare a scoped name, that can be customized without re-opening the scope, than the other options available. 

I tend to expect that compilers will improve their handling of an idiom once it starts impacting them. I care about perf, but not over good design.

miscco:
> My objection to those is mostly aesthetic.
> 
> I never felt that opening a namespace, that I do not own, to specialize a type (or var or alias), that I do not own, was ok. It violates my understanding of scoping/partitioning/layering for isolated component design.
> 
> I have used the return type of function declarations (that have no definition) to replace traits in my code for years.
> 
> To me tag_invoke is just a better way to declare a scoped name, that can be customized without re-opening the scope, than the other options available.
> 
> I tend to expect that compilers will improve their handling of an idiom once it starts impacting them. I care about perf, but not over good design.

I am not fully sure I understand the objection.
```cpp
struct with_static_member {
     static inline constexpr bool is_receiver_querry = true;
};

struct with_extern_specialization {};

template<>
static inline constexpr bool is_receiver_querry<with_extern_specialization > = true;
```

There is no difference between those two types. The only difference is that the latter is what most developers know about and what provides a well defined default value.

The issue with the former is that it complicates checking
```cpp
template <class T>
concept is_receiver_querry = requires {
    T:typename is_receiver_querry;
} && T::is_receiver_querry;
```

One provides a default the other not. There is no other difference between the two (Same for type aliases)

kirkshoop:
> Either the machinery of the framework has to "invade" the user types scope or the other way around

ADL does not require namespace invasion. Thus function overloading is the best way to find customizations imo. 

If I add the namespaces:

```
// in the <execution> header

namespace std::execution {
template<typename T>
static inline constexpr bool is_receiver_query<T> = false;
}

// in third_party.hpp

struct with_extern_specialization {};

namespace std::execution {
// opening std in third-party feels incorrect to me. 
template<>
static inline constexpr bool is_receiver_query<with_extern_specialization > = true;
}

```

Vs function customization

```
// in the <execution> header

namespace std::execution {
template<typename T>
static inline constexpr bool is_receiver_query_v = is_receiver_query(T{}); // T{} is exposition only
// Allows defaults
template<typename T> 
constexpr inline bool is_receiver_query() { return false; }
}

// in third_party.hpp

struct with_extern_specialization {
private:
// extern defined inside 
constexpr friend bool is_receiver_query(with_extern_specialization&) { return true; }
};

// or extern defined outside 
constexpr bool is_receiver_query(with_extern_specialization&) { return true; }

```

ericniebler:
OK @kirkshoop and @miscco, this one is ready for another look.

include/execution.hpp
```diff
@@ -313,49 +318,76 @@ namespace std::execution {
...
+  // [execution.receivers.queries], receiver queries
+  inline namespace __receiver_queries {
+    namespace __impl {
+      struct forwarding_receiver_query_t {
+        template <class _Tag>
+        constexpr bool operator()(_Tag __tag) const noexcept {
+          if constexpr (tag_invocable<forwarding_receiver_query_t, _Tag>) {
+            return tag_invoke(*this, (_Tag&&) __tag);
+          } else {
+            return __none_of<_Tag, set_value_t, set_error_t, set_done_t>;
```
kirkshoop, line 376: I expected this to return false if there is no customization- this looks like it returns true. Am I confused?

Edit: Ah I see that it is replacing the previous usage of the none_of expression. This would mean that every cpo defaults to being forwarded on a receiver and the only customizations are to prevent being forwarded on a receiver.

I am conflicted but don't have a strong reason to flip the excludelist to a includelist. I have a discomfort with excludelists from my time in Windows security. They are generally considered a mistake in the security space.

include/execution.hpp
```diff
@@ -599,23 +626,49 @@ namespace std::execution {
...
 
-  template <__one_of<set_value_t, set_error_t, set_done_t> _CPO>
-    inline constexpr get_completion_scheduler_t<_CPO> get_completion_scheduler{};
+      struct forwarding_sender_query_t {
+        template <class _Tag>
+        constexpr bool operator()(_Tag __tag) const noexcept {
+          if constexpr (tag_invocable<forwarding_sender_query_t, _Tag>) {
+            return tag_invoke(*this, (_Tag&&) __tag);
+          } else {
+            return false;
```
kirkshoop, line 652: Hmm - here we use a default of false which makes these customizations an includelist. That is inconsistent with the receiver queries cpo.

kirkshoop:
I like these changes. I hope that you are happy with them as well.

include/execution.hpp
ericniebler: Yes, I think false is the right default for senders, and true is the right default for receivers.

ericniebler:
> I like these changes. I hope that you are happy with them as well.

You know, I think I still prefer a simple boolean variable template. But I'm interested in hearing more opinions.

std_execution.bs
```diff
@@ -2641,13 +2641,33 @@ namespace std::execution {
...
   // [execution.schedulers.queries], scheduler queries
   enum class forward_progress_guarantee;
   inline namespace <i>unspecified</i> {
+    struct forwarding_scheduler_query_t;
+    inline constexpr forwarding_scheduler_query_t forwarding_scheduler_query{};
+
+    template &lt;class T>
+      concept <i>forwarding-scheduler-query</i> = // exposition only
+        requires(T t) { requires forwarding_scheduler_query(t); };
+
```
kirkshoop, line 2670: I do not see an implementation of forwarding scheduler query. Did I miss it?

kirkshoop (approved):

std_execution.bs
ericniebler: I haven't implemented any scheduler queries yet.

kirkshoop: I would expect the forwarding cpo to be implemented, and used by scheduler implementations to forward cpos, even if there are no cpos to forward in the proposal.

PR #291, opened 2021-11-30
Refine when receiver adaptor forwards completion to inner receiver

ericniebler:
I had a surprise recently. I wrote something morally equivalent to this:

```c++
struct __receiver : receiver_adaptor<__receiver, inner_receiver> {
  template <class... As>
    requires some_condition<As...>
  void set_value(As&&... as) && .......
...
};
```

My intention here had been that if `some_condition<As...>` is not satisfied, that `__receiver` not model `receiver_of<As...>`. But what actually happened is this: `receiver_adaptor` looks for a named member it can use. It doesn't find one (the constraint failed), so it tries to pass the `set_value` signal through to `inner_receiver`.

By some fluke in my case, `inner_receiver` happened to model `receiver_of<As...>` so the set_value signal got passed through, and `__receiver` appeared to model `receiver_of<As...>`, but it behaved strangely.

This PR tweaks `receiver_adaptor` so that if the derived type defines _any_ member `set_value`, regardless of whether it accepts the arguments or not, the forwarding to the inner receiver is disabled. It achieves this with name hiding by defining `set_xxx` type aliases. If `Derived::set_value` names a type (for instance), then we know that there is no named member function `set_value`.

cc'ing @kirkshoop for his thoughts.

### DIFF
--- a/include/execution.hpp
+++ b/include/execution.hpp
@@ -1122,6 +1122,9 @@ namespace std::execution {
       struct __receiver_adaptor {
         class __t : __adaptor_base<Base> {
           friend Derived;
+          using set_value = void;
+          using set_error = void;
+          using set_done = void;
 
           template <class D>
             static decltype(auto) __get_base(D&& self) noexcept {
@@ -1142,11 +1145,11 @@ namespace std::execution {
           }
 
           template <class D = Derived, class... As>
-            requires (!__has_set_value<Derived, As...>) &&
+            requires requires {typename D::set_value;} &&
               receiver_of<__base_t<D>, As...>
           friend void tag_invoke(set_value_t, Derived&& self, As&&... as)
             noexcept(nothrow_receiver_of<__base_t<D>, As...>) {
-            set_value(__get_base((Derived&&) self), (As&&) as...);
+            execution::set_value(__get_base((Derived&&) self), (As&&) as...);
           }
 
           template <class E>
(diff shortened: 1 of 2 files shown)


kirkshoop:
Very clever 😁 

Is this ultimately because of the extra work to allow the receiver to be stored in the adapter?

Actually, now I seem to remember the the forwarding behavior was intended in order for something like then() to not define set_error or set_done because they did not impact its single concern. 

How does this interact with the then() scenario?

Is it better to remove the storage from the adapters entirely (to simplify everything and ensure there are not more edge cases to fix)?

ericniebler:
> Is this ultimately because of the extra work to allow the receiver to be stored in the adapter?

No, it has nothing to do with that.

> Actually, now I seem to remember the the forwarding behavior was intended in order for something like then() to not define set_error or set_done because they did not impact its single concern.

Yes, that's right.

> How does this interact with the then() scenario?

It fixes a bug in the reference implementation of `then`. If the member `set_value()` is ill-formed, `receiver_adaptor` was erroneously looking to pass the `set_value` call to the inner receiver.

The alternative to this diff is to take away the automatic forwarding of the `set_xxx` signals to the inner receiver and require all derived types implement all three signals. This PR is my preference.

ericniebler:
@kirkshoop Any more comments?

kirkshoop (approved):

PR #299, opened 2021-12-03
Elaborate on the history of potentially-eager in the paper.

griwes:


std_execution.bs
```diff
@@ -1325,6 +1325,25 @@ executing their logic eagerly; <i>i.e.</i>, before the returned sender has been
 connected to a receiver and started. These algorithms were removed because eager
 execution has a number of negative semantic and performance implications.
 
+We have originally included this functionality in the paper because of along-standing
```
ericniebler, line 1328:
```suggestion
We have originally included this functionality in the paper because of a long-standing
```

ericniebler (commented):
LGTM, but I don't have a bead on whether this is enough detail to satisfy other accelerator vendors concerned about eager execution.

PR #301, opened 2021-12-03
strengthen requirements on typed_senders

ericniebler:
The current spec for `sender_traits` says the program is ill-formed if it _invokes_ a `set_*` overload that doesn't correspond to the types reported by the sender traits, but that's not strong enough. Even if it's in a branch that never executes at runtime, it would be problematic to even just _form_ a `set_*` expression that downstream code isn't expecting because it could cause hard errors during template instantiation.

Deal with this problem by replacing "invokes" with "odr-uses".

### DIFF
--- a/std_execution.bs
+++ b/std_execution.bs
@@ -3194,13 +3194,13 @@ enum class forward_progress_guarantee {
     </pre>
 
 6. If `value_types_of_t<S, Tuple, Variant>` for some sender type `S` is well formed, it shall be a type <code>Variant&lt;Tuple&lt;Args<sub>0</sub>...>, Tuple&lt;Args<sub>1</sub>...>, ..., Tuple&lt;Args<sub>N</sub>...>>></code>, where the type packs <code>Args<sub>0</sub></code> through <code>Args<sub>N</sub></code> are the packs of types the sender `S` passes as
-    arguments to `execution::set_value` after a receiver object. If such sender `S` invokes `execution::set_value(r, args...)` for some receiver `r`, where `decltype(args)` is not one of the type packs <code>Args<sub>0</sub></code> through <code>Args<sub>N</sub></code>, the program is ill-formed with no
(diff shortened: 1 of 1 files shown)


std_execution.bs
```diff
@@ -3194,13 +3194,13 @@ enum class forward_progress_guarantee {
     </pre>
 
 6. If `value_types_of_t<S, Tuple, Variant>` for some sender type `S` is well formed, it shall be a type <code>Variant&lt;Tuple&lt;Args<sub>0</sub>...>, Tuple&lt;Args<sub>1</sub>...>, ..., Tuple&lt;Args<sub>N</sub>...>>></code>, where the type packs <code>Args<sub>0</sub></code> through <code>Args<sub>N</sub></code> are the packs of types the sender `S` passes as
-    arguments to `execution::set_value` after a receiver object. If such sender `S` invokes `execution::set_value(r, args...)` for some receiver `r`, where `decltype(args)` is not one of the type packs <code>Args<sub>0</sub></code> through <code>Args<sub>N</sub></code>, the program is ill-formed with no
+    arguments to `execution::set_value` after a receiver object. If such sender `S` odr-uses ([basic.def.odr]) `execution::set_value(r, args...)` for some receiver `r`, where `decltype(args)...` is not one of the type packs <code>Args<sub>0</sub>...</code> through <code>Args<sub>N</sub>...</code>, the program is ill-formed with no
```
ericniebler, line 3197: FWIW, I think requiring _exact_ type matching here seems over-strict to me. If a sender reports that it sends an `int` but actually sends an `int&` or `const int&`, IFNDR seems a bit harsh. I'm also curious if we really mean `decltype(args)` here instead of `decltype((args))`. Only questions, no answers at this point.

griwes: Right, but if it reports that it sends `int &` but actually sends an rvalue, that's problematic, right?

ericniebler: Agreed. I haven't yet come up with a way of matching types here that says what we want it to say, mostly because I don't know yet exactly what we want to say.

ericniebler: I've filed this as #302 so we don't lose it.

PR #305, opened 2021-12-06
specify that all standard operation states are non-movable types

ericniebler:


kirkshoop (approved):

std_execution.bs
```diff
@@ -3066,6 +3066,8 @@ enum class forward_progress_guarantee {
         };
     </pre>
 
+2. Any operation state types defined by the implementation are non-movable types.
+
```
lewissbaker, line 3070: Do we need to require that they are non-movable (i.e. have a deleted move-constructor)?
Or can we just not say they are guaranteed to have a move constructor (or maybe add a non-normative note to this effect)?

villevoutilainen: It's a portability guarantee. Non-movable as in completely immovable, no move ctor, no copy ctor either. It's like a mutex or an atomic, you can use its memory address and stability thereof for useful purposes. If some operation_states suddenly are movable, they no longer provide those guarantees, and they don't work where other operation_states do.

Nota Bene: "non-movable" is not a term in the standard. At least I can't find it in the current draft. We could just as well say that operation states defined by the implementation do not satisfy copy_constructible or move_constructible.

PR #307, opened 2021-12-08
Implement start_detached

ericniebler:
For the sake of completeness.

### DIFF
--- a/include/execution.hpp
+++ b/include/execution.hpp
@@ -21,6 +21,7 @@
 #include <cassert>
 #include <condition_variable>
 #include <stdexcept>
+#include <memory>
 #include <mutex>
 #include <optional>
 #include <tuple>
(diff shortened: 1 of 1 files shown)


kirkshoop (approved):

include/execution.hpp
```diff
@@ -838,50 +839,81 @@ namespace std::execution {
...
+            __operation* __op_state_;
+            // Forward all the receiver ops, and delete the operation state.
+            template <__one_of<set_value_t, set_error_t, set_done_t> _Tag, class... _As>
+              requires invocable<_Tag, _Receiver, _As...>
+            friend auto tag_invoke(_Tag __tag, __receiver&& __self, _As&&... __as)
+                noexcept(is_nothrow_invocable_v<_Tag, _Receiver, _As...>)
+                -> invoke_result_t<_Tag, _Receiver, _As...> {
+              // Delete the state as cleanup:
+              unique_ptr<__operation> __g{__self.__op_state_};
+              return __tag((_Receiver&&) __self.__op_state_->__rcvr_, (_As&&) __as...);
```
lewissbaker, line 859: This implementation has a bug in that if the call to `set_value` throws then the operation-state is destroyed and a subsequent call to, say `set_error()` will access a dangling receiver/op-state.

include/execution.hpp
```diff
@@ -838,50 +839,81 @@ namespace std::execution {
...
+      void operator()(_Sender&& __sndr) const
+        noexcept(nothrow_tag_invocable<start_detached_t, _Sender>) {
+        (void) tag_invoke(start_detached_t{}, (_Sender&&) __sndr);
       }
-    } submit {};
-  }
+      template <sender _Sender>
+        requires (!tag_invocable<start_detached_t, _Sender>) &&
+          sender_to<_Sender, __impl::__detached_receiver>
+      void operator()(_Sender&& __sndr) const noexcept(false) {
```
lewissbaker, line 912: I know we are planning to remove support for this in favour of `async_scope` so may be it's not a big issue, but if we were to keep `start_detached()` there should be a way to support custom-allocator for the heap-allocation of the op-state here. e.g. see the `unifex::submit()` implementation.

PR #311, opened 2021-12-09
implement done_as_error and done_as_optional

ericniebler:


include/execution.hpp
```diff
@@ -1922,17 +1918,60 @@ _PRAGMA_POP()
...
+    template <__movable_value _Error, sender _Sender>
+      auto done_as_error(_Sender&& __sndr, _Error __err = _Error{}) {
+        return __done_as_error((_Sender&&) __sndr, (_Error&&) __err);
+      }
+
+    template <__movable_value _Error>
+      auto done_as_error(_Error __err) noexcept {
+        return __done_as_error((_Error&&) __err);
+      }
+  }
```
kirkshoop, line 1973: why is done_as_error() an overload set while done_as_optional is a function object? 
I would have expected both to be function objects.

kirkshoop (approved):

include/execution.hpp
ericniebler: It support the syntax: `done_as_error<Error>(sender)`, which default-constructs the error. Not terribly compelling, is it?

kirkshoop: No, I do not find that compelling.

PR #316, opened 2021-12-12
Add example for let_* functions

lucteo:
Add both C++ code and bikeshed wording.

More examples with the server theme will follow.

### DIFF
--- a/examples/server_theme/let_value.cpp
+++ b/examples/server_theme/let_value.cpp
@@ -65,21 +65,6 @@ struct http_response {
   std::string body_;
 };
 
-// TODO: remove this when moving to start_detached
-struct my_receiver {
-  template <typename... Vs>
-  friend void tag_invoke(ex::set_value_t, my_receiver, Vs&&... vals) {
-    // std::cout << "completed successfully\n";
-  }
-  template <typename E>
-  friend void tag_invoke(ex::set_error_t, my_receiver, E&& err) noexcept {
-    // std::cout << "completed with error\n";
-  }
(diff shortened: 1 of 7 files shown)


ericniebler:
These look good to me. Thanks, @lucteo. @kirkshoop, any thoughts?

examples/server_theme/let_value.cpp
```diff
@@ -0,0 +1,153 @@
...
+        | ex::let_error(error_to_response)
+        // If the flow is cancelled, send back a proper response
+        | ex::let_done(done_to_response)
+        // write the result back to the client
+        | ex::let_value(send_response)
+        // done
+        ;
+
+    // execute the whole flow asynchronously
+    ex::start_detached(std::move(snd));
```
kirkshoop, line 147: How do these examples exit safely?

The following request_stop call will cancel the work- but I do not see anything that waits for the work. 

This scenario is something that we would like to support with an async_scope in a forthcoming paper. 

For now there should be some way to wait for the pool to empty.

kirkshoop (commented):
I love these samples. Especially the http example.

examples/server_theme/let_value.cpp
kirkshoop: Never mind. I have trouble keeping track of all the static_thread_pool impls. I see that this one blocks in the destructor, which I deeply dislike.

kirkshoop (approved):
These will need to be used to show the benefits of async_scope.

PR #317, opened 2021-12-12
Wording defect on the "transfer" algorithms w.r.t. get_completion_scheduler

lucteo:
The working implies that get_completion_scheduler for all 3 CPOs will
return a scheduler equivalent to the one received as an argument. But
this can be only guaranteed in the case of set_value_t and set_done_t,
and cannot be guaranteed in case of set_error.

If an exception is thrown while trying to use the given scheduler to
schedule work, then the specification indicates that `set_error` is
called directly, i.e., so it may not be on the given scheduler.

This fix applies to `transfer`, `schedule_when` and `transfer_when_all`

### DIFF
--- a/include/execution.hpp
+++ b/include/execution.hpp
@@ -2571,7 +2571,7 @@ namespace std::execution {
             return {__self.__sched_, ((_Self&&) __self).__sndr_, (_Receiver&&) __rcvr};
           }
 
-          template <__one_of<set_value_t, set_error_t, set_done_t> _Tag>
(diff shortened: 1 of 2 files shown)


ericniebler:
Would you please add a note to the revision history for R4 about this change? Otherwise, looks good.

lucteo:
Done.

ericniebler:
Oh, just one more thing. Can you update the reference implementation to conform to the new wording? It's a simple change.

PR #323, opened 2021-12-19
update `completion schedulers` to `Completion Schedulers`

kshitij12345:
`Completion Schedulers` is first introduced in this section. Without capitalization it's hard to understand from the heading that it is a new concept. At first glance it is easy to think that `Senders can propagate completion to schedulers`?


kshitij12345:
Gentle ping @ericniebler :)
Please take a look to see if it makes sense (Should be easy review)

Thanks!

ericniebler:
The standard doesn't generally capitalize technical terms. They're generally lowercase and in normal font, apart from where the terms are defined, in which case their in *italics*. But italics would look odd in a section title, so leaving as-is.

PR #326, opened 2021-12-20
add completion_signatures: a utility for defining sender traits

ericniebler:
This adds a `completion_signatures` utility for defining sender traits. Usage is like:

```c++
using _traits = completion_signatures<
  set_value_t(Vals0...),
  set_value_t(Vals1...),
  set_value_t(Vals2...),
  ...
  set_error_t(Err0),
  set_error_t(Err1),
  ...
  set_done_t()
>;
```

This defines a struct with nested `value_types`, `error_types` and `sends_done` members that can be used to implement `sender_traits`, or as a base class when writing a typed sender.

I plan to add this to R4. Comments welcome.

### DIFF
--- a/include/execution.hpp
+++ b/include/execution.hpp
@@ -118,20 +118,59 @@ namespace std::execution {
       nothrow_tag_invocable<set_value_t, _Receiver, _As...>;
 
   /////////////////////////////////////////////////////////////////////////////
+  // receiver_signatures
+  // NOT TO SPEC
+  namespace __receiver_signatures {
+    template <same_as<set_value_t> _Tag, class _Ty = __q<__types>, class... _Args>
+      __types<__minvoke<_Ty, _Args...>> __test(_Tag(*)(_Args...));
+    template <same_as<set_error_t> _Tag, class _Ty = __q<__types>, class _Error>
+      __types<__minvoke1<_Ty, _Error>> __test(_Tag(*)(_Error));
+    template <same_as<set_done_t> _Tag, class _Ty = __q<__types>>
+      __types<__minvoke<_Ty>> __test(_Tag(*)());
+    template <class, class = void>
+      __types<> __test(...);
+
+    template <class _Sig, class _Tag, class _Ty = __q<__types>>
+      using __signal_args_t =
+        decltype(__test<_Tag, _Ty>((_Sig*) nullptr));
+
+    template <class _Sig>
+      concept __receiver_signal =
+        requires { typename __id<decltype(__test((_Sig*) nullptr))>; };
+
+    template <class... _Sigs>
+      struct receiver_signatures {
+        struct type {
+          template <template <class...> class _Tuple, template <class...> class _Variant>
+            using value_types =
+              __minvoke<
+                __concat<__q<_Variant>>,
+                __signal_args_t<_Sigs, set_value_t, __q<_Tuple>>...>;
+
+          template <template <class...> class _Variant>
+            using error_types =
+              __minvoke<
+                __concat<__q<_Variant>>,
+                __signal_args_t<_Sigs, set_error_t, __q1<__id>>...>;
+
+          static constexpr bool sends_done =
+            __minvoke<
+              __concat<__count>,
+              __signal_args_t<_Sigs, set_done_t>...>::value != 0;
+        };
+      };
+  } // namespace __receiver_signatures
+
+  template <__receiver_signatures::__receiver_signal... _Sigs>
+    using receiver_signatures =
+      __t<__receiver_signatures::receiver_signatures<_Sigs...>>;
+
+  /////////////////////////////////////////////////////////////////////////////
   // [execution.senders.traits]
   namespace __sender_base {
     struct sender_base {};
   }
   using __sender_base::sender_base;
-
-  template <bool _SendsDone, class... _Ts>
-    struct __sender_of {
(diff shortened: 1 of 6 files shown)


include/execution.hpp
```diff
@@ -117,22 +117,61 @@ namespace std::execution {
...
+    template <class, class = void>
+      __types<> __test(...);
+
+    template <class _Sig, class _Tag, class _Ty = __q<__types>>
+      using __signal_args_t =
+        decltype(__test<_Tag, _Ty>((_Sig*) nullptr));
+
+    template <class _Sig>
+      concept __receiver_signal =
+        requires { typename __id<decltype(__test((_Sig*) nullptr))>; };
```
lewissbaker, line 139: Won't this check also match the ellipsis-taking overload of `__test()` above?
Is this supposed to just be testing that the signature is one of the set_value/set_error/set_done signatures?

std_execution.bs
```diff
@@ -4555,6 +4567,71 @@ from the operation before the operation completes.
 
         - Otherwise, <code>execution::set_done(<i>GET-BASE</i>(static_cast&lt;Derived&amp;&amp;>(self)))</code>
 
+### `execution::receiver_signatures` <b>[execution.snd_rec_utils.rcvr_sigs]</b> ### {#spec-execution.snd_rec_utils.rcvr_sigs}
```
lewissbaker, line 4570: I wonder if `completion_signatures` would be a better name than `receiver_signatures`?

std_execution.bs
```diff
@@ -4555,6 +4567,71 @@ from the operation before the operation completes.
...
+    1. A type `Fn` satisfies <code><i>receiver-signature</i></code> if it is a function type with one of the following forms:
+
+        * <code>set_value_t(<i>Vs</i>...)</code>, where <code><i>Vs</i></code> is an arbitrary parameter pack.
+        * <code>set_error_t(<i>E</i>)</code>, where <code><i>E</i></code> is an arbitrary type.
+        * `set_done_t()`
+
+    2. Otherwise, `Fn` does not satisfy <code><i>receiver-signature</i></code>.
+
+4.  <pre highlight="c++">
+    template &lt;<i>receiver-signature</i>... Fns>
```
lewissbaker, line 4611: Should this be later extended to support listing other queries on the receiver?
e.g. that it needs to implement `get_scheduler()`?

std_execution.bs
```diff
@@ -4555,6 +4567,71 @@ from the operation before the operation completes.
...
+      using receiver_signagures = <i>see below</i>;
+    </pre>
+
+    Given a template parameter pack <code><i>Fns</i></code> of types that satisfy <code><i>receiver-signature</i></code>, <code>receiver_signatures&lt;<i>Fns</i>...></code> names a non-template class type equivalent to the following:
```
lewissbaker, line 4615: > ... names a non-template class type ...

I will try to write up a paper proposing your "private" template parameters for class templates as a way of giving better control of ADL so we can avoid having to jump through hoops like this.

villevoutilainen: > > ... names a non-template class type ...
> 
> I will try to write up a paper proposing your "private" template parameters for class templates as a way of giving better control of ADL so we can avoid having to jump through hoops like this.

That's nice, but please refrain from using 'private' in that even as a stopgap working todo-spelling of the facility. It's a colossal distraction, considering that name lookup and access control are more or less completely orthogonal in C++.

include/execution.hpp
ericniebler: No, because the non-defaulted template parameter is non-deducible.

std_execution.bs
ericniebler: I like that.

std_execution.bs
ericniebler: Well, it's not really necessary. This utility is for providing properties of the _sender_ that are not otherwise inspectable. The only time you would need information about the receiver queries is when type-erasing the sender, in which case we could have a sister utility: `receiver_queries`.

std_execution.bs
ericniebler: Obviously, you're right, `private` is for controlling access, and it's helpful to know what to expect of we bring this to EWG.

Access control is, in part, for marking parts of a class as implementation details. And this is another time you want to mark a part of a class as an implementation detail. That's why I think `private` isn't the worst option.

But I'm curious if you have any other suggestions.

std_execution.bs
ericniebler: The other reason I would like to avoid putting receiver queries in here is that the syntax for the completion signals would have to change to something like this:

```c++
using _traits = completion_signatures<
  void(set_value_t, Vals0...),
  void(set_value_t, Vals1...),
  void(set_value_t, Vals2...),
  ...
  void(set_error_t, Err0) noexcept,
  void(set_error_t, Err1) noexcept,
  ...
  void(set_done_t) noexcpet
>;
```

... which I don't consider to be an improvement.

Also, then the name `completion_signatures` no longer makes sense. It would have to be something like `receiver_requirements` or some such, and even that would stop making sense if we go the direction of #320 and move the receiver queries into a separate context object.

lewissbaker (approved):
lgtm

Another thought on the naming of `completion_signatures` is `make_sender_traits<...>` (similar vein to `std::make_integer_sequence<int, 42>`).

PR #327, opened 2021-12-21
Add then and upon_* example

lucteo:
Add "server theme" example that uses then and upon_* to show how single-
threaded actions can be chained together with senders.

This chaining will properly react to errors and cancellations.

Also remove TODO in "on_transfer.cpp" example, as now the code works
fine.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -220,6 +220,7 @@ def_example(example.retry "examples/retry.cpp")
 def_example(example.then "examples/then.cpp")
 def_example(example.server_theme.let_value "examples/server_theme/let_value.cpp")
 def_example(example.server_theme.on_transfer "examples/server_theme/on_transfer.cpp")
+def_example(example.server_theme.then_upon "examples/server_theme/then_upon.cpp")
 
 ##############################################################################
 # Install targets ------------------------------------------------------------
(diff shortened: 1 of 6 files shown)


ericniebler:
Thanks for these! Part of me wants to wait until I've implemented the `upon_*` algorithms before committing this. Also, I've noticed that, although your examples build fine, if I change a header, your examples don't get rebuilt. I'm not GNUmake-savvy enough to grok why, though.

> Also remove TODO in "on_transfer.cpp" example, as now the code works
fine.

These examples exposed a couple of bugs that I've squashed. These extra examples are serving multiple purposes!

lucteo:
It was my pleasure.
The main purpose of the patch was to add the example in the .bs main text. The .cpp example is just a double-check on my side. I would prefer to get the text in before the next revision of the paper, but otherwise I'm fine merge it whenever it works best for you.
If you are planning a delay into merging this, probably I'll put the next 1-2 examples I had in mind on this PR as well.

On the makefile issue: I also find this annoying, but just a small annoyance. Do you want me to make a PR for fixing this issue? Also, in my experience I find CMake to be much better; should we switch to CMake?

ericniebler:
> On the makefile issue: I also find this annoying, but just a small annoyance. Do you want me to make a PR for fixing this issue? 

Yes, please.

> Also, in my experience I find CMake to be much better; should we switch to CMake?

If you want to fix this by using cmake in the `examples/` directory, that's ok with me. But leave the top-level `Makefile`, please.

ericniebler:
Needs rebase.

PR #330, opened 2021-12-22
Workaround concat for nvc++

senior-zero:
NVC++ fails to compile `__concat` implementation. This affects most of the senders. If the fix doesn't introduce any issues with clang, it would be nice to have it in the main branch.


ericniebler:
I decided to fix this problem more generally: 1a74218.

PR #332, opened 2021-12-22
Fix transfer_when_all

senior-zero:
Not checking for the opposite restriction of the previous overload (`tag_invocable<transfer_when_all_t, _Sched, _Senders...>`) leads to compilation errors:
```
more than one instance of overloaded function "std::execution::__when_all::transfer_when_all_t::operator()" matches the argument list
```

### DIFF
--- a/include/execution.hpp
+++ b/include/execution.hpp
@@ -3036,7 +3036,9 @@ namespace std::execution {
       }
 
       template <scheduler _Sched, typed_sender... _Senders>
-        requires (__impl::__zero_or_one_alternative<_Senders> &&...)
+        requires ((!tag_invocable<transfer_when_all_t, _Sched, _Senders...>) ||
+          (!sender<tag_invoke_result_t<transfer_when_all_t, _Sched, _Senders...>>)) &&
+          (__impl::__zero_or_one_alternative<_Senders> &&...)
       auto operator()(_Sched&& __sched, _Senders&&... __sndrs) const {
         return transfer(when_all((_Senders&&) __sndrs...), (_Sched&&) __sched);
       }


include/execution.hpp
```diff
@@ -3036,7 +3036,8 @@ namespace std::execution {
       }
 
       template <scheduler _Sched, typed_sender... _Senders>
-        requires (__impl::__zero_or_one_alternative<_Senders> &&...)
+        requires (!tag_invocable<transfer_when_all_t, _Sched, _Senders...>) &&
+                 (__impl::__zero_or_one_alternative<_Senders> &&...)
```
ericniebler, line 3040: This isn't quite right. I agree, there's an ambiguity since both overloads are constrained, and the constraints of one overload do not subsume the constraints of the other; that is, it's possible that a particular type satisfies the constraints of _both_ overloads, and the compiler has no way to choose which overload has the stronger set of constraints.

The first overload is constrained both that there is a `tag_invoke` overload _and_ that it returns a `sender`. We do that elsewhere, and I think it's a little user-hostile. If the user bothers to provide an overload for `transfer_when_all`, then falling back to the default implementation because it doesn't return a sender is likely hiding a bug. I would rather we changed all the algorithms to not _require_ that the returned type is a `sender`, but rather to _mandate_ it (in LEWG parlance); that is, turn the `sender<tag_invoke_result_t<....>>` constraint into a `static_assert`. I'm curious to hear @griwes's thoughts. But that's really not this PR's concern.

So for now, I would strictly negate the constraint of the first overload in the constraint of the second, and we can make a separate pass and turn all return type constraints into `static_assert`'s (if we all agree that's an improvement).

```suggestion
        requires ((!tag_invocable<transfer_when_all_t, _Sched, _Senders...>) ||
          (!sender<tag_invoke_result_t<transfer_when_all_t, _Sched, _Senders...>>)) &&
          (__impl::__zero_or_one_alternative<_Senders> &&...)
```

ericniebler (changes requested):

PR #333, opened 2021-12-22
Fix transfer_just

senior-zero:
Not checking for the opposite restriction of the previous overload (`tag_invocable<tag_invocable<transfer_just_t...`) leads to compilation errors:

```
error: more than one instance of overloaded function "std::execution::__transfer_just::transfer_just_t::operator()" matches the argument list
```

### DIFF
--- a/include/execution.hpp
+++ b/include/execution.hpp
@@ -2685,6 +2685,8 @@ namespace std::execution {
         return tag_invoke(*this, (_Scheduler&&) __sched, (_Values&&) __vals...);
       }
       template <scheduler _Scheduler, __movable_value... _Values>
+        requires (!tag_invocable<transfer_just_t, _Scheduler, _Values...> ||
+          !typed_sender<tag_invoke_result_t<transfer_just_t, _Scheduler, _Values...>>)
       auto operator()(_Scheduler&& __sched, _Values&&... __vals) const
         -> decltype(transfer(just((_Values&&) __vals...), (_Scheduler&&) __sched)) {
         return transfer(just((_Values&&) __vals...), (_Scheduler&&) __sched);


ericniebler:
The compiler is supposed to prefer constrained overloads to non-constrained ones during overload resolution. This is a workaround for a compiler bug. I would rather not add this to the reference implementation.

senior-zero:
Doesn't these
```
template <scheduler /* <------ */ _Scheduler, __movable_value /* <------ */ ... _Values>
auto operator()(_Scheduler&& __sched, _Values&&... __vals) const
```

make the second overload constrained as well?

senior-zero:
Here's [the reproducer](https://godbolt.org/z/fb16586GP). GCC, MSVC and Clang consider these overloads ambiguous. Am I missing anything?

ericniebler:
Huh, that's really strange. The compiler(s) don't seem to be using subsumption to tie-break the constrained overloads. Can you try this example with nvc++?

https://godbolt.org/z/KMh51c6Mx

villevoutilainen:
> Huh, that's really strange. The compiler(s) don't seem to be using subsumption to tie-break the constrained overloads. Can you try this example with nvc++?

Subsumption works with concepts. If you have fold-expressions in requires-clauses, you're using arbitrary expressions, and those don't subsume.

villevoutilainen:
> > Huh, that's really strange. The compiler(s) don't seem to be using subsumption to tie-break the constrained overloads. Can you try this example with nvc++?
> 
> Subsumption works with concepts. If you have fold-expressions in requires-clauses, you're using arbitrary expressions, and those don't subsume.

..oh, and Dr. Sutton says there's no subsumption of packs, at all.

include/execution.hpp
```diff
@@ -2681,6 +2681,8 @@ namespace std::execution {
         return tag_invoke(*this, (_Scheduler&&) __sched, (_Values&&) __vals...);
       }
       template <scheduler _Scheduler, __movable_value... _Values>
+        requires (!tag_invocable<transfer_just_t, _Scheduler, _Values...> &&
+          !typed_sender<tag_invoke_result_t<transfer_just_t, _Scheduler, _Values...>>)
```
ericniebler, line 2685: Shouldn't this be:
```suggestion
        requires (!tag_invocable<transfer_just_t, _Scheduler, _Values...> ||
          !typed_sender<tag_invoke_result_t<transfer_just_t, _Scheduler, _Values...>>)
```

senior-zero: I've missed this, thanks for noticing! Updated the commit.

senior-zero:
> Huh, that's really strange. The compiler(s) don't seem to be using subsumption to tie-break the constrained overloads. Can you try this example with nvc++?
> 
> https://godbolt.org/z/KMh51c6Mx

nvc++ compiles this without any issues.

PR #350, opened 2021-12-29
add heuristic for guessing whether a generally-awaitable type can complete with set_done

ericniebler:
It occurs to me that the default `sender_traits<>::sends_done` value for awaitables (`false`) is possibly incorrect. Some awaitables _may_ complete with a cancellation "exception", which should be translated to a call to `set_done`. We can get a pretty good idea if that's the case if `coroutine_traits<A>::promise_type` names a type that implements `unhandled_done()`. I think that can give both false positives _and_ false negatives, but I think 99% of the time it will be accurate.

@lewissbaker thoughts?

### DIFF
--- a/include/execution.hpp
+++ b/include/execution.hpp
@@ -124,6 +124,19 @@ namespace std::execution {
   }
   using __sender_base::sender_base;
 
+  template <class _Awaitable>
+    using __promise_type_of_t =
+      typename __coro::coroutine_traits<_Awaitable>::promise_type;
+
+  template <class _Awaitable>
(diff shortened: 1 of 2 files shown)


kirkshoop:
I prefer to avoid heuristics. In my experience they make a system much harder to reason about. 

Having algos that explicitly map an error type to a set_done signal seems like a better solution, since usage of the algo documents the behavior in the code and can be replaced with new algos over time as the set of desired behaviors becomes obvious.

ericniebler:
Then I think the right thing to do is to change the default value of `set_done` from false to true.

PR #353, opened 2021-12-29
Fix inlining for some in_place_stop_source functions

lucteo:
This prevents the "stop_token" header for being included in multiple
.cpp files.


lewissbaker (approved):
lgtm. Thanks for the fix!

Issue #294, opened 2021-12-01
with_delegee_scheduler CPO


LeeHowes:
We have a model for forward progress delegation such that:
```
sync_wait(foo());
```

will propagate sync_wait's internal scheduler to foo's sender, such that foo can delegate to it. This is important if foo is, say, `std::execution::async_sort`.

However, `on` and other scheduler-changing break this. Inside on we have a scheduler-provider at connect time, and we propagate a scheduler-provider upstream, but the schedulers they provide are different by definition.

We therefore need a hook that `on` would call to inject this delegation relationship back into the chain.

```
scheduler with_delegee_scheduler(scheduler delegant, scheduler delegee);
```

The default implementation would be a no op, returning `delegant` directly. Any scheduler that understands delegation, like the system thread pool, would implement this to do the appropriate hook ups internally, which would give the same delegation behaviour as we would want to see in:
```
on(delegee, foo(schedule(delegant)));
```
where foo builds a chain of senders coming out of the `schedule` sender, but does not call `on` internally.

PR #354, opened 2021-12-31
with_delegee_scheduler

LeeHowes:
Adds with_delegee_scheduler as a scheduler mutator to implement issue #294 . This CPO allows a scheduler to customise the behaviour when it is offered a second scheduler to delegate to. This behaviour makes customisation of delegation interactions independent of the behaviour of execution::on (and other future transferring operations). The PR also modifies execution::on to make use of this operation to propagate delegation through execution::on.

### DIFF
--- a/std_execution.bs
+++ b/std_execution.bs
@@ -3207,10 +3207,10 @@ enum class forward_progress_guarantee {
 
 1. `execution::with_delegee_scheduler` is used to offer a scheduler an alternative scheduler to delegate work to.
 
-2. The name `execution::with_delegee_scheduler` denotes a customization point object. For some subexpressions `s` and `s'`, let `S` be `decltype((s))` and let `S'` be `decltype((s'))`. If `S` does not satisfy `execution::scheduler` or `S'` does not satisfy `execution::scheduler`, `execution::with_delegee_scheduler` is ill-formed.
+2. The name `execution::with_delegee_scheduler` denotes a customization point object. For some subexpressions `s` and `s2`, let `S` be `decltype((s))` and let `S2` be `decltype((s2))`. If `S` does not satisfy `execution::scheduler` or `S'` does not satisfy `execution::scheduler`, `execution::with_delegee_scheduler` is ill-formed.
     Otherwise, `execution::with_delegee_scheduler(s, s')` is expression equivalent to:
 
-    1. `tag_invoke(execution::with_delegee_scheduler, as_const(s), s')`, if this expression is well formed and its type is `execution::with_delegee_scheduler`, and is `noexcept`.
+    1. `tag_invoke(execution::with_delegee_scheduler, as_const(s), s2)`, if this expression is well formed and its type is `execution::with_delegee_scheduler`, and is `noexcept`.
 
     2. Otherwise, `s`.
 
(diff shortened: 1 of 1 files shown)


ericniebler:
Could this need be satisfied with an additional receiver query, `get_delegee_scheduler()`? `sync_wait` would have both `get_scheduler()` and `get_delegee_scheduler()` return the `run_loop` scheduler. Algorithms like `on()` would change the scheduler returned by `get_scheduler()` but leave the `get_delegee_scheduler()` query alone, so that the `run_loop` scheduler is always available.

LeeHowes:
I think it could. It depends on what we want to achieve, really.

First, I think sync_wait is a distraction. It is true that we want to support that for boost blocking, but it probably isn't the only case. A GPU that can fall back to the CPU would be a reasonable second example.

I have been working on the basis that we want to be able to delegate through a chain. So `on(s1) | foo() | on(s2) | foo2() | on(s3) | foo3 | on(s4)...` (using piping for clarity of the chain) would delegate through. The actual example Kirk and I discussed here was to write: `async_scope.spawn(on(sync_wait_scheduler, on(thread_pool, foo()))` so the inner `on` gives a well-scheduled non-blocking task, and the outer gives the delegee. Clearly that can work though with `with_query_value` and either a new query or get_scheduler. We also don't really use `get_scheduler` yet, so this was a way to ensure it was used. 

On the other hand, if we have free get_scheduler() that can head a customisation chain, that's a very good use for get_scheduler anyway. Also, that GPU/CPU example is an interesting case where what you suggest may be cleaner in the sense that it allows us to have two separate ideas here. We can clearly propagate the delegee scheduler just as deeply if we want to, with new algorithms. In the GPU/CPU case it might make sense to write something like:

```
sender foo() {
  return get_scheduler() | bulk(...);
}

auto gpu_sender = on(gpu, foo());
sync_wait(gpu_sender);
```

with the meaning that bulk itself might decide to delegate and have access to the sync_wait scheduler, where in my PR it would only have access to the gpu_scheduler, limiting its options. My proposal was for the gpu_scheduler to make that delegating decision, but is entirely possible that something in bulk (or some more complex algorithm replacing bulk there) might want to make those decisions separately.

One risk of this though is that if an algorithm is allowed to delegate and schedule to its primary sender there is a risk of racey behaviour. If the scheduler itself makes all decisions about delegation then it can give correct ordering of work.

ericniebler:
It's not clear to me why a scheduler would be able to make intelligent decisions about what work to delegate and what not to. That's why I suggested the query. The algorithm author may have a better idea what work needs to be delegated.

LeeHowes:
I see it the other way around in most cases. The algorithm can't possibly know when to delegate. Think about the general boost blocking example:
```
sync_wait(sort(...));
```

If sort is implemented as a few `bulk` algorithms, which are customised, then. I can see those knowing how to delegate. Sort itself, though, has no way to communicate with the execution context to know if queues are full, if all of the threads are blocked or anything like that. It also can't really know if it should reissue work that is already in the thread pool short of injecting cancellation everywhere and being very careful that work isn't run twice.

The scheduler, or its context at least, certainly can know that it has no threads that can make progress, and as long as it maintains some sort of scheduling thread there as well or has some way to run occasional check up work it can decide to delegate whole queues. It also has control of the difference between work that is running and work that is in a queue
waiting for another part of the process to free the threads. The algorithm has no way to know that.

ericniebler:
OK, that makes sense, thanks. But still, if we had a `get_delegee_scheduler()` receiver query, the scheduler's sender could query it and make a decision at `start()` time about whether to delegate or not. So this still doesn't suggest to me it needs to be a scheduler adaptor.

LeeHowes:
Indeed, both approaches will work. It just becomes a policy question of whether we want separate queries for the concepts or whether we expect to delegate hierarchically.

LeeHowes:
@ericniebler #358 implements your proposed approach.

ericniebler:
Having tried both approaches now, do you have an opinion about which you like better?

LeeHowes:
It's hard to tell before we see delegation being used in practice. We know we need it but not necessarily how a system thread pool might want to use it. I think the query value is a simpler solution though in that it is consistent with other queries.

ericniebler:
Ok then I suggest we go with the simpler solution -- a receiver query -- until experience shows it to be insufficient.

LeeHowes:
Yes, agreed.

PR #355, opened 2022-01-01
Build system for tests & examples

lucteo:
Uses conan/cmake.


lucteo:
This works just fine with my AppleClang13. But, for strange reason it doesn't work with GCC-11 (although, it works when I import the P2300 code in concore). Also, it doesn't work with plain clang out of the box.

lucteo:
Build job is now checking tests & examples with clang-12 for every commit. I think I'm done with this PR. I'll focus next on adding unit tests.

ericniebler:
This is AWESOME and a huge help. Thank you @lucteo.

lucteo:
My pleasure.

BTW. On my library, I also have the following checks that are not turned on here:
- zero warnings for one compiler
- clang-format doesn't find any differences (I find it hard to work without clang-format these days)
- static checks (clang-tidy and cppcheck)
- sanitizers

Does it make sense to have them here as well?

ericniebler:
Hi @lucteo. I'm unfamiliar with this conan/cmake combo. I have been developing locally on Linux using clang++-12 with `-std=c++20 -stdlib=libc++`. I know how to specify the compiler and flags to cmake, but it seems that this command:

```bash
conan install .. --build=missing -s build_type=Release
```

is just using the system compiler and stdlib, which then causes cmake to error out due to a mismatch. Could you tell me how to make conan do my bidding?

Incidentally, problems like this is why I generally prefer using cmake itself to fetch dependencies. For instance, to get catch2:

```cmake
Include(FetchContent)

FetchContent_Declare(
  Catch2
  GIT_REPOSITORY https://github.com/catchorg/Catch2.git
  GIT_TAG        v3.0.0-preview3
)

FetchContent_MakeAvailable(Catch2)

add_executable(tests test.cpp)
target_link_libraries(tests PRIVATE Catch2::Catch2WithMain)
```

(taken from https://github.com/catchorg/Catch2/blob/devel/docs/cmake-integration.md)

lucteo:
Hi

First question: does `test-clang-12.sh` script work for you?
Hopefully the answer is _yes_. If that's the case, then that's one good way to run everything.
The main advantage is that is using exactly the same build steps performed by the CI, so it can be used to fix strange CI failures.
The disadvantage is that it's using docker, so it's not native and it tends to consume a bit more resources.

Running `test-clang-12.sh` will show you the actual commands that can be used to build the solution. Here is a copy-paste from running the script:
```
export CC=/usr/bin/clang-12 CXX=/usr/bin/clang++-12 && conan install "/github/workspace" --build=missing -s build_type=Release  -s compiler=clang -s compiler.version=12 -s compiler.libcxx=libc++
export CC=/usr/bin/clang-12 CXX=/usr/bin/clang++-12 && cmake   -DCMAKE_C_FLAGS=" " -DCMAKE_CXX_FLAGS=" " "/github/workspace"
cmake --build . -v -j 4
```
The key here is to expose the compiler as environment variables before calling `conan`.

Instead of specifying the conan settings to be used, you can also define a [conan profile](https://docs.conan.io/en/latest/reference/profiles.html). For example, on my mac, I've created `~/.conan/profiles/clang-12` with the following content:
```
[settings]
os=Macos
os_build=Macos
arch=x86_64
arch_build=x86_64
compiler=clang
compiler.version=12
compiler.libcxx=libc++
build_type=Release
[options]
[build_requires]
[env]
CC=/usr/local/Cellar/llvm@12/12.0.1_1/bin/clang
CXX=/usr/local/Cellar/llvm@12/12.0.1_1/bin/clang++
CFLAGS=-L/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/usr/lib
CXXFLAGS=-L/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/usr/lib
```

With that, I can simply run: `conan install <dir> --build=missing -pr clang-12`

Please note that you still need to tell CMake what compiler you are using. Unfortunately, CMake doesn't pick up the compiler that conan uses. You can do that by using environment variables (like shown above), or by passing specifying the compiler as CMake parameters: `cmake -DCMAKE_C_COMPILER=/usr/bin/clang -DCMAKE_CXX_COMPILER=/usr/bin/clang++ -G Ninja <dir>`

On conan vs fetch_content
I'm not a build system expert, but my understanding is that CMake is not a package manager. I has some facilities that can be used when needing other packages (like `fetch_content`), but it's not complete. My understanding is that `fetch_content` always builds the dependent library along with the program; it can't use pre-built binaries. 

A couple of years back, when I worked on my own programming language, I needed to depended on LLVM. You cannot simply build LLVM inside the CI. This is why I chose to go to conan in the first place.

Bottom line: I can't argue that this setup is the best. I just knows it's good and flexible enough to fit our purposes; and it was relatively easy for me to set it up.

---

If you still have problems setting up your environment, give me a sign before 2:00 PM PST, and I can sit with you and get this fixed.

PR #357, opened 2022-01-03
Add first set of tests

lucteo:
Testing:
- main CPOs
- concepts
- sender factories


lucteo:
waiting for the build system PR to go in first

ericniebler:
This one needs to be rebased now.

lucteo:
Rebased.

This PR contains some TODOs corresponding to some problems in the implementation.
My plan is to write as many tests as I can first, then look at fixing some of these problems.

ericniebler:
Does that mean this PR is still a work in progress?

lucteo:
I think I'm done with this PR. My plan is to create relatively small PRs, easier to review, to incrementally add functionality/fixes.

test/concepts/test_concept_scheduler.cpp
```diff
@@ -0,0 +1,129 @@
...
+
+  friend bool operator==(my_scheduler, my_scheduler) noexcept { return true; }
+  friend bool operator!=(my_scheduler, my_scheduler) noexcept { return false; }
+};
+
+TEST_CASE("type with schedule CPO models scheduler", "[concepts][scheduler]") {
+  REQUIRE(ex::scheduler<my_scheduler>);
+  // NOLINTNEXTLINE
+  auto snd = ex::schedule(my_scheduler{});
+  REQUIRE(ex::sender<decltype(snd)>);
```
miscco, line 46: I would rather combine that into a single line
```suggestion
  REQUIRE(ex::sender<decltype(ex::schedule(my_scheduler{}))>);
```

test/cpos/test_cpo_schedule.cpp
```diff
@@ -0,0 +1,45 @@
...
+#include <execution.hpp>
+
+namespace ex = std::execution;
+
+struct my_sender {
+  template <template <class...> class Tuple, template <class...> class Variant>
+  using value_types = Variant<Tuple<>>;
+  template <template <class...> class Variant>
+  using error_types = Variant<std::exception_ptr>;
+  static constexpr bool sends_done = true;
```
miscco, line 27: Should we use a common base class for those as they are repeated quite often?

miscco (commented):
Quick glance

test/cpos/test_cpo_schedule.cpp
lucteo: My experience tells me that it would not have the desired effect. Most of the senders will have their own definition of these 3 members, so it's hard to abstract it out wile reducing the text size. If that could have been done, it probably  should have been in the proposal. In my experience, the extra indirection added by another abstraction compensates for the small gains you might have.

Also, after a while, you can get accustomed to this, and it's not that bad. Ignoring the `template` lines, the 3 remaining lines are quite simple.

ericniebler, line 27: These days I would use `completion_signatures` for this.
```suggestion
struct my_sender
  : std::execution::completion_signatures<
      std::execution::set_value_t(),
      std::execution::set_error_t(std::exception_ptr),
      std::execution::set_done_t()> {
```

test/concepts/test_concept_scheduler.cpp
```diff
@@ -0,0 +1,129 @@
...
+
+namespace ex = std::execution;
+
+struct my_scheduler {
+  struct my_sender {
+    template <template <class...> class Tuple, template <class...> class Variant>
+    using value_types = Variant<Tuple<>>;
+    template <template <class...> class Variant>
+    using error_types = Variant<std::exception_ptr>;
+    static constexpr bool sends_done = true;
```
ericniebler, line 28: Maybe use `completion_signatures` here as well.

test/cpos/test_cpo_connect.cpp
```diff
@@ -0,0 +1,87 @@
...
+  auto op = ex::connect(my_sender{10}, expect_value_receiver{10});
+  ex::start(op);
+  // the receiver will check the received value
+}
+
+TEST_CASE("cannot connect sender with invalid receiver", "[cpo][cpo_connect]") {
+  // REQUIRE_FALSE(std::tag_invocable<ex::connect_t, my_sender, int>);
+  // TODO: this should not work
+  // invalid check:
+  REQUIRE(std::tag_invocable<ex::connect_t, my_sender, int>);
```
ericniebler, line 58: This works because you have not constrained `my_sender::tag_invoke(connect_t...)` to accept only receivers. It is not a problem with the `connect` CPO or with the `tag_invoke` machinery.

test/cpos/test_cpo_connect.cpp
```diff
@@ -0,0 +1,87 @@
...
+struct my_sender {
+  template <template <class...> class Tuple, template <class...> class Variant>
+  using value_types = Variant<Tuple<>>;
+  template <template <class...> class Variant>
+  using error_types = Variant<std::exception_ptr>;
+  static constexpr bool sends_done = false;
+
+  int value_{0};
+
+  template <typename R>
```
ericniebler, line 42:
```suggestion
  template <receiver_of R>
```

test/cpos/test_cpo_schedule.cpp
lucteo: Didn't know about this. Thank you

test/cpos/test_cpo_connect.cpp
lucteo: Please see [9.6.2](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2021/p2300r3.html#spec-execution.senders.connect), bullet 2:
>  If `R` does not satisfy `execution::receiver`, `execution::connect(s, r)` is ill-formed. 

I agree that for the vast majority of users this is not a problem; I was just trying to test this edge-case

test/concepts/test_concept_scheduler.cpp
lucteo: Made this change all over the place.

test/cpos/test_cpo_schedule.cpp
lucteo: Fixed with Eric's proposal.

test/cpos/test_cpo_connect.cpp
lucteo: Created a new sender without `receiver_of` to expose the problem of connecting to non-receivers, and keep the test the same way.

test/cpos/test_cpo_connect.cpp
ericniebler: Using connect as an invokable and using it as a tag in a call to tag_invoke are two different things.

miscco: Sorry to be late.

I had the exact same issue you are facing when I wrote my initial tests. So I would like to comment on why the code in question is correct, but the tests are actually invalid.

In a nutshell you are getting the right answer to the wrong question.

```std::tag_invocable<ex::connect_t, my_sender, int>``` answers the question of whether the user declared an overload of `tag_invoke(ex::connect_t, mysender, int)`

It does however *not* answer the question whether that is a valid invocation of `ex::connect` because the constraints are checked in the CPO and not in the user provided implementation.

The *right* check it to test `std::invocable<ex::connect_t, my_sender, int>` as that will check whether we can actually call `ex::connect_t(my_sender, int)` and applies all the constraint checks.

From my experience writing some of the tests we should replace most of the instances of `tag_invocable<CPO, ...>` with `invocable<CPO, ...>` except when we want to make clear, that the CPO has additional constraints as in this example, where we should expand to: 

```cpp
  REQUIRE(std::tag_invocable<ex::connect_t, my_sender, int>);
  REQUIRE(!std::invocable<ex::connect_t, my_sender, int>);
```

lucteo: Thank you for the explanation. Makes a lot of sense.
It was also bothering me that I didn't properly test the connect CPO, but I didn't think of using std::invocable.

I will make this change in a followup patch.

lucteo: Stupid me. If I change this to `std::invocable` the test works as expected. I could have sworn that `connect` doesn't constrain the receiver to model the `receiver` concept. But, looking at the git history it's just my imagination.

Sorry, and thank you again!

lucteo: See #372 for the fix

PR #358, opened 2022-01-03
Add get_delegee_scheduler query.

LeeHowes:
Adds the `get_delegee_scheduler` receiver query to implement issue #294. `get_delegee_scheduler` is forwarded independently from `get_scheduler` and forwards through the `on` algorithm. This allows a forward progress delegation target, like `sync_wait`'s scheduler, to forward through to schedulers and algorithms that may use it to support forward progress delegation.

For example, in:

```
sync_wait(on(system_thread_pool, just() | sort(...)));
```

the `sort` implementation and/or `system_thread_pool` scheduler may delegate work to the blocked thread to ensure forward progress in the same fashion that we see now with the standard parallel algorithms.

### DIFF
--- a/include/execution.hpp
+++ b/include/execution.hpp
@@ -404,6 +404,17 @@ namespace std::execution {
         }
       };
 
+      struct get_delegee_scheduler_t {
+        template <class _T>
+          requires nothrow_tag_invocable<get_delegee_scheduler_t, __cref_t<_T>> &&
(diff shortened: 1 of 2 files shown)


ericniebler:
Can you update the reference implementation to match?

LeeHowes:
I did. Somehow I just forgot to add it to git.

ericniebler (approved):

ericniebler:
Oops, a test failure.

PR #363, opened 2022-01-05
factor the context out of the receiver and into a separate parameter to connect

ericniebler:


ericniebler:
Opting not to pursue this design direction for now.

PR #372, opened 2022-01-06
Add when_all tests; fix tag_invocable

lucteo:
Add tests for `when_all`, `when_all_with_variant`, `transfer_when_all`, `transfer_when_all_with_variant`.
Add test placeholders for `split` and `bulk`

Change `std::tag_invocable` to `std::invocable`.
Add "TODO" in the name of the tests that contain TODOs

### DIFF
--- a/test/algos/adaptors/test_when_all.cpp
+++ b/test/algos/adaptors/test_when_all.cpp
@@ -161,7 +161,7 @@ TEST_CASE("when_all terminates with done if one child is cancelled", "[adaptors]
   ex::start(op);
 }
 
-TEST_CASE("TODO: when_all cancels remaining children if error is detected", "[adaptors][when_all]") {
+TEST_CASE("when_all cancels remaining children if error is detected", "[adaptors][when_all]") {
   impulse_scheduler sched;
   error_scheduler err_sched;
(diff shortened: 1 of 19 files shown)


test/algos/adaptors/test_when_all.cpp
```diff
@@ -0,0 +1,328 @@
...
+  CHECK_FALSE(called1);
+  CHECK_FALSE(called3);
+  sched.start_next(); // start the first child
+  CHECK(called1);
+  sched.start_next(); // start the second child; this will call set_done
+  CHECK_FALSE(called3);
+  sched.start_next(); // start the third child
+  // TODO: the third child sender should have been canceled
+  // CHECK_FALSE(called3);
+  // CHECK(cancelled);
```
ericniebler, line 223: I think this doesn't work because `start` on `impulse_scheduler`'s operation state doesn't check the receiver's stop token. It should call `set_done` if stop has been requested.

lucteo: Hehe. I expected magic inside `when_all`, and magic doesn't simply happen :). Fixed it. Thank you!

Issue #366, opened 2022-01-06
Editorial: remove std:: qualification from proposed wording


ericniebler:
`std::` is only used to qualify `std::move` and `std::forward` within the standard.

lucteo:
See PR #376

PR #376, opened 2022-01-07
Fix for #366 - remove `std::` from specification part

lucteo:
Exception made for `std::terminate` that appear in plain text.

### DIFF
--- a/std_execution.bs
+++ b/std_execution.bs
@@ -2637,14 +2637,14 @@ Insert this section as a new subclause, between Searchers <b>[func.search]</b> a
 <ins>
 <blockquote>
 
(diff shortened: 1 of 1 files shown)


std_execution.bs
```diff
@@ -3205,23 +3205,23 @@ enum class forward_progress_guarantee {
...
     <pre highlight=c++>
     template&lt;class T, class E = exception_ptr>
     concept receiver =
       move_constructible&lt;remove_cvref_t&lt;T>> &&
       constructible_from&lt;remove_cvref_t&lt;T>, T> &&
       requires(remove_cvref_t&lt;T>&& t, E&& e) {
-        { execution::set_done(std::move(t)) } noexcept;
-        { execution::set_error(std::move(t), (E&&) e) } noexcept;
+        { execution::set_done(move(t)) } noexcept;
+        { execution::set_error(move(t), (E&&) e) } noexcept;
```
ericniebler, line 3217: Please leave `std::` for `std::move` and `std::forward`. The standard does that for historical reasons, I'm told.

ericniebler (changes requested):

ericniebler (approved):

Issue #364, opened 2022-01-06
Fix incorrect uses of inline namespaces in spec


ericniebler:
```c++
  // [execution.queries], general queries
  inline namespace unspecified {
    struct get_scheduler_t;
    inline constexpr get_scheduler_t get_scheduler{};
    struct get_allocator_t;
    inline constexpr get_allocator_t get_allocator{};
    struct get_stop_token_t;
    inline constexpr get_stop_token_t get_stop_token{};
  }
```

The idea was for `get_scheduler_t` to _not_ have `std::execution` as an associated namespace, since we don't want to end up looking there for `tag_invoke` overloads. However, `inline namespace` doesn't achieve that.

Proposed resolution
===================

Change this to:

```c++
  // [execution.queries], general queries
  namespace <i>general-queries</i> {
    struct get_scheduler_t;
    struct get_allocator_t;
    struct get_stop_token_t;
  }
  using <i>general-queries</i>::get_scheduler_t;
  using <i>general-queries</i>::get_allocator_t;
  using <i>general-queries</i>::get_stop_token_t;

  inline constexpr get_scheduler_t get_scheduler{};
  inline constexpr get_allocator_t get_allocator{};
  inline constexpr get_stop_token_t get_stop_token{};
```

Apply this change everywhere.

lucteo:
@ericniebler : I can work on this, but I propose a 2-phase fix for this issue:
1) make the changes just in .bs spec
2) also make the changes in the code

I think it would make sense to fix 2) only after the unit tests that I have are merged (one PR pending and another set of commits that are not yet up). This way, we also have the change tested.

lucteo:
See PR #377  for the changes in the spec only.

lucteo:
@ericniebler: I'm trying to fix this in the code, and I realize the this introduces a usability problem.

types like `set_value_t` are not directly visible by the user. The best thing that the user can do is use `decltype(set_value)` instead. But, this resolves into `const hidden_namespace::set_value_t`.

While I was trying to make the code work by taking the decayed type, this will not work, as we are violating the specification. For example, take the following code:
```
    friend inline_scheduler tag_invoke(
        ex::get_completion_scheduler_t<decltype(ex::set_value)>, my_sender) {
      return {};
    }
```

[...]

PR #377, opened 2022-01-07
Fix incorrect uses of inline namespaces in spec (#364)

lucteo:


std_execution.bs
```diff
@@ -2825,24 +2826,24 @@ namespace std::execution {
...
   enum class forward_progress_guarantee;
-  inline namespace <i>unspecified</i> {
+  namespace <i>unspecified</i> {
     struct forwarding_scheduler_query_t;
-    inline constexpr forwarding_scheduler_query_t forwarding_scheduler_query{};
 
     template &lt;class T>
       concept <i>forwarding-scheduler-query</i> = // exposition only
-        forwarding_scheduler_query(T{});
+        forwarding_scheduler_query_t{}(T{});
```
lucteo, line 2835: Do we need to specify this concept? Is implementation detail.

std_execution.bs
```diff
@@ -2853,33 +2854,33 @@ namespace std::execution {
...
-  inline namespace <i>unspecified</i> {
+  namespace <i>unspecified</i> {
     struct forwarding_receiver_query_t;
-    inline constexpr forwarding_receiver_query_t forwarding_receiver_query{};
 
     template &lt;class T>
       concept <i>forwarding-receiver-query</i> = // exposition only
-        forwarding_receiver_query(T{});
+        forwarding_receiver_query_t{}(T{});
   }
```
lucteo, line 2873: Do we need to specify this concept? Is implementation detail.

std_execution.bs
```diff
@@ -2950,79 +2944,90 @@ namespace std::execution {
...
+    struct when_all_t;
+    struct when_all_with_variant_t;
+    struct transfer_when_all_t;
+    struct transfer_when_all_with_variant_t;
+    struct done_as_optional_t;
+    struct done_as_error_t;
+    struct ensure_started_t;
+    struct start_detached_t;
+  }
+  inline namespace <i>unspecified-inline</i> {
```
lucteo, line 2969: I'm not sure if it makes sense to have this inline namespace anymore.

std_execution.bs
```diff
@@ -2805,16 +2805,17 @@ namespace std::execution {
     concept <i>class-type</i> = <i>decays-to</i>&lt;T, T> && is_class_v&lt;T>;  // exposition only
 
   // [execution.queries], general queries
-  inline namespace <i>unspecified</i> {
+  namespace <i>unspecified</i> {
```
ericniebler, line 2808: It would be better to be clear about the fact that these types occupy their _own_ namespace, different from all the others.

```suggestion
  namespace <i>general-queries</i> { <i>// Exposition only</i>
```

Please make this change everywhere.

std_execution.bs
ericniebler: We can remove this exposition only concept.

std_execution.bs
ericniebler: Leave it for now. There's been some drift. When `get_scheduler` and friends became "general" queries, `forwarding_receiver_query` didn't come along for the ride. I'll file an issue.

std_execution.bs
```diff
@@ -2950,79 +2944,90 @@ namespace std::execution {
...
+    struct bulk_t;
+    struct split_t;
+    struct when_all_t;
+    struct when_all_with_variant_t;
+    struct transfer_when_all_t;
+    struct transfer_when_all_with_variant_t;
+    struct done_as_optional_t;
+    struct done_as_error_t;
+    struct ensure_started_t;
+    struct start_detached_t;
```
ericniebler, line 2967: It has always seemed odd to me that all these CPOs were dumped into the same namespace. Let's at least move the adaptors and consumers (everything from `transfer_just_t`) into its own namespace.

std_execution.bs
```diff
@@ -2950,79 +2944,90 @@ namespace std::execution {
...
     <i>unspecified</i> just_done() noexcept;
     template&lt;<i>movable-value</i>... Ts>
       <i>unspecified</i> just(Ts &&...);
-    struct transfer_just_t;
-    inline constexpr transfer_just_t transfer_just{};
+    inline constexpr <i>unspecified</i>::transfer_just_t transfer_just{};
 
     // [execution.senders.adaptors], sender adaptors
     template&lt;<i>class-type</i> D>
       struct sender_adaptor_closure { };
```
ericniebler, line 2995: Put `sender_adaptor_closure` in its own (non-`inline`) namespace please.

std_execution.bs
```diff
@@ -2950,79 +2944,90 @@ namespace std::execution {
...
-    struct schedule_t;
-    inline constexpr schedule_t schedule{};
+    inline constexpr <i>unspecified</i>::schedule_t schedule{};
     template&lt;<i>movable-value</i>... Ts>
       <i>unspecified</i> just(Ts &&...) noexcept(<i>see below</i>);
     template&lt;<i>movable-value</i> Error>
       <i>unspecified</i> just_error(Error &&) noexcept(<i>see below</i>);
     <i>unspecified</i> just_done() noexcept;
     template&lt;<i>movable-value</i>... Ts>
       <i>unspecified</i> just(Ts &&...);
```
ericniebler, line 2990: Since these are functions (not CPOs) and we don't want them findable with ADL, please put them into their own (non-`inline`) namespace.

std_execution.bs
ericniebler: Right, the CPOs can live directly in `std::execution`.

std_execution.bs
```diff
@@ -2950,79 +2944,90 @@ namespace std::execution {
...
+    inline constexpr <i>unspecified</i>::when_all_t when_all{};
+    inline constexpr <i>unspecified</i>::when_all_with_variant_t when_all_with_variant{};
+    inline constexpr <i>unspecified</i>::transfer_when_all_t transfer_when_all{};
+    inline constexpr <i>unspecified</i>::transfer_when_all_with_variant_t
       transfer_when_all_with_variant{};
 
     template&lt;typed_sender S>
       using <i>into-variant-type</i> = <i>see-below</i>; // exposition-only
     template&lt;typed_sender S>
       <i>see-below</i> into_variant(S &&);
```
ericniebler, line 3021: `into_variant` should live in its own namespace, or with the `just*` functions whichever.

ericniebler (changes requested):
This is fantastic, thanks. A few nits.

std_execution.bs
```diff
@@ -2853,33 +2849,33 @@ namespace std::execution {
...
   // [execution.receivers.queries], receiver queries
-  inline namespace <i>unspecified</i> {
+  namespace <i>receivers-queries</i> { // exposition only
     struct forwarding_receiver_query_t;
-    inline constexpr forwarding_receiver_query_t forwarding_receiver_query{};
 
     template &lt;class T>
       concept <i>forwarding-receiver-query</i> = // exposition only
-        forwarding_receiver_query(T{});
+        forwarding_receiver_query_t{}(T{});
```
ericniebler, line 2867: Move this into `std::execution` please.

ericniebler (commented):
Just one more tweak, then this one is good to go.

PR #378, opened 2022-01-07
Add tests for remaining sender algos

lucteo:


lucteo:
I'm happy to report that we have now 321 tests. There are still gaps, and there are still issues to be fixed, but this is a good start.

image

PR #381, opened 2022-01-08
Delegee scheduler on

LeeHowes:
I realised we need to forward get_delegee_scheduler to the internal receiver of on() to make delegation make sense. It occurred to me that all receiver queries should forward so that the scheduler can use the right allocator etc too.

### DIFF
--- a/std_execution.bs
+++ b/std_execution.bs
@@ -3825,11 +3825,13 @@ are all well-formed.
 
             3. When `execution::set_done(r)` is called, it calls `execution::set_done(out_r)`.
 
+            4. Implements the `get_delegee_scheduler` receiver query to return the result of `get_delegee_schedler(out_r)`.
+
         2. Calls `execution::schedule(sch)`, which results in `s2`. It then calls `execution::connect(s2, r)`, resulting in `op_state2`.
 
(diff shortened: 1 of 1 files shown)


ericniebler (commented):
We have blanket wording that all receiver adaptors created by the sender adaptors forward receiver queries by default. This change isn't necessary, I don't think.

LeeHowes:
`on` also has wording that says:
" 2. `r2` is a receiver that wraps a reference to `out_r`. It forwards all receiver completion signals and receiver queries to `out_r`. "

So it struck me that if we are specific about the forwarding for the upstream receiver, we should be specific about the one we connect to the scheduler too.

You are right though that this is fairly clear now I find it:

> A type `T` is a <i>forwarding receiver query</i> if it is the type of a customization point object that models <code>forwarding-receiver-query</code>. Unless otherwise specified, whenever a sender adaptor constructs a receiver that it passes to another sender's connect, that receiver shall propagate forwarding receiver queries to a receiver accepted as an argument of `execution::connect`. This requirements applies to any sender returned from a function that is selected by the implementation of such sender adaptor.

ericniebler:
So I think we don't need this PR, and the note about forwarding of receiver queries in `on` is superfluous and could (should) be removed.

Please reopen if you disagree and feel this change is still necessary.

PR #391, opened 2022-01-14
Add make_completion_signatures utility, a then() alg example, __debug_sender utility

ericniebler:
## `make_completion_signatures`

`completion_signatures` is useful for authors of sender _factories_, but sender _adaptors_ often can't use it because their completions usually depend in a complex way on the completions of the sender they're adapting.

Enter `make_completion_signatures`, which takes a sender, and environment, and a bunch of other template arguments for munging the completion signatures of a sender in interesting ways.

```c++
template <class... Args>
  using __default_set_value = set_value_t(Args...);

template <class Err>
  using __default_set_error = set_error_t(Err);

template <
  sender Sndr,
  class Env = no_env,
  class AddlSigs = completion_signatures<>,
  template <class...> class SetValue = __default_set_value,
  template <class> class SetError = __default_set_error,
  bool SendsStopped = sender_traits_t<Sndr, Env>::sends_stopped>
    requires typed_sender<Sndr, Env>
using make_completion_signatures =
  completion_signatures< ... >;
```

* `SetValue` : an alias template that accepts a set of value types and returns a function type of the form `set_value_t(Types...)`, or `void` to mean "no signature".
* `SetError` : an alias template that accepts an error types and returns a function type of the form `set_error_t(T)` or `void` to mean "no signature".
* `SendsStopped` : a bool that controls whether the signature `set_stopped_t()` will be added to the list of completion signatures or not.
* `AddlSigs` : an instantiation of `completion_signatures` with a list of completion signatures `Sigs...` to the added to the list.

`make_completion_signatures` does the following:
* Let `Vs...` be a pack of the non-`void` types in the `__typelist` named by `value_types_of_t<Sndr, Env, SetValue, __typelist>`.
* Let `Es...` be a pack of the non-`void` types in the `__typelist` named by `error_types_of_t<Sndr, Env, __errorlist>`, where `__errorlist` is an alias template such that `__errorlist<Ts...>` names `__typelist<SetError<Ts>...>`.
* Let `Ss...` be an empty pack if `SendsStopped` is `false`; otherwise, a pack containing the single type `set_stopped_t()`.
* Let `MoreSigs...` be a pack of the template arguments of the `completion_signatures` instantiation named by `AddlSigs`.

Then `make_completion_signatures<Sndr, Env, AddlSigs, SetValue, SetDone, SendsStopped>` names the type `completion_signatures< Sigs... >` where `Sigs...` is the unique set of types in `[Vs..., Es..., Ss..., MoreSigs...]`.

## `__debug_sender`

Understanding why a particular sender doesn't connect to a particular receiver is nigh impossible in the current design due to limitations in how the compiler reports overload resolution failure in the presence of constraints. `__debug_sender` is a utility to assist with the process. It gives you the deep template instantiation backtrace that you need to understand where in a chain of senders the problem is occurring.

```c++
template <class _Sigs, class _Env = __empty_env, class _Sender>
  void __debug_sender(_Sender&& __sndr, _Env = {});

template <class _Sender>
  void __debug_sender(_Sender&& __sndr);

template <class _Env, class _Sender>
  void __debug_sender(_Sender&& __sndr, _Env);
```

**Usage:**

To find out where in a chain of senders, a sender is failing to connect to a receiver, pass it to `__debug_sender`, optionally with an environment argument; e.g. `__debug_sender(sndr [, env])`

To find out why a sender will not connect to a receiver of a particular signature, specify the error and value types as an explicit template argument that names an instantiation of `completion_signatures`; e.g.: `__debug_sender<completion_signatures<set_value_t(int)>>(sndr [, env])`.

**How it works**

The `__debug_sender` function `connect`'s the sender to a `__debug_receiver`, whose environment is augmented with a special `__is_debug_env_t` query. An additional fall-back overload is added to the `connect` CPO that recognizes receivers whose environments respond to that query and lets them through. Then in a non-immediate context, it looks for a `tag_invoke(connect_t...)` overload for the input sender and receiver. This will recurse until it hits the `tag_invoke` call that is causing the failure.

At least with clang, this gives me a nice backtrace, at the bottom of which is the faulty `tag_invoke` overload with a mention of the constraint that failed.


kirkshoop (commented):
Can't wait to use this!

PR #392, opened 2022-01-14
Cleanup inline namespaces

lucteo:
Remove inline namespaces surrounding CPO types

### DIFF
--- a/examples/algorithms/retry.hpp
+++ b/examples/algorithms/retry.hpp
@@ -101,7 +101,7 @@ struct _retry_sender {
   }
 
   template <class Env>
-  friend auto tag_invoke(stdex::get_sender_traits_t, const _retry_sender&, Env)
+  friend auto tag_invoke(decltype(stdex::get_sender_traits), const _retry_sender&, Env)
     -> stdex::sender_traits_t<const S&, Env>;
 };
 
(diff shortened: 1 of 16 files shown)


lucteo:
work in progress; currently we do this just for the set_* CPOs; we need it for all the CPOs

examples/algorithms/retry.hpp
```diff
@@ -101,7 +101,7 @@ struct _retry_sender {
   }
 
   template <class Env>
-  friend auto tag_invoke(stdex::get_sender_traits_t, const _retry_sender&, Env)
```
ericniebler, line 104: The `get_sender_traits_t` type needs to be available. IIRC, LEWG asked for the tags to have name-able types.

lucteo: All the tag types?
- `set_value_t`, `set_error_t`, `set_stopped_t`
- `connect_t`, `start_t`, `schedule_t`, ...
- `transfer`, `let_value`, `bulk`, ...

It means, I also need to change specification to show this.

ericniebler: Yes

lucteo:
Dropping this in the favour of #396

PR #396, opened 2022-01-15
Make the namespaces non-inline

lucteo:
The CPOs are still isolated in the namespaces, but we use the names
outside in `std::execution` namespaces. The CPO objects are also defined
outside.


lucteo:
I did a bad merge. Need to abandon this and start again :(

PR #397, opened 2022-01-15
Make the namespaces non-inline (another attempt)

lucteo:
The CPOs are still isolated in the namespaces, but we use the names
outside in `std::execution` namespaces. The CPO objects are also defined
outside.

### DIFF
--- a/std_execution.bs
+++ b/std_execution.bs
@@ -1039,6 +1039,15 @@ This proposal also draws heavily from our experience with [Thrust](https://githu
 
 # Revision history # {#revisions}
 
+## R5 ## {#r5}
+
+The changes since R4 are as follows:
+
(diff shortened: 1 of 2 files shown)


std_execution.bs
```diff
@@ -1039,6 +1039,15 @@ This proposal also draws heavily from our experience with [Thrust](https://githu
...
 
+## R5 ## {#r5}
+
+The changes since R4 are as follows:
+
+<b>Fixes:</b>
+  * Properly expose the tag types in the header `<execution>` synopsis.
+
+<b>Enhancements:</b>
+
```
ericniebler, line 1050: This can be part of R4. I haven't shared it with a P number yet.

Issue #398, opened 2022-01-16
Typos/issues Section 8 (General utilities library)


dietmarkuehl:
There are a number of issue with Section 8:

1. Some declarations in the synopsis have a mismatch between the name used in the template head and the function declaration: `Fn` vs. `Tag`.
2. There are numerous places where the `<` isn't HTML escaped and the explicit argument list is swallowed when formatted.
3. Some arguments to `std::forward` are plain missing.
4. The expression for the definition of `call-result-t` is a list of objects rather than a call expression.
5. There are some casts which are either unnecessary (`auto(tag)`) or should use `std::forward`. However, see next item.
6. The specification in 8.1.2 (`tag_invoke`) is too verbose. I think it can be expressed in terms of [customization.point.object] (whose reference is missing) and [defns.expression-equivalent] as is done for other customization point objects.

I'm fairly certain about all but the last part. For the last part it may be reasonable to get an opinion from other LWG members. I will add a PR shortly.

ericniebler:
Thank you! Regarding `tag_invoke`, I've found that the diagnostics are very bad unless its implementation matches that in P2300. So it borders on overspecialization simply because it's so easy to do it poorly.

dietmarkuehl:
I have created [PR 399](https://github.com/brycelelbach/wg21_p2300_std_execution/pull/399) to address the issues mentioned here.

Regarding Eric's comment on recommending an exact implementation: if an exact implementation should be suggested, I think it would best become a Note, e.g., to allow implementations to deviate if they have a better implementation like explicit compiler support.

villevoutilainen:
> Regarding Eric's comment on recommending an exact implementation: if an exact implementation should be suggested, I think it would best become a Note, e.g., to allow implementations to deviate if they have a better implementation like explicit compiler support. 

Implementations are always allowed to "deviate" based on the as-if rule.

dietmarkuehl:
I don't think this section will stay as is. At the very least, a note should be added to motivate why the exact implementation is given. ...and if the exact implementation is given, the casts need to go or it needs to be explicitly mentioned that this formulation is needed as well.

I have already mentioned that I'm not as confident about this particular change than about the others. My objective of this exercise is to streamline the LWG processing of this paper: unless it is of much better quality than nearly all papers we receive we will not manage to get through it for the C++23 time-frame.

PR #399, opened 2022-01-16
Various fixes for section 8 (General utilities library)

dietmarkuehl:
This PR is intended to fix the issues reported by [#398](https://github.com/brycelelbach/wg21_p2300_std_execution/issues/398).

### DIFF
--- a/std_execution.bs
+++ b/std_execution.bs
@@ -3002,7 +3002,15 @@ Insert this section as a new subclause, between Searchers <b>[func.search]</b> a
 <ins>
 <blockquote>
 
-1. The name `std::tag_invoke` denotes a customization point object [customization.point.object]. Given subexpressions `T` and `A...` the expression `tag_invoke(T, A...)` is expression-equivalent [defns.expression-equivalent] to `tag_invoke(T, A...)` if it is a well-formed expression with overload resolution performed in a context that does not include the declaration of `std::tag_invoke`. Otherwise, `tag_invoke(T, A...)` is ill-formed.
+1. The name `std::tag_invoke` denotes a customization point object [customization.point.object]. Given subexpressions `T` and `A...`, the expression `std::tag_invoke(T, A...)` is expression-equivalent [defns.expression-equivalent] to `tag_invoke(T, A...)` if it is a well-formed expression with overload resolution performed in a context in which unqualified lookup for `tag_invoke` finds only the declaration
+
+    ```c++
+    void tag_invoke();
+    ```
+
+    Otherwise, `std::tag_invoke(T, A...)` is ill-formed.
+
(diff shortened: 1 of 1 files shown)


dietmarkuehl:
As a side note: I mentioned the idea of splitting P2300 up into different parts. This section may be a prime candidate for a stand-alone change.

ericniebler:
The rest of the spec depends on ` tag_invoke`, so I'm not sure what would be gained by splitting it out.

dietmarkuehl:
There were proposals in that space in the past which caused substantial discussion and I think it is more effective to discuss/process this particular part independent of the rest of the paper. It would be a shame if we spend a lot of effort to get things sunk for C++23 due to objection to the overall customization approach.

ericniebler:
There is no practical way of re-specifying P2300 without `tag_invoke`. If the Committee disapproves of `tag_invoke`, I think they disapprove of P2300.

std_execution.bs
```diff
@@ -3002,19 +3002,7 @@ Insert this section as a new subclause, between Searchers <b>[func.search]</b> a
...
-      template &lt;class Tag, class... Args>
-          requires tag_invocable&lt;Tag, Args...>
-        constexpr auto operator()(Tag tag, Args&&... args) const
-          noexcept(nothrow_tag_invocable&lt;Tag, Args...>)
-          -> tag_invoke_result_t&lt;Tag, Args...> {
-          return tag_invoke(auto(tag), (Args&&) args...);
-        }
-    };
-    </pre>
+1. The name `std::tag_invoke` denotes a customization point object [customization.point.object]. Given subexpressions `T` and `A...` the expression `tag_invoke(T, A...)` is expression-equivalent [defns.expression-equivalent] to `tag_invoke(T, A...)` if it is a well-formed expression with overload resolution performed in a context that does not include the declaration of `std::tag_invoke`. Otherwise, `tag_invoke(T, A...)` is ill-formed.
```
ericniebler, line 3005: Cribbing from the `ranges::begin` wording and adding the `void tag_invoke();` declaration to avoid the degenerate case where phase 1 name lookup on `tag_invoke` finds an object instead of a function:

```suggestion
1. Given subexpressions `T` and `A...`, if `tag_invoke(T, A...)` is a valid expression with overload resolution performed in a context in which unqualified lookup for `tag_invoke` finds only the declaration

    ```c++
    void tag_invoke();
    ```

then `std::tag_invoke(T, A...)` is expression-equivalent [defns.expression-equivalent] to `tag_invoke(T, A...)` with overload resolution performed in the above context.

2. [Note: Diagnosable ill-formed cases above result in substitution failure when `std::tag_invoke(T, A...)` appears in the immediate context of a template instantiation. —end note]
```

ericniebler (changes requested):
Thoughts about this change?

ericniebler (approved):

PR #400, opened 2022-01-18
remove untyped senders, rename sender_traits to completion_signatures

ericniebler:
* Nukes untyped senders
* Nukes `sender_base`
* Renames `sender_traits` to `completion_signatures`
* When customization of `get_completion_signatures` isn't found, falls back to looking for a `completion_signatures` nested type on the sender, rather than looking for nested `value_types`/`error_types`/`sends_stopped`.

### DIFF
--- a/std_execution.bs
+++ b/std_execution.bs
@@ -4348,7 +4348,7 @@ enum class forward_progress_guarantee {
     struct <i>no-completion-signatures</i> {};
     </pre>
 
-#### `completion_signatures_of_t` <b>[exec.sndtraitst]</b> #### {#spec-exec.sndtraitst}
+#### `execution::completion_signatures_of_t` <b>[exec.sndtraitst]</b> #### {#spec-exec.sndtraitst}
 
 1. The alias template `completion_signatures_of_t` is used to query a sender type for facts associated with the signals it sends.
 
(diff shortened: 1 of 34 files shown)


std_execution.bs
```diff
@@ -3643,16 +3698,16 @@ namespace std::execution {
            class E = no_env,
            template &lt;class...> class Tuple = <i>decayed-tuple</i>,
            template &lt;class...> class Variant = <i>variant-or-empty</i>>
-      requires typed_sender&lt;S, E>
+      requires sender&lt;S, E>
     using value_types_of_t =
-      typename sender_traits_t&lt;S, E>::template value_types&lt;Tuple, Variant>;
+      typename completion_signatures_of_t&lt;S, E>::template value_types&lt;Tuple, Variant>;
```
kirkshoop, line 3703: I was thinking that we could make the value_types&error_types&sends_stopped unspecified implementation details - even if we kept the sender_value_types_t etc.. alias'.

As I continued it seemed that there was still a lot of the spec directly dependent on the value_types&error_types&sends_stopped instead of through the alias'.

I think it is a nice-to-have to make the internals impl defined.

kirkshoop (approved):
Nice!

std_execution.bs
ericniebler: I'm going to leave it for now. Bryce needs an R4 for the polls ASAP.

griwes (approved):

PR #402, opened 2022-01-19
Change "ill-formed, no diagnostic required" to undefined behavior

villevoutilainen:
IFNDR is the biggest hammer in C++. It renders the whole program UB if run,
regardless of whether the construct that is IFNDR is ever reached.
Furthermore, the cases that this patch changes are downright impossible to diagnose,
so UB seems like a much better fit.

The other changes in this patch change IFNDR to "shall not". There's no real
spec difference between those, I just prefer that phrasing.


ericniebler:
Thank you! We just published R4 yesterday, but there will be an RT.

PR #410, opened 2022-01-21
Tweak the receiver concepts to not mention exception_ptr

ericniebler:
This redesign changes `receiver_of` to take a completion signatures struct, and it checks that the receiver can receive each signal in the list of completion signatures.

It also redesigns `make_completion_signatures` to be more general and adds a `sends_stopped<Sndr, Env>` constexpr bool.

This PR also changes all the sender adaptors to express their completion signatures in terms of `completion_signatures` and removes the old unloved template template sender traits design. Everything is now expressed in terms of `completion_signatures` directly.

Proposed Resolution
==================

* Remove the default implementation of the `get_env` receiver query. The
    presence of a customization of `get_env` becomes the distinguishing feature
    of receivers. A "receiver" no longer needs to provide any completion
    channels at all to be considered a receiver, only `get_env`.

* The `receiver_of` concept, rather than accepting a receiver and some value
    types, is changed to take a receiver and an instance of the
    `completion_signatures<>` class template. A sender uses
    `completion_signatures<>` to describe the signals with which it completes. The
    `receiver_of` concept ensures that a particular receiver is capable of
    receiving those signals.

    Notably, if a sender only sends a value (i.e., can never send an error or a
    stopped signal), then a receiver need only provide a value channel to be
    compatible with it.

* A receiver's customization of `set_value` is required to be `noexcept`. This
    makes it possible for many senders to become "no-fail"; that is, they cannot
    complete with an error. `just(1)`, for instance, will only ever successfully
    send an integer through the value channel. An adaptor such as
    `then(sndr, fun)` can check whether `fun` can ever exit exceptionally when
    called with all the sets of values that `sndr` may complete with. If so, the
    `then` sender must add `set_error_t(exception_ptr)` to its list of completions.
    Otherwise, it need not.

* The `sender_to` concept, which checks whether a sender and a receiver can be
    connected, now enforces that the sender's completion signatures can in fact
    be handled by the receiver. Previously, it relied on sender authors to
    properly constrain their `connect` customizations.

* For good measure, the `connect` customization point also checks whether a
    receiver can receive all of the sender's possible completions before trying
    to dispatch via `tag_invoke` to a `connect` customization. This often
    entirely frees sender authors from having to constrain their `connect`
    customizations at all. It is enough to customize
    `get_completion_signatures`, and the type checking is done automatically.

* `get_completion_signatures` becomes conceptually simpler; it is now required
    to return an instantiation of the `completion_signatures` class template,
    which is now enforced by the `sender` concept. The old sender traits
    design, with its unloved `template template` alias interface, is no more.

* The `value_types_of_t` and `error_types_of_t` template aliases remain however,
    and are expressed in terms of `completion_signatures`. In addition, they are
    joined by a `sends_stopped<Sndr, Env>` Boolean variable template.

* `dependent_completion_signatures` is replaced with `completion_signatures<>`
    instantiated with zero arguments. If a sender sends no signals, it never
    completes, and cannot rightly be considered a sender.

* The `make_completion_signatures` design is slightly tweaked to be more general
    and more powerful to accommodate the larger role `completion_signatures`
    plays in the design.

### DIFF
--- a/include/execution.hpp
+++ b/include/execution.hpp
@@ -532,27 +532,25 @@ namespace std::execution {
   /////////////////////////////////////////////////////////////////////////////
   namespace __completion_signatures {
     template <class... _Args>
-      using __set_value_sig = set_value_t(_Args...);
+      using __set_value_sig = completion_signatures<set_value_t(_Args...)>;
 
     template <class _Err>
-      using __set_error_sig = set_error_t(_Err);
+      using __set_error_sig = completion_signatures<set_error_t(_Err)>;
 
     template<class _Sender, class _Env, class _Sigs, class _SetValue, class _SetError, class _SendsStopped>
     using __compl_sigs_t =
       __minvoke<
-        __concat<__remove<void, __munique<__q<completion_signatures>>>>,
+        __concat<__munique<__q<completion_signatures>>>,
         _Sigs,
-        __value_types_of_t<_Sender, _Env, _SetValue, __q<__types>>,
-        __error_types_of_t<_Sender, _Env, __transform<_SetError, __q<__types>>>,
-        __if<_SendsStopped, __types<set_stopped_t()>, __types<>>>;
+        __value_types_of_t<_Sender, _Env, _SetValue, __concat<>>,
+        __error_types_of_t<_Sender, _Env, __transform<_SetError, __concat<>>>,
+        __if<__sends_stopped<_Sender, _Env>, _SendsStopped, __types<>>>;
 
-    template<class _Sender, class _Env, class _Sigs, class _SetValue, class _SetError, bool _SendsStopped>
+    template<class _Sender, class _Env, class _Sigs, class _SetValue, class _SetError, class _SetStopped>
     auto __make(int) ->
-        __compl_sigs_t<
-          _Sender, _Env, _Sigs, _SetValue, _SetError, __bool<_SendsStopped>>
-        requires true;
+      __compl_sigs_t<_Sender, _Env, _Sigs, _SetValue, _SetError, _SetStopped>;
 
-    template<class, __one_of<no_env> _Env, class, class, class, bool>
+    template<class, __one_of<no_env> _Env, class, class, class, class>
     auto __make(long) -> dependent_completion_signatures<_Env>;
   } // namespace __completion_signatures
 
@@ -568,10 +566,10 @@ namespace std::execution {
   //
   //  ```c++
   //  template <class... Args>
-  //    using __default_set_value = set_value_t(Args...);
+  //    using __default_set_value = completion_signatures<set_value_t(Args...)>;
   //
   //  template <class Err>
-  //    using __default_set_error = set_error_t(Err);
+  //    using __default_set_error = completion_signatures<set_error_t(Err)>;
   //
   //  template <
   //    sender Sndr,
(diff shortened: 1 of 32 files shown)


miscco:
I love the direction, but could you please pull out the stdex ->exec changes and merge them separately?

This is adding a lot of churn and makes reviewing much harder

ericniebler:
Will do.

test/concepts/test_concepts_receiver.cpp
```diff
@@ -1,129 +1,148 @@
...
- *
- *   https://llvm.org/LICENSE.txt
- *
- * Unless required by applicable law or agreed to in writing, software
- * distributed under the License is distributed on an "AS IS" BASIS,
- * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
- * See the License for the specific language governing permissions and
- * limitations under the License.
- */
+// /*
```
villevoutilainen, line 1: This change looks like it's commenting out everything??

ericniebler: Yes, temporary change while I've been monkeying with the concepts. The receiver concept tests all need to be rewritten, but that didn't seem like the highest priority at the moment.

include/execution.hpp
```diff
@@ -2889,227 +2709,202 @@ namespace std::execution {
...
-    } // namespace __impl
+        template <__decays_to<__sender> _Self, class _Env>
+          friend auto tag_invoke(get_completion_signatures_t, _Self&&, _Env) ->
+            make_completion_signatures<
+              __member_t<_Self, _Sender>,
+              _Env,
+              make_completion_signatures<
+                schedule_result_t<_Scheduler>,
+                _Env,
+                completion_signatures<set_error_t(exception_ptr)>,
```
ben-craig: General readability note that is impeding my ability to review the code:

It is very hard for me to tell which facility this is even involved with.  There's __sender, __operation1, __receiver1, and other structs preceding this line.  My main hint as to what facility this belongs to is a comment a couple hundred lines up though, where there's a mention of execution.senders.adaptors.schedule_from.

My suggestion is to add more redundancy in naming.  Instead of __sender, use __sender_schedule_from.  The extra redundancy can help readers like myself.  It might even help diagnostics a bit.

include/execution.hpp
```diff
@@ -2252,142 +2183,28 @@ namespace std::execution {
...
+            __is_typed_completions<
+              completion_signatures_of_t<invoke_result_t<_Fun, _Args...>, _Env>>
+        struct __tfx_signal<_Env, _Fun, _Set, _Set(_Args...)> {
+          using __completions =
+            completion_signatures_of_t<invoke_result_t<_Fun, _Args...>, _Env>;
+          using type =
+            __minvoke2<
+              __push_back<__q<completion_signatures>>,
+              __completions,
+              set_error_t(exception_ptr)>;
```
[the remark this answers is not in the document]
ericniebler: > Does these mean let_stopped (I think that's where we are?) requires exception_ptr?

I'm being sloppy here. What this code _should_ be doing is checking to see whether the function call is potentially throwing. If so, it should add `set_error_t(exception_ptr)` to the list of completions. Otherwise, it should leave it off.

> New Line 2229, there's comments about "for the purposes of the receiver concept..."
Is that still accurate?

Nope, no longer accurate. I'll remove it.

include/execution.hpp
ericniebler: I get where you're coming from. All these types are in an appropriately named namespace, but the namespace is often far from where the sender/receiver types are defined, so it might be better to do as you suggest.

include/execution.hpp
```diff
@@ -2252,142 +2180,28 @@ namespace std::execution {
...
-                __right_fold<false_type, __q2<__or>>>>::value;
+      template <class _Env, class _Fun, class _Set, class... _Args>
+          requires invocable<_Fun, _Args...> &&
+            sender<invoke_result_t<_Fun, _Args...>, _Env> &&
+            __is_typed_completions<
+              completion_signatures_of_t<invoke_result_t<_Fun, _Args...>, _Env>>
+        struct __tfx_signal<_Env, _Fun, _Set, _Set(_Args...)> {
+          using __completions =
+            completion_signatures_of_t<invoke_result_t<_Fun, _Args...>, _Env>;
+          using type =
```
lucteo: (minor) If I'm reading this correctly we are not properly computing the end completion signatures.

If the input sender completes with `set_value_t(int)`, `set_value_t(string)`, `set_error_t(error_code)`, and the transformation function is  `[](int x) { return just(x + 0.1415); }`, I would expect the output completions to be `set_value_t(int)`, `set_value_t(string)`, `set_error_t(error_code)` (two remain, one removed, one added).

On the other hand, this doesn't seem important at this point

ericniebler: No, assuming you're talking about a call to `let_value`, this should not require. The function should be able to handle whatever (sets of) value types with which the input sender can complete. Since `string` doesn't convert to `int`, I wouldn't expect the `connect` call to type-check.

include/execution.hpp
```diff
@@ -1471,7 +1403,7 @@ namespace std::execution {
     } // namespace __impl
 
     struct __submit_t {
-      template <receiver _Receiver, sender_to<_Receiver> _Sender>
+      template <class _Receiver, sender_to<_Receiver> _Sender>
```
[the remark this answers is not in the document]
ericniebler: Could be an artifact of an earlier edit in which I did away with the `receiver` concept entirely. I should bring these back, but I'll note that in this particular case it's unnecessary since `sender_to` enforces that `_Receiver` is a `receiver`.

include/execution.hpp
```diff
@@ -250,45 +144,149 @@ namespace std::execution {
...
+        _Completion::template __with_receiver<_Receiver>::value;
+  } // namespace __receiver_concepts
+
+  /////////////////////////////////////////////////////////////////////////////
+  // [execution.receivers]
+  template <class _Receiver>
+    concept receiver =
+      environment_provider<__cref_t<_Receiver>> &&
+      move_constructible<remove_cvref_t<_Receiver>> &&
+      constructible_from<remove_cvref_t<_Receiver>, _Receiver>;
```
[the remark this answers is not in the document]
ericniebler: I don't know how to check for that from the `receiver` concept. I could see how such a check in `sender_to` might be useful, though.

test/test_common/type_helpers.hpp
```diff
@@ -34,20 +34,20 @@ struct empty_env {};
 //! Check that the value_types of a sender matches the expected type
 template <typename ExpectedValType, typename Env = empty_env, typename S>
 inline void check_val_types(S snd) {
-  using t = typename ex::completion_signatures_of_t<S, Env>::template value_types<type_array, type_array>;
+  using t = typename ex::value_types_of_t<S, Env, type_array, type_array>;
```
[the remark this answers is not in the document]
ericniebler: `value_types_of_t` isn't new. It's been there for many revisions. You could have made this change at any time.

test/test_common/senders.hpp
```diff
@@ -0,0 +1,57 @@
+/*
+ * Copyright (c) Lucian Radu Teodorescu
```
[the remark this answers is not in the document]
ericniebler: Sure you're name. You wrote it. You are entitled to hold copyright. Pretty sure the license requires you to share copyright, though.

PR #411, opened 2022-01-21
[WIP] ports async_scope from libunifex

kirkshoop:
this is not done, but the example works.
there are bugs and missing functionality.
I did update the semantics a bit, this doe not match the libunifex semantics.

empty() completes without latching the scope into the stopped state. The only way to enter the stopped state is to request_stop() on the scope or to destruct the scope while it is empty.

if spawn is called on an empty scope that is not in the stopped state, then the empty event will be reset. 

these semantics allow the scope to be used repeatedly for different sets of spawned senders with joins in between.

empty is a const& method.
async_wait is a const& method.
get_stop_token is a const& method.

the future does not provide split() semantics. I have a plan to add that without adding a method to the scope.


ericniebler (dismissed):
I feel like `async_scope` deserves its own header.

ericniebler:
@kirkshoop I took the liberty of merging `origin/main` into your branch, __uglified the names, and fixed the gcc compile errors. I'd additionally like to factor `async_scope` into its own header and give it an __ugly_name until you are done tweaking the interface. Sounds good?

examples/scope.cpp
```diff
@@ -0,0 +1,107 @@
+/*
+ * Copyright (c) NVIDIA
```
ericniebler, line 2: I'm pretty sure you are free to use your own copyright for code that you wrote, as long as it's all under the same license.

PR #413, opened 2022-01-22
Fix error_types and sends_stopped for the `transfer` family of algos

lucteo:
The `schedule_from` algo has its own completion signals computation that
joins the `error_types` and `sends_stopped` from the input scheduler and
the sender created by the given scheduler.

This fixes also the bad behavior of `transfer` and `transfer_just`, and
helps on fixing some tests based on these.

Adding more tests for completion signatures and `__transform` in the
process of chasing a hard-to-find bug that manifested on downstream
senders (using the `Variant` template parameter multiple times in the
`error_types` definition; depending on the transformation passed to
`error_types`, that leads to strange errors).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -116,6 +116,8 @@ set(test_sourceFiles
     test/algos/adaptors/test_ensure_started.cpp
     test/algos/consumers/test_start_detached.cpp
     test/algos/consumers/test_sync_wait.cpp
+    test/detail/test_utility.cpp
+    test/detail/test_completion_signatures.cpp
     )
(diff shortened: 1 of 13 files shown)


ericniebler:
Thanks for this. I'm going to hold off on merging this since I have already fixed this problem and many more like it in #410, and this PR will add massive merge conflicts. I'd prefer to wait until I get direction from LEWG on the `receiver` concept design changes, and have merged #410.

ericniebler:
Needs a rebase on `main` now that I have merged the receiver changes.

PR #421, opened 2022-01-24
Implementation defined order of `value_types` and `error_types`

lucteo:
Add a note in the spec indicating that the order of variant alternatives
in `value_types` and `error_types` is implementation defined.

### DIFF
--- a/std_execution.bs
+++ b/std_execution.bs
@@ -4459,6 +4459,8 @@ is undefined.
           name the same type as `error_types_of_t<S, no_env, Variant2>`, and
       * `completion_signatures_of_t<S, E>::sends_stopped` shall have the same
           value as `completion_signatures_of_t<S, no_env>::sends_stopped`.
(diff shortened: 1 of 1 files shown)


std_execution.bs
```diff
@@ -4459,6 +4459,8 @@ is undefined.
           name the same type as `error_types_of_t<S, no_env, Variant2>`, and
       * `completion_signatures_of_t<S, E>::sends_stopped` shall have the same
           value as `completion_signatures_of_t<S, no_env>::sends_stopped`.
+10. [<i>Note</i> The order in which <code>Args<i><sub>i</sub></i>...</code> and <code>E<i><sub>i</sub></i>...</code> are captured in `value_types` and `error_types` are implementation defined.
```
villevoutilainen, line 4462: These are requirements for all senders; implementation-defined is thus not quite the right wording tool, because it's not meant to be used with user code, i.e. things that are program-defined. Perhaps just say "The types <code>Args<i><sub>i</sub></i>...<code> and <code>e<i><sub>i</sub></i>...</code> can appear in any order. For example...(the rest of this change)"

lucteo: Done

PR #424, opened 2022-01-24
Fixes/changes a few minor formatting and wording issues

msimberg:
- Fix `<` to `&lt;` in for `execution::just`
- Change `adapt a sender in a sender` to `adapt a sender into a sender` for `execution::on` (the latter is used for `execution::transfer` just after and sounds better to me)
- Change `they are` to `it is` for `execution::schedule_from`


ericniebler:
Thank you!

PR #433, opened 2022-01-26
Move env up

miscco:
We need no_env for the querries, so introduce it first


miscco:
(Note, my editor automatically removes trailing whitespace, I hope that does not screw things up)

ericniebler:
We need to carve P2300 up into tiers, which will involve shuffling everything around. Let's not bother with this one, though I agree with the intent here.

PR #436, opened 2022-01-28
Add bulk, split, ensure_started, and upon* CPOs

senior-zero:
This PR introduces a few CPOs without default implementations: 

- bulk
- split
- ensure_started
- upon_error
- upon_stopped

These names are required to unblock specialized senders development.


ericniebler:
Did you mean for this to merged to `main`? Looks like it's based on the `receiver-redesign` branch.

PR #444, opened 2022-02-03
Remove default from bulk, make it clear no ordering or sequencing is required from customizations.

### LINKED ISSUE #428: Remove default implementation of `bulk`
`bulk` should require customization. Per LEWG request.

griwes:
Resolves #428.

### DIFF
--- a/std_execution.bs
+++ b/std_execution.bs
@@ -5310,34 +5310,11 @@ are all well-formed.
 
         * <i>Mandates:</i> The type of the `tag_invoke` expression above satisfies `execution::sender`.
 
-    3. Otherwise, constructs a sender `s2`. When `s2` is connected with some receiver `out_r`, it:
-
-        1. Constructs a receiver `r`:
-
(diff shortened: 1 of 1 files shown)


ericniebler:
I know LEWG asked for this, but I would like to push back. The default `bulk` implementation meets the semantic, syntactic, and complexity constraints of the `bulk` CPO. _Not_ providing it is anti-generic: a generic algorithm cannot be implemented in terms of `bulk`; it would need to handle the case where `bulk` is unimplemented.

LeeHowes:
I agree, Eric. I dislike this for the same reason I want algorithms to fall back to serial rather than fail to compile deep in someone's implementation when a scheduler changes. Bulk should fall back too to be consistent with that and allow solutions to be fully generic.

ericniebler:
Then what is our course of action? I want to close this PR and go back to LEWG to make the case for keeping the default `bulk` implementation. Once we rip this out, it's only harder to put it back in. Feels?

ericniebler:
Closing per discussion.

PR #445, opened 2022-02-03
Require that the customizations of get_env are noexcept.

### LINKED ISSUE #430: get_env should be noexcept
`execution::get_env` should be specified to be `noexcept`, as this is a fundamental building block function for all the traits queries.
It should thus mandate that the `tag_invoke` expression is `noexcept` if it is customized.

griwes:
Resolves #430.

### DIFF
--- a/std_execution.bs
+++ b/std_execution.bs
@@ -4050,7 +4050,7 @@ namespace std::execution {
 
     1. `tag_invoke(execution::get_env, r)` if that expression is well-formed.
 
-        * <i>Mandates:</i> The decayed type of the above expression is not `no_env`.
+        * <i>Mandates:</i> The above expression is not potentially throwing, and its decayed type is not `no_env`.
 
     2. Otherwise, `get_env(r)` is ill-formed.
 


miscco (approved):

std_execution.bs
```diff
@@ -4050,7 +4050,7 @@ namespace std::execution {
 
     1. `tag_invoke(execution::get_env, r)` if that expression is well-formed.
 
-        * <i>Mandates:</i> The decayed type of the above expression is not `no_env`.
+        * <i>Mandates:</i> The decayed type of the above expression is not `no_env` and `noexcept(tag_invoke(execution::get_env, r))` is `true`.
```
ericniebler, line 4053: Elsewhere we use the phrase: "and the expression XXX is not potentially throwing."

griwes: How about "The above expression is not potentially throwing, and its decayed type is not `no_env`."?

griwes: Done.

ericniebler:
As noted in #430, this has implementation implications. My preference would be for this PR to come with associated changes to the reference implementation, but I would be satisfied with a new issue for that so we don't lose track of it.

ericniebler:
This is now brycelelbach#1

PR #446, opened 2022-02-03
Specify the completion signatures for on.

### LINKED ISSUE #431: specify the completion signatures of the `execution::on` sender

griwes:
Resolves #431.

Also, realign the list elements in the spec of on - the current indentation made no sense to me, and resulted in some very confusing bikeshed behavior when I added the new bullet point. I'm happy to reformat in another way if that would be better.

### DIFF
--- a/std_execution.bs
+++ b/std_execution.bs
@@ -4913,7 +4913,7 @@ are all well-formed.
         6. The lifetime of `op_state2`, once constructed, lasts until either `op_state3` is constructed or `op_state1` is destroyed, whichever comes first. The lifetime of `op_state3`, once constructed, lasts until `op_state1` is destroyed.
 
     3. Given subexpressions `s1` and `e`, where `s1` is a sender returned from `on` or a copy of such, let `S1` be `decltype((s1))`.
-        Let `E'` be <code>decltype((<i>replace-scheduler</i>(e, sch)))</code.
+        Let `E'` be <code>decltype((<i>replace-scheduler</i>(e, sch)))</code>.
         Then the type of `tag_invoke(get_completion_signatures, s1, e)` shall be:
 
         <pre highlight="c++">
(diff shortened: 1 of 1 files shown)


griwes:
I believe that this specifies the same thing as what the reference implementation does, except the reference implementation also inserts an exception_ptr overload of set_error. @ericniebler is there a reason for that?

ericniebler:
> reference implementation also inserts an exception_ptr overload of set_error. @ericniebler is there a reason for that?

Because after the scheduler's sender completes with `set_value`, the input sender needs to be connected and started. `connect` can throw.

std_execution.bs
```diff
@@ -4895,22 +4898,37 @@ are all well-formed.
...
-           `execution::get_scheduler_t`.
+           `execution::get_env(r2)` returns <code><i>replace-scheduler</i>(e, sch)</code>.
+
+        5. When `execution::start` is called on `op_state1`, it calls `execution::start` on `op_state2`.
+
+        6. The lifetime of `op_state2`, once constructed, lasts until either `op_state3` is constructed or `op_state1` is destroyed, whichever comes first. The lifetime of `op_state3`, once constructed, lasts until `op_state1` is destroyed.
 
-      3. When `execution::start` is called on `op_state1`, it calls `execution::start` on `op_state2`.
+    3. Given subexpressions `s1` and `e`, where `s1` is a sender returned from `on` or a copy of such, let `S1` be `decltype((s1))`.
+        Let `E'` be <code>decltype((<i>replace-scheduler</i>(e, sch)))</code.
```
ericniebler, line 4916:
```suggestion
        Let `E'` be <code>decltype((<i>replace-scheduler</i>(e, sch)))</code>.
```

ericniebler (approved):
Needs one markdown fix.

PR #452, opened 2022-02-03
Add Georgy Evtushenko and Lucian Radu Teodorescu as authors.

brycelelbach:


brycelelbach:
@lucteo and @senior-zero please confirm you are okay with being added as authors; we already have consensus among the current authors that we'd like to add you.

brycelelbach:
@lucteo and @senior-zero agreed to be made authors at the 2022-02-03 executors group meeting.

PR #455, opened 2022-02-06
Fix some fallout from the move to std::forward

miscco:


ericniebler:
The changes in the `pre` code blocks are correct. Everything else is not.

miscco:
> The changes in the `pre` code blocks are correct. Everything else i thought i saw them used elsewhere bjt that was only in code Blocks, will change

PR #456, opened 2022-02-06
Fix TODOs in tests

lucteo:
Fixed most of the tests with TODOs. Exception: `sync_wait` and the tests for the not-yet-implemented algorithms


ericniebler (approved):
Wow, THANKS!

PR #460, opened 2022-02-11
Split

senior-zero:
This PR introduces a default implementation of split and fixes a few issues in sync_wait and schedule_from.


include/execution.hpp
```diff
@@ -2186,6 +2437,12 @@ namespace std::execution {
         noexcept(nothrow_tag_invocable<split_t, _Sender>) {
         return tag_invoke(split_t{}, (_Sender&&) __sndr);
       }
+      template <sender _Sender>
+        requires (!__tag_invocable_with_completion_scheduler<split_t, set_value_t, _Sender>) &&
+          (!tag_invocable<split_t, _Sender>)
+      sender auto operator()(_Sender&& __sndr) const {
```
ericniebler, line 2443: I've been burned so many time by automatically deduced return types that I tend to avoid them in code that matters. When I know the type, I make it explicit.

include/execution.hpp
```diff
@@ -2171,7 +2172,257 @@ namespace std::execution {
...
+          if (__state == __state_t::__completed) {
+            __lock.unlock();
+            __self.__propagate_signal();
+          }
+          else {
+            __self.__on_stop_.emplace(
+                get_stop_token(get_env(__self.__recvr_)),
+                __on_stop_requested{__shared_state->__stop_source_});
+
+            if (__shared_state->__stop_source_.stop_requested()) {
```
ericniebler, line 2343: @lewissbaker can you pls weigh in on whether this registration of the stop callback is safe? I think it is, but I wanted to be sure.

include/execution.hpp
```diff
@@ -2171,7 +2172,257 @@ namespace std::execution {
...
+      struct __sh_state : __operation_base {
+        using __nullable_variant_t = __bind_front<__munique<__q<variant>>, tuple<>>;
+        using __error = __error_types_of_t<_Sender,
+                                           __empty_env,
+                                           __bind_front<__q<__variant>,
+                                                        exception_ptr>>;
+        using __value = __value_types_of_t<_Sender,
+                                           __empty_env,
+                                           __q<__decayed_tuple>,
+                                           __nullable_variant_t>;
```
ericniebler, line 2234: I think these last two template parameters are the defaults.

include/execution.hpp
```diff
@@ -2171,7 +2172,257 @@ namespace std::execution {
...
+        using __receiver = __receiver<__value, __error>;
+
+        in_place_stop_source __stop_source_{};
+        vector<__operation_base*> __operation_states_;
+        connect_result_t<_Sender, __receiver> __op_state2_;
+        variant<__value, __error, set_stopped_t> __data_;
```
ericniebler, line 2240: I'm curious why you build a variant of variants instead of just one flat variant.

include/execution.hpp
```diff
@@ -2171,7 +2172,257 @@ namespace std::execution {
...
+                                           __empty_env,
+                                           __bind_front<__q<__variant>,
+                                                        exception_ptr>>;
+        using __value = __value_types_of_t<_Sender,
+                                           __empty_env,
+                                           __q<__decayed_tuple>,
+                                           __nullable_variant_t>;
+        using __receiver = __receiver<__value, __error>;
+
+        in_place_stop_source __stop_source_{};
```
ericniebler, line 2237: I think this needs to be (and according to P2300R4 this is) `std::stop_source`, not `std::in_place_stop_source`. I think the cancellation scopes of the different `split()` senders are non-overlapping. Registration and unregistration of `in_place_stop_callback`s assume the cancellation scopes are strictly nested.

@lewissbaker can correct me if I'm wrong.

include/execution.hpp
```diff
@@ -2171,7 +2172,257 @@ namespace std::execution {
...
+        vector<__operation_base*> __operation_states_;
```
ericniebler, line 2238: I think we can avoid the `vector`. We can have singly-linked intrusive list of operation states instead. You can use the `intrusive_queue` in the examples/detail/ directory. You just need to add a `__operation_base* __next_;` member to `__operation_base`.

include/execution.hpp
```diff
@@ -2171,7 +2172,257 @@ namespace std::execution {
...
+          } catch(...) {
+            execution::set_error((_Receiver&&) __recvr_, current_exception());
+          }
+        }
+
+        template <class... _As>
+          requires (!tag_invocable<set_value_t, _Receiver, decay_t<_As>&...>)
+        void __apply_values(tuple<_As...>&) noexcept {
+          assert(!"_Should never get here");
+        }
```
ericniebler, line 2297: Why do you need this overload?

include/execution.hpp
```diff
@@ -2171,7 +2172,257 @@ namespace std::execution {
...
+          if (__data.index() == 0) {
+            visit([&](auto& __tupl) -> void {
+              __apply_values(__tupl);
+            }, std::get<0>(__data));
+          }
+          if (__data.index() == 1) {
+            visit([&](auto& __err) -> void {
+              execution::set_error((_Receiver&&) __recvr_, __err);
+            }, std::get<1>(__data));
+          }
```
ericniebler, line 2311: The lambdas here can be `noexcept`.

test/algos/adaptors/test_split.cpp
```diff
@@ -19,26 +20,185 @@
...
+}
+TEST_CASE("split advertises completion scheduler", "[adaptors][split]") {
+  inline_scheduler sched;
+
+  SECTION("for value channel")
+  {
+    auto snd = ex::transfer_just(sched, 42) | ex::split();
+    REQUIRE(ex::get_completion_scheduler<ex::set_value_t>(snd) == sched);
+  }
+}
```
ericniebler, line 204: Would be nice to have cancellation tests also. Bit tricky that. I would be ok with filing an issue and doing it later.

ericniebler (changes requested):
@lewissbaker a couple of specific questions for you on this one.

include/execution.hpp
```diff
@@ -2171,7 +2172,257 @@ namespace std::execution {
...
+        using _Sender = __t<_SenderId>;
+        using __sh_state = __sh_state<_Sender>;
+        template <class _Receiver>
+          using __operation = __operation<_SenderId, __x<remove_cvref_t<_Receiver>>>;
+
+        _Sender __sndr_;
+        shared_ptr<__sh_state> __shared_state_;
+
+      public:
+        template <__decays_to<__sender> _Self, receiver _Receiver>
```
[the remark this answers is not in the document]
senior-zero: Thanks for spotting this! I've forgotten to constrain the connect method. Although it's possible to use the constrain you've mentioned, it won't reduce the overload set (since `_receiver` doesn't depend on `_Receiver` in any way). Also, the completion signatures of `__member_t<_Self, _Sender>` and `_Self` are quite different. If you don't mind, I'll use a bit different constrain here:
```
receiver_of<_Receiver, completion_signatures_of_t<_Self>>
```

include/execution.hpp
senior-zero: The last one is different. This difference is not required after redesigning the storage to be a flat vector.

include/execution.hpp
senior-zero: It's not required after redesigning storage to be a flat vector.

ericniebler (changes requested):

include/execution.hpp
```diff
@@ -2171,7 +2173,233 @@ namespace std::execution {
...
+          __sh_state<_Sender> *__shared_state = __self.__shared_state_.get();
+          unique_lock __lock{__shared_state->__mutex_};
+          __state_t __state = __shared_state->__state_;
+
+          if (__state == __state_t::__completed) {
+            __lock.unlock();
+            __self.__propagate_signal();
+          }
+          else {
+            __self.__on_stop_.emplace(
```
lewissbaker, line 2322: Registering the callback here while holding a mutex lock is dangerous.
The callback might be executed inline if a stop-request has already been issued.
This could then result in a call to request_stop() which might then go on to call a callback registered by the underlying operation which can then complete synchronously (if this is the second consumer to call start(), the first consumer already calling start() on the underlying operation-state). If it completes synchronously then it will call back to `__sh_state::__notify()` which will then try to lock the mutex again - which is UB.

Having said that, see the other comment for how to implement this without a mutex to totally avoid this issue.

lewissbaker: The `start()` implementation here would then become something like:
```
void tag_invoke(start_t, __operation& __self) noexcept {
  void* const complete_state = static_cast<void*>(__shared_state.get());
  void* old = __shared_state->__head.load(std::memory_order_acquire);
  if (old != complete_state) {
    __on_stop.emplace(get_stop_token(get_env(__self.__recvr_)), __on_stop_requested{__shared_state->__stop_source_});
  }
  
  do {
    if (old == complete_state) {
      __self.notify();
      return;
    }
    __self.next = static_cast<__operation_base*>(old);
  } while (!__shared_state->__head.compare_exchange_weak(
      old, static_cast<__operation_base*>(__self),
      std::memory_order_release,
      std::memory_order_acquire));
}
```

include/execution.hpp
```diff
@@ -2171,7 +2173,233 @@ namespace std::execution {
...
+                        get_completion_scheduler_t<set_error_t>,
+                        get_completion_scheduler_t<set_stopped_t>>
+          friend auto tag_invoke(_Tag __tag, const __sender& __self, _As&&... __as)
+            noexcept(__nothrow_callable<_Tag, const _Sender&, _As...>)
+            -> __call_result_if_t<__sender_queries::__sender_query<_Tag>, _Tag, const _Sender&, _As...> {
+            return ((_Tag&&) __tag)(__self.__sndr_, (_As&&) __as...);
+          }
+
+        template <class... _Tys>
+        using __set_value_t = completion_signatures<set_value_t(decay_t<_Tys>&...)>;
```
lewissbaker, line 2379: Should it be sending `const T&` to avoid data-races on multiple consumers accessing result concurrently?
Or is `T&` ok here?

lewissbaker (changes requested):

include/execution.hpp
```diff
@@ -2171,7 +2173,233 @@ namespace std::execution {
...
+          __error_types_of_t<
+            _Sender,
+            make_env_t<get_stop_token_t, stop_token>,
+            __transform<
+              __bind_front_q<__decayed_tuple, set_error_t>,
+              __bound_values_t>>;
+
+        using __receiver = __receiver<__variant_t>;
+
+        stop_source __stop_source_{};
```
lewissbaker, line 2246: You should be able to use `in_place_stop_source` here and avoid the heap allocation and ref-counting.

senior-zero: It used to be `in_place_stop_source`, but as @ericniebler rightly [noted](https://github.com/brycelelbach/wg21_p2300_std_execution/pull/460#discussion_r806296821) that P2300 uses `stop_source`. To be on the same page, here's the train of thoughts that justified the usage of the `in_place_stop_source` initially. 
- If `r`'s receiver contract has already been satisfied, an `op_state` won't use stop source.
- Otherwise, `op_state` constructs a callback that uses the stop source. But the lifetime of this callback ends when `op_state` is notified. Since `op_state` shares ownership of `sh_state` (which owns the stop source), the lifetime of the stop source is not an issue. 

What was unclear to me is the case when the stop token is used outside of the `op_state` context, for instance:
```
stop_token token{};

{
  auto snd = ex::read(ex::get_stop_token) | ex::split();
  sync_wait(snd | then([&token](auto t){ token = t; }));
}

if (token.stop_requested()) { // illegal for in place version
  // ...
}
```

The in-place version is not allowed in the code above since any calls to the `stop_requested` must strongly happen before the beginning of the invocation of the destructor of the source. I haven't considered updating the wording. As I understand, to use `in_place_stop_source` in the reference implementation, we'd have to:
- Explicitly state that stop tokens retrieved from the split environment shouldn't outlive`op_state`s (or split sender).
- Change the type of the stop token/source/callback in the wording. 

I suggest we use `stop_token` for now since it matches the current wording. @ericniebler, @lewissbaker if you agree with the steps above, I'll create a separate issue.

lewissbaker: The requirements for get_stop_token() should already be requiring that the token is not used after the receiver it was obtained from has completion signalled.

ericniebler: @lewissbaker's analysis here is correct. Sorry I gave you bad advice earlier.

include/execution.hpp
```diff
@@ -2171,7 +2173,233 @@ namespace std::execution {
...
+        using __sh_state = __sh_state<_Sender>;
+        template <class _Receiver>
+          using __operation = __operation<_SenderId, __x<remove_cvref_t<_Receiver>>>;
+
+        _Sender __sndr_;
+        shared_ptr<__sh_state> __shared_state_;
+
+      public:
+        template <__decays_to<__sender> _Self, receiver _Receiver>
+            requires receiver_of<_Receiver, completion_signatures_of_t<_Self>>
```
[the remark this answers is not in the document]
senior-zero: This seems to be a duplicate of [the discussion above](https://github.com/brycelelbach/wg21_p2300_std_execution/pull/460#discussion_r806818910) where you also suggested:
```cpp
sender_to<__member_t<_Self, _Sender>, __receiver<...>>
```
If there are any points I'm missing, let's continue in the original discussion.

include/execution.hpp
```diff
@@ -2171,7 +2171,222 @@ namespace std::execution {
...
+        using __set_value_t = completion_signatures<set_value_t(const decay_t<_Tys>&...)>;
+
+        template <class _Ty>
+        using __set_error_t = completion_signatures<set_error_t(const decay_t<_Ty>&)>;
+
+        template <__decays_to<__sender> _Self, class _Env>
+          friend auto tag_invoke(get_completion_signatures_t, _Self&&, _Env) ->
+            make_completion_signatures<
+              __member_t<_Self, _Sender>,
+              _Env,
```
ericniebler, line 2375: `_Env` doesn't matter here because the operation is going to be started in an environment that has only `get_stop_token`. So I think this should be:

```suggestion
              __if<
                is_same<_Env, no_env>,
                no_env,
                make_env_t<get_stop_token_t, in_place_stop_token>>
```

include/execution.hpp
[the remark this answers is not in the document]
ericniebler: I think this needs to be at least:

```c++
receiver_of<
  _Receiver,
  completion_signatures_of_t<_Self, __empty_env>>
```

... along with the change below to the customization of `get_completion_signatures`.

ericniebler (commented):
Two more simple fixes and then this is good. I haven't reviewed the lock-free logic, yet. I'll take a harder look on Monday.

include/execution.hpp
lewissbaker: It should just unconditionally be whatever the environment returned from `__split::__receiver` get_env call is.

i.e. it should unconditionally be `make_env_t<get_stop_token_t, in_place_stop_token>`

include/execution.hpp
```diff
@@ -2171,7 +2171,222 @@ namespace std::execution {
...
+              completion_signatures<set_error_t(const exception_ptr&)>,
+              __set_value_t,
+              __set_error_t>;
+
+        explicit __sender(_Sender __sndr)
```
lewissbaker, line 2380: Since the `split()` algorithm might heap allocate, do we need to consider adding support for custom allocators here?
Should we get the allocator from the sender?
And call `std::allocate_shared<__sh_state>(get_allocator(__sndr_), __sndr)`?

Or allow calling `split()` with an extra allocator argument?
And then call `std::allocate_shared<__sh_state>(__alloc, __sndr_)` instead.

include/execution.hpp
```diff
@@ -2171,7 +2171,222 @@ namespace std::execution {
...
+        using __on_stop = optional<typename stop_token_of_t<
+            env_of_t<_Receiver> &>::template callback_type<__on_stop_requested>>;
+
+        _Receiver __recvr_;
+        __on_stop __on_stop_{};
+        shared_ptr<__sh_state<_SenderId>> __shared_state_;
+
+      public:
+        __operation(_Receiver&& __rcvr,
+                    shared_ptr<__sh_state<_SenderId>> __shared_state)
```
lewissbaker, line 2289: Could be marked conditionally noexcept:
```
noexcept(std::is_nothrow_move_constructible_v<_Receiver>)
```

include/execution.hpp
ericniebler: This is better dealt with as an issue.

lewissbaker: #463

include/execution.hpp
```diff
@@ -2171,7 +2171,222 @@ namespace std::execution {
...
+                                          __self.__shared_state_};
+          }
+
+        template <__sender_queries::__sender_query _Tag, class... _As>
+            requires (!__is_instance_of<_Tag, get_completion_scheduler_t>) &&
+              __callable<_Tag, const _Sender&, _As...>
+          friend auto tag_invoke(_Tag __tag, const __sender& __self, _As&&... __as)
+            noexcept(__nothrow_callable<_Tag, const _Sender&, _As...>)
+            -> __call_result_if_t<__sender_queries::__sender_query<_Tag>, _Tag, const _Sender&, _As...> {
+            return ((_Tag&&) __tag)(__self.__sndr_, (_As&&) __as...);
```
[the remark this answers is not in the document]
ericniebler: It is forwarding those sender queries which have specified that they should be forwarded by default. Maybe what you're saying is that you don't think `blocking` should be a forwarding sender query?

include/execution.hpp
```diff
@@ -2171,7 +2171,222 @@ namespace std::execution {
...
+
+          do {
+            if (__old == __completion_state) {
+              __self.__notify();
+              return;
+            }
+            __self.__next_ = static_cast<__operation_base*>(__old);
+          } while (!__head.compare_exchange_weak(
+              __old, static_cast<void *>(&__self),
+              memory_order_acq_rel,
```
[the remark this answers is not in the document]
senior-zero: I agree. This should be `memory_order_release`. I've used `memory_order_acq_rel` to workaround false-positive in clang/tsan 12.0.0-3ubuntu1~20.04.4. I'll use `memory_order_release` for now and add a comment when I have time to write a reproducer.

include/execution.hpp
```diff
@@ -2171,7 +2171,222 @@ namespace std::execution {
...
+      };
+
+    template <class _SenderId>
+      class __sender {
+        using _Sender = __t<_SenderId>;
+        using __sh_state = __sh_state<_SenderId>;
+        template <class _Receiver>
+          using __operation = __operation<_SenderId, __x<remove_cvref_t<_Receiver>>>;
+
+        _Sender __sndr_;
```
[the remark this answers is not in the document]
senior-zero: In this case, I won't be able to forward sender queries.

include/execution.hpp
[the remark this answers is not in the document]
senior-zero: I use sender queries forwarding for collection of temporary storage requirements. I wouldn't like the default split implementation to isolate predecessor senders from queries like that. You have a point on the blocking query, though. Queries on execution guarantees might have to have a specialization in `split`.

include/execution.hpp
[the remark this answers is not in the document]
senior-zero: I'll move it into the shared state for now. Can we consider having sender environments for cases like this one?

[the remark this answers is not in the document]
ericniebler: I agree that this is a problem, but I think it's a problem with the design of `split()` and not this implementation. `split()` needs to decide whether it supports move-only senders or the forwarding of sender queries. Given the current design, it can't do both.

One possibility is to move sender queries into an environment, as with receivers. Then we can copy the environment into the shared state and throw away the sender.

ericniebler (approved):

PR #461, opened 2022-02-17
Give `std::execution::then` proper diagnostics

sithhell:
In the current code, expressions like `std::execution::just() | std::execution::then([](int){});` are being allowed without giving any diagnostics.
What happened instead is that `std::execution::set_value` on the base receiver is called through the fall-through set_value implementation in `receiver_adaptor`.
This happened because the concept constraints could not be fulfilled (`std::invocable<_Fun, _As...>`) and the `get_value` member function is being silently ignored, which might lead to very subtile bugs.

This patch is fixing this situation by not constraining the `set_value` member function and instead giving diagnostics via static_asserts.

### DIFF
--- a/include/execution.hpp
+++ b/include/execution.hpp
@@ -1977,25 +1977,21 @@ namespace std::execution {
         // Customize set_value by invoking the invocable and passing the result
         // to the base class
         template <class... _As>
-          requires invocable<_Fun, _As...> &&
-            tag_invocable<set_value_t, _Receiver, invoke_result_t<_Fun, _As...>>
         void set_value(_As&&... __as) && noexcept try {
-          execution::set_value(
+          static_assert(invocable<Fun, Args...>,
+                      "std::execution::then(Fun) is not invocable with the passed values");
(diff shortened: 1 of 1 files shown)


ericniebler:
The diagnostics problem is real, but removing constraints is the wrong solution. I don't want the `receiver_of` and `sender_to` concepts to give the wrong results.

One thing we can consider doing is eagerly checking `sender_to` when passing a non-dependently-typed sender to `then()`. (When the sender is dependently typed, we can't check `sender_to` until we know the type of the receiver to which the returned sender will be connected.)

sithhell:
You are right. The check needs to happen when computing the dependent completion signature. Without having checked, I think that `sender_to` and `receiver_from` should give the right answer now since `set_value` is ill-formed. You also get a nice static assert failure instead of a cryptic concept constraints failure...

ericniebler:
I'm closing this Thomas since it's the wrong fix for the diagnostics problem. FWIW, the problem you noted about the fall-through `set_value` getting called erroneously in the `receiver_adaptor`, that problem has since been fixed.

PR #478, opened 2022-03-05
Fix sync wait tests

lucteo:


test/algos/consumers/test_sync_wait.cpp
```diff
@@ -24,6 +24,7 @@
 
 #include <thread>
 #include <chrono>
+#include <system_error>
```
ericniebler, line 27: This include should be in the execution.hpp header.

test/algos/consumers/test_sync_wait.cpp
```diff
@@ -154,14 +174,24 @@ optional<tuple<std::string>> tag_invoke(
...
-  auto op = ex::connect(std::move(s), expect_value_receiver_ex{&res});
-  ex::start(op);
-  CHECK(res == "hello");
-  // change the string
-  res = "ciao";
-  return {res};
+struct my_other_string_sender_t {
+  std::string str_;
+
+  using completion_signatures = typename decltype(ex::just(std::string{}))::completion_signatures;
```
ericniebler, line 180: I would prefer if tests did not assume implementation details like this. Either be explicit about the completion signatures, or use `completion_signatures_of_t`.

ericniebler (approved):
Just a few nits, then this looks good, thanks!

test/algos/consumers/test_sync_wait.cpp
ericniebler:
```suggestion
```

test/algos/consumers/test_sync_wait.cpp
ericniebler:
```suggestion
  using completion_signatures = ex::completion_signatures_of_t<decltype(ex::just(std::string{}))>;
```

ericniebler (approved):

Issue #481, opened 2022-03-18
appleclang 13.1 crashes on hello_world example


kirkshoop:
Appears to be a known bug.

https://github.com/llvm/llvm-project/issues/53354

PR #483, opened 2022-03-19
workaround for ICE in appleclang 13.1

kirkshoop:
I tracked the ICE to `__check_type_alias_exists` usage in `stoppable_token`


kirkshoop:
workaround for #481

PR #489, opened 2022-04-22
execution: fix non-virtual-dtor warning

williamspatrick:
Add empty virtual dtor to avoid warning:

execution.hpp:2909:14: error: ‘struct std::execution::__loop::__impl::__task’ has virtual functions and accessible non-virtual destructor [-Werror=non-virtual-dtor]
 2909 |       struct __task {

### DIFF
--- a/include/execution.hpp
+++ b/include/execution.hpp
@@ -2907,6 +2907,7 @@ namespace std::execution {
 
     namespace __impl {
       struct __task {
+        virtual ~__task() {};
         virtual void __execute_() noexcept = 0;
         __task* __next_ = nullptr;
       };


include/execution.hpp
```diff
@@ -2907,6 +2907,7 @@ namespace std::execution {
 
     namespace __impl {
       struct __task {
+        virtual ~__task() {};
```
lewissbaker, line 2910: It should not be necessary to declare the destructor as virtual here - the destructor is never called on a polymorphic pointer to a `__task` base class. The derived type is always declared as final and the owner of this object is responsible for calling the destructor on the derived type.

Adding a virtual destructor is adding an extra entry to every vtable which will never be called.

An alternative approach would be to instead store a function pointer and make `__execute_()` a non-virtual function which calls the function pointer, passing `this` as an argument.

williamspatrick: Would you prefer I open an issue to track the problem then?  As you see, I tried to fix a similar problem in unifex and didn't get good feedback on the "right way".  I'd rather not spend even more time on it here too.

ericniebler:
This should be fixed as of 5dde5e5.

PR #495, opened 2022-05-04
make all operation states non-movable

ericniebler:


test/cpos/test_cpo_start.cpp
```diff
@@ -16,28 +16,28 @@
...
 
-struct my_oper {
+struct my_oper : non_movable {
   bool started_{false};
-
   friend void tag_invoke(ex::start_t, my_oper& self) { self.started_ = true; }
 };
 
-struct op_value {
+struct op_value /*: non_movable*/ {
```
miscco, line 28: Are those comments intentional?

ericniebler: Yeah, the test case fails if the types is immovable. I want to remember that this one is intentionally movable, so I left it commented.

miscco: I would have preferred something along `/* intentionally movable */` but a comment is definitely better than none

ericniebler:
```suggestion
struct op_value /*: non_movable*/ { // Intentionally movable!
```

PR #508, opened 2022-06-09
Feature/improve conan recipe

paulbendixen:
Updates the conanfile.py to have a better package.

By setting it up to be a header only-package, maintaining it is much easier.

We might have to update the compiler flags in the `package_info` function, but this can easily be added now there is an example.


ericniebler:
Is this one ready to go?

conanfile.py
```diff
@@ -2,57 +2,33 @@
...
+        tools.check_min_cppstd(self,"20")
+
+    def set_version(self):
+        # Get the version from the spec file
+        content = tools.load(path.join(self.recipe_folder, "std_execution.bs"))
+        rev = re.search(r"Revision: (\d+)", content).group(1).strip()
+        self.version = f"0.{rev}.0"
+
+    def package(self):
+        self.copy("*.hpp")
```
ericniebler, line 29: Would this pick up `.hpp` files under `test/`?

conanfile.py
```diff
@@ -2,57 +2,33 @@
...
-
-   def package_info(self):
-      self.cpp_info.libs = self.collect_libs()
-
-   def _configure_cmake(self):
-      cmake = CMake(self)
-      if self.settings.compiler == "Visual Studio" and self.options.shared:
-         cmake.definitions["CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS"] = True
-      cmake.configure(source_folder=None)
-      return cmake
```
ericniebler, line 56: I'm not the original author of this conan recipe, and I'm no conan expert by a long-shot, but I'm a bit concerned that we're dropping these cmake bits. Can you say why they're not needed?

@lucteo, can you weigh in on this diff?

paulbendixen: The approach to packaging is different.
The original packaging strategy is to compile and export, but since the library is header only, this is overkill.
Normally you wouldn't have the compiler as settings in a header only library, but since different libraries need different flags to compile (e.g. -fconcepts) I have left it in.

I would love some more comments.

paulbendixen:
> Is this one ready to go?

I'm happy with it, it should be generally usable, but I've only tested with GCC (my clang is not new enough and I'm on Linux so MSVC is kinda out)

ericniebler:
Thanks

PR #510, opened 2022-06-15
Port sender creation utility, `create<completions...>(lambda)` from libunifex

ericniebler:
`create` is a utility for creating a sender from a lambda. It is especially useful when wrapping legacy C-style async APIs in senders.

For example, a C-style async API such as the following:

```c++
using do_thing_callback_t = void(*)(void*, error_code, ResultT);

extern "C" void async_do_thing(void* context, int a, const char* b, do_thing_callback_t callback);
```

Can be wrapped in a sender as follows:

```c++
sender_of<int> auto async_do_thing(int a, const char* b) {
  return create<set_value_t(ResultT), set_error_t(error_code)>(
    [a, b]<class Context>(Context& context) noexcept {
      sync_do_thing(&context, a, b, [](void* pv, error_code err, ResultT result) noexcept {
        auto& context = *static_cast<Context*>(pv);
        if (err) {
          set_error(move(context.receiver), err);
        } else {
          set_value(move(context.receiver), result);
        }
      });
    };
  );
}
```

Ultimately, this will get its own P-numbered paper.

### DIFF
--- a/test/detail/test_create.cpp
+++ b/test/detail/test_create.cpp
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) Lucian Radu Teodorescu
+ * Copyright (c) NVIDIA
  *
  * Licensed under the Apache License Version 2.0 with LLVM Exceptions
  * (the "License"); you may not use this file except in compliance with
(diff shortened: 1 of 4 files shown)


test/detail/test_create.cpp
```diff
@@ -0,0 +1,105 @@
+/*
+ * Copyright (c) Lucian Radu Teodorescu
```
ericniebler, line 2:
```suggestion
 * Copyright (c) NVIDIA
```

PR #513, opened 2022-06-21
change nest() to delay taking a hold on the async_scope until start

kirkshoop:
change nest() to return a sender that delays taking a hold on the async_scope until start()

### DIFF
--- a/include/async_scope.hpp
+++ b/include/async_scope.hpp
@@ -257,7 +257,7 @@ namespace std::execution::P2519 {
       using __future_operation_t =
         connect_result_t<_Sender, __future_receiver<__x<_Sender>>>;
 
-    bool __try_record_start_(async_scope*) noexcept;
+    void __record_start_(async_scope*) noexcept;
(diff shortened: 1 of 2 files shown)


LeeHowes:
Am I correct that in this design it's UB to call `start()` on the result of nest if the scope has gone away?

kirkshoop:
> Am I correct that in this design it's UB to call `start()` on the result of nest if the scope has gone away?

Yes. Just like calls to spawn.. calls to nest()->connect()->start() must occur while the async_scope object is alive.

LeeHowes:
Yes, but `connect()`/`start()` is tightly coupled in library code for the most part. `nest()` is more likely to be used by the user. So it's a reference capturing operation that hides the reference capture. I liked the option to `terminate`. The silent reference capture bothers me more. Hopefully ASAN would catch errors consistently.

ericniebler:
C++ has this problem in spades, and this is nothing new. Just consider iterators. I don't think the possibility of a dangling sender is a problem.

Issue #466, opened 2022-02-21
`split()` can't support both move-only senders and forwarding sender queries


ericniebler:
Currently the sender is copied into each of the split sender so that the split senders can respond to sender queries by forwarding them to the (copy of the) original sender. However, this precludes `split()` being used with move-only/single-shot senders, which is one of its most important use cases.

One possibility is to move sender queries into an environment, as with receivers. Then we can copy the environment into the shared state and throw away the sender.

See also #641

PR #519, opened 2022-06-27
Store 'split' sender in shared state

ccotter:
I was looking to get a version of "split" working on top of libunifex, but found the reference implementation of `split` here did not allow me to fork a `unifex::task` (e.g., as with the example in https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2022/p2300r5.html#design-forkable).

I didn't quite follow https://github.com/brycelelbach/wg21_p2300_std_execution/issues/466 as I'm not sure I fully understand what environments are, but I can look into that approach as well.

### DIFF
--- a/include/execution.hpp
+++ b/include/execution.hpp
@@ -2260,13 +2260,15 @@ namespace std::execution {
 
         using __receiver_ = __receiver<__sh_state>;
 
+        _Sender __sndr_;
         in_place_stop_source __stop_source_{};
         connect_result_t<_Sender, __receiver_> __op_state2_;
         __variant_t __data_;
         atomic<void*> __head_;
 
-        explicit __sh_state(_Sender& __sndr)
-          : __op_state2_(connect((_Sender&&) __sndr, __receiver_{*this}))
+        explicit __sh_state(_Sender __sndr)
+          : __sndr_((_Sender&&)__sndr)
+          ,  __op_state2_(connect((_Sender&&) __sndr_, __receiver_{*this}))
           , __head_{nullptr}
         {}
 
@@ -2358,7 +2360,6 @@ namespace std::execution {
         template <class _Receiver>
           using __operation = __operation<_SenderId, __x<remove_cvref_t<_Receiver>>>;
 
-        _Sender __sndr_;
         shared_ptr<__sh_state_> __shared_state_;
 
       public:
@@ -2395,9 +2396,9 @@ namespace std::execution {
               __set_value_t,
               __set_error_t>;
 
-        explicit __sender(_Sender __sndr)
-            : __sndr_((_Sender&&) __sndr)
-            , __shared_state_{make_shared<__sh_state_>(__sndr_)}
+        template <class _Sender2>
+        explicit __sender(_Sender2&& __sndr)
+            : __shared_state_{make_shared<__sh_state_>((_Sender2&&)__sndr)}
         {}
       };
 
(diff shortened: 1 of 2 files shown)


ericniebler:
> I didn't quite follow #466 as I'm not sure I fully understand what environments are, but I can look into that approach as well.

An environment is basically just a bag of queryable properties. Today, all sender queries are queries on the sender itself. I think what I was suggesting in #446 is that have _one_ sender query: `get_env(sndr)`, and all of the other sender queries then become queries on the sender's environment. 

In that world, `split()` would store a `variant<InputSender, env_of_t<InputSender>>` in the shared state which initially stores the input sender. When `connect`-ing the input sender, the sender's environment is saved in a local, which is then moved into the `variant`, destroying the input sender.

Every copy of the `split()` sender would hold a strong reference to the shared state. Queries on those senders' environments would forward to the `variant` in the shared state.

In this way, move-only and single-shot senders could be supported while also preserving the ability of issuing sender queries on the resulting `split()` senders.

ccotter:
Thanks for the explanation. I understand this better now.

When we preserve the value sent by the input sender, the returned split sender sends `const&` lvalues - which is the necessary sort of "hoisting" required to make a move-only sender value accessible by multiple consumers. It's not clear how sender queries could be "hoisted" or changed to provide forwarded queries in a generic way, even with a sender environment concept. I suppose the similar notion of *const lvalue*-ifying the sender environment could work?

ericniebler:
I think we would have to treat queries like the values and errors of a single-shot sender: on the first call to fetch that value, cache it in the operation state and return a `const&` to the cached value. We could avoid doing that if the query is already returning a reference.

The problem with this approach is that there is no way to enumerate the set of sender queries, and thus no way to compute how big of a cache the operation state needs to cache the results of the queries.

ccotter:
Thanks for explaining! My curiosity is satisfied.

ericniebler:
I'm curious why you closed this.

ccotter:
Ah, I should have asked instead of assuming - where do we see a `split` that can support move only input senders, one where there are the deficiencies with sender queries as noted above? Would that still be useful in the meantime? I'll reopen as a draft for the time being, and with some more clarity, I'm happy to work to evolve this change to document what's supported (and not) via tests, etc.

msimberg:
Do I understand correctly that with this both senders and receivers would be environment providers? Would there be differences in the kinds of environments they provide or would they both be just "bags of things that can be queried"?

> I think we would have to treat queries like the values and errors of a single-shot sender: on the first call to fetch that value, cache it in the operation state and return a `const&` to the cached value. We could avoid doing that if the query is already returning a reference.
> 
> The problem with this approach is that there is no way to enumerate the set of sender queries, and thus no way to compute how big of a cache the operation state needs to cache the results of the queries.

I'm not sure if I'm misunderstanding, but are there cases where it wouldn't be enough to only store the environment of the wrapped sender in the shared state and then `get_env(split_sender)` simply returns a (possibly const) reference to that environment? Queries from successor senders would then be done on the reference to that environment. In other words, I'm missing why the individual queries would have to be cached rather than the environment as a whole?

Finally, I've seen mentions of moving away from customization based on `get_completion_scheduler`. Does that reduce the need for getting an environment from a sender? Probably does not remove the need though since there can be user-defined queries on senders...

ericniebler:
@ccotter the committee is moving ahead with putting the sender queries into a separate "attributes" object, which changes how this issue will be fixed. Feel like taking a crack at that change first? See brycelelbach/wg21_p2300_execution#17 for the details.

ccotter:
Will do!

ericniebler:
Hi @ccotter, I'm closing this PR for now. I expect you'll want to fix this issue in a new PR as part of integrating the `get_attrs` changes.

PR #520, opened 2022-07-02
Allow calling sync_wait_with_variant

ccotter:
Declare the sync_wait CPO before sync_wait_with_variant attempts to call
it.


include/execution.hpp
```diff
@@ -4279,15 +4279,15 @@ namespace std::this_thread {
         nothrow_tag_invocable<sync_wait_with_variant_t, _Sender>) {
         return tag_invoke(sync_wait_with_variant_t{}, (_Sender&&) __sndr);
       }
-      template <execution::__single_value_variant_sender<__impl::__env> _Sender>
+      template <execution::sender _Sender>
```
ericniebler, line 4282: Maybe?

```suggestion
      template <execution::sender<__impl::__env> _Sender>
```

I'm on my phone, I'll need to double check this. Also, if you change this constraint here, do you need to change it for the other overloads?

ericniebler:
Better late than never? Sorry this took me so long.

ccotter:
thanks! I actually just came back to rebase against upstream and just realized this was merged

PR #523, opened 2022-07-06
Add upon_error and upon_stopped implementations.

gonidelis:
Adds `upon_stopped` and `upon_error` implementations with the corresponding testing.

### DIFF
--- a/include/execution.hpp
+++ b/include/execution.hpp
@@ -2205,7 +2205,6 @@ namespace std::execution {
       };
 
     struct upon_error_t {
-
       template <class _Sender, class _Fun>
         using __sender = __sender<__x<remove_cvref_t<_Sender>>, __x<remove_cvref_t<_Fun>>>;
 
@@ -2289,7 +2288,6 @@ namespace std::execution {
       struct __sender {
         using _Sender = __t<_SenderId>;
         using _Fun = __t<_FunId>;
-        friend sender_adaptor<__sender, _Sender>;
         template <class _Receiver>
           using __receiver = __receiver<__x<remove_cvref_t<_Receiver>>, _FunId>;
 
@@ -2334,14 +2332,13 @@ namespace std::execution {
       };
 
     struct upon_stopped_t {
-
       template <class _Sender, class _Fun>
         using __sender = __sender<__x<remove_cvref_t<_Sender>>, __x<remove_cvref_t<_Fun>>>;
       
       template <sender _Sender, __movable_value _Fun>
         requires 
-        __tag_invocable_with_completion_scheduler<upon_stopped_t, set_stopped_t, _Sender, _Fun> &&
-        __callable<_Fun>
+          __tag_invocable_with_completion_scheduler<upon_stopped_t, set_stopped_t, _Sender, _Fun> &&
+          __callable<_Fun>
       sender auto operator()(_Sender&& __sndr, _Fun __fun) const
         noexcept(nothrow_tag_invocable<upon_stopped_t, __completion_scheduler_for<_Sender, set_stopped_t>, _Sender, _Fun>) {
         auto __sched = get_completion_scheduler<set_stopped_t>(__sndr);
@@ -2349,20 +2346,19 @@ namespace std::execution {
       }
       template <sender _Sender, __movable_value _Fun>
         requires (!__tag_invocable_with_completion_scheduler<upon_stopped_t, set_stopped_t, _Sender, _Fun>) &&
-          tag_invocable<upon_stopped_t, _Sender, _Fun> && invocable<_Fun> && __callable<_Fun>
+          tag_invocable<upon_stopped_t, _Sender, _Fun> && __callable<_Fun>
       sender auto operator()(_Sender&& __sndr, _Fun __fun) const
         noexcept(nothrow_tag_invocable<upon_stopped_t, _Sender, _Fun>) {
         return tag_invoke(upon_stopped_t{}, (_Sender&&) __sndr, (_Fun&&) __fun);
       }
       template <sender _Sender, __movable_value _Fun>
         requires (!__tag_invocable_with_completion_scheduler<upon_stopped_t, set_stopped_t, _Sender, _Fun>) &&
-          (!tag_invocable<upon_stopped_t, _Sender, _Fun>) && __callable<_Fun>
-          && sender<__sender<_Sender, _Fun>>
+          (!tag_invocable<upon_stopped_t, _Sender, _Fun>) && __callable<_Fun> &&
+          sender<__sender<_Sender, _Fun>>
       __sender<_Sender, _Fun> operator()(_Sender&& __sndr, _Fun __fun) const {
         return __sender<_Sender, _Fun>{(_Sender&&) __sndr, (_Fun&&) __fun};
       }
-      template <class _Fun>
-        requires __callable<_Fun>
+      template <__callable _Fun>
       __binder_back<upon_stopped_t, _Fun> operator()(_Fun __fun) const {
         return {{}, {}, {(_Fun&&) __fun}};
       }
@@ -2377,7 +2373,7 @@ namespace std::execution {
     struct bulk_t {
       template <sender _Sender, integral _Shape, __movable_value _Fun>
         requires __tag_invocable_with_completion_scheduler<bulk_t, set_value_t, _Sender, _Shape, _Fun> 
-        sender auto operator()(_Sender&& __sndr, _Shape __shape, _Fun __fun) const
+      sender auto operator()(_Sender&& __sndr, _Shape __shape, _Fun __fun) const
         noexcept(nothrow_tag_invocable<bulk_t, __completion_scheduler_for<_Sender, set_value_t>, _Sender, _Shape, _Fun>) {
         auto __sched = get_completion_scheduler<set_value_t>(__sndr);
         return tag_invoke(bulk_t{}, std::move(__sched), (_Sender&&) __sndr, (_Shape&&) __shape, (_Fun&&) __fun);
--- a/include/execution.hpp
+++ b/include/execution.hpp
@@ -2112,10 +2112,11 @@ namespace std::execution {
   /////////////////////////////////////////////////////////////////////////////
   // [execution.senders.adaptors.upon_error]
   namespace __upon_error {
-    template <class _ReceiverId, class _Fun>
+    template <class _ReceiverId, class _FunId>
       class __receiver
-        : receiver_adaptor<__receiver<_ReceiverId, _Fun>, __t<_ReceiverId>> {
+        : receiver_adaptor<__receiver<_ReceiverId, _FunId>, __t<_ReceiverId>> {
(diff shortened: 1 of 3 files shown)


miscco (commented):
Some minor nits

gonidelis:
@miscco Thanks, will take a look.

include/execution.hpp
```diff
@@ -2093,7 +2093,100 @@ namespace std::execution {
...
+       public:
+        explicit __receiver(_Receiver __rcvr, _Fun __fun)
+          : receiver_adaptor<__receiver, _Receiver>((_Receiver&&) __rcvr)
+          , __f_((_Fun&&) __fun)
+        {}
+      };
+
+    template <class _SenderId, class _Fun>
+      class __sender
+        : sender_adaptor<__sender<_SenderId, _Fun>, __t<_SenderId>> {
```
ericniebler, line 2141: Hrm. On the `main` branch, the `then` sender (far above) does _not_ inherit from `sender_adaptor`, so it looks like you've haven't based this work on the most recent commit. I changed `then` sender's `connect` customization from a member function to `tag_invoke` so as to be sensitive to the cv- and ref-qualification of the sender. You'll need to make similar changes to the `upon_error` and `upon_stopped` senders.

See the changes I made in 1952036 for reference.

ericniebler (changes requested):
Looking very good. A few suggestions...

include/execution.hpp
```diff
@@ -2120,7 +2221,99 @@ namespace std::execution {
...
+        } catch(...) {
+          execution::set_error(
+              ((__receiver&&) *this).base(),
+              current_exception());
+        }
+        // Handle the case when the invocable returns void
+        template <class _R2 = _Receiver>
+          requires invocable<_Fun> &&
+            same_as<void, invoke_result_t<_Fun>> &&
+              __callable<set_value_t, _R2>
```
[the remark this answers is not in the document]
miscco: That type of constraint checking is happening a lot. In the future we should provide a helper concept that subsumes both void and non void checks and then internally dispatch via `if constexpr`

No change requested here

include/execution.hpp
```diff
@@ -2120,7 +2221,99 @@ namespace std::execution {
...
+          using __set_stopped =
+            completion_signatures<
+              __minvoke1<
+                __remove<void, __qf<set_value_t>>,
+                invoke_result_t<_Fun>>>;
+
+        template <class _Env>
+          using __completion_signatures =
+            make_completion_signatures<
+              _Sender, _Env, __with_exception_ptr,
```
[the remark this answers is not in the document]
miscco: I was wondering about the requirement on `set_error_t(exception_ptr)` to be valid.

I *think* we can inside the catch bock add an `if constexpr` that checks whether `set_error_t` is `tag_invocable` with whatever exception we got

include/execution.hpp
miscco: From a user perspective this is bad. We should be able to reuse `sender_adaptor` here and everywhere else too.

Given that this is about qualification of the sender, I have some high hopes that `deduce this` will give us a way out of this.

No change requested

gonidelis:
@ericniebler  The changes regarding `_Fun` to be callable instead of `tag_invocable` should also be propagated to `upon_error` too, right?

include/execution.hpp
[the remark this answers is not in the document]
ericniebler: You can't replace these overloads with a single function that does `if constexpr` unfortunately. With `if constexpr`, if neither condition is satisfied, it's a hard error. You don't want a hard error. You want both overloads to be SFINAE-d away.

include/execution.hpp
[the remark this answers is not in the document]
ericniebler: We wouldn't want to do that since it might try to instantiate the error channel. What we want is to avoid instantiating _any_ channel that will be unused, based on the sender's completion signatures.

include/execution.hpp
ericniebler: Yes, "deduced `this`" makes the adaptors much more useful.

include/execution.hpp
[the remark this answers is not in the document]
miscco: I meant to create a combined constraint 
```cpp
template<class _Fun, class _Receiver>
concept _Continuation = (same_as<void, __call_result_t<_Fun>> && __callable<set_value_t, _Receiver>) || __callable<set_value_t, _Receiver, __call_result_t<_Fun>>;
```

And then dispatch inside that constrained function depending on `same_as<void, __call_result_t<_Fun>>`

gonidelis:
@ericniebler  Thanks a lot! I kinda figured while working on it today.

include/execution.hpp
```diff
@@ -2112,7 +2112,103 @@ namespace std::execution {
...
+          requires __callable<_Tag, const _Sender&, _Err>
+        friend auto tag_invoke(_Tag __tag, const __sender& __self, _Err&& __err)
+          noexcept(__nothrow_callable<_Tag, const _Sender&, _Err>)
+          -> __call_result_if_t<__sender_queries::__sender_query<_Tag>, _Tag, const _Sender&, _Err> {
+          return ((_Tag&&) __tag)(__self.__sndr_, (_Err&&) __err);
+        }
+      };
+
     struct upon_error_t {
+
```
ericniebler, line 2208:
```suggestion
```

include/execution.hpp
```diff
@@ -2139,22 +2243,126 @@ namespace std::execution {
...
+          : receiver_adaptor<__receiver, _Receiver>((_Receiver&&) __rcvr)
+          , __f_((_Fun&&) __fun)
+        {}
+      };
+
+    template <class _SenderId, class _FunId>
+      struct __sender {
+        using _Sender = __t<_SenderId>;
+        using _Fun = __t<_FunId>;
+        friend sender_adaptor<__sender, _Sender>;
```
ericniebler, line 2292:
```suggestion
```

include/execution.hpp
```diff
@@ -2139,22 +2243,126 @@ namespace std::execution {
...
+          requires __callable<_Tag, const _Sender&>
+        friend auto tag_invoke(_Tag __tag, const __sender& __self)
+          noexcept(__nothrow_callable<_Tag, const _Sender&>)
+          -> __call_result_if_t<__sender_queries::__sender_query<_Tag>, _Tag, const _Sender&> {
+          return ((_Tag&&) __tag)(__self.__sndr_);
+        }
+      };
+
     struct upon_stopped_t {
+
```
ericniebler, line 2337:
```suggestion
```

include/execution.hpp
```diff
@@ -2139,22 +2243,126 @@ namespace std::execution {
...
+      template <class _Sender, class _Fun>
+        using __sender = __sender<__x<remove_cvref_t<_Sender>>, __x<remove_cvref_t<_Fun>>>;
+      
       template <sender _Sender, __movable_value _Fun>
-        requires __tag_invocable_with_completion_scheduler<upon_stopped_t, set_stopped_t, _Sender, _Fun>
+        requires 
+        __tag_invocable_with_completion_scheduler<upon_stopped_t, set_stopped_t, _Sender, _Fun> &&
+        __callable<_Fun>
```
ericniebler, line 2344:
```suggestion
          __tag_invocable_with_completion_scheduler<upon_stopped_t, set_stopped_t, _Sender, _Fun> &&
          __callable<_Fun>
```

include/execution.hpp
```diff
@@ -2139,22 +2243,126 @@ namespace std::execution {
...
       sender auto operator()(_Sender&& __sndr, _Fun __fun) const
         noexcept(nothrow_tag_invocable<upon_stopped_t, __completion_scheduler_for<_Sender, set_stopped_t>, _Sender, _Fun>) {
         auto __sched = get_completion_scheduler<set_stopped_t>(__sndr);
         return tag_invoke(upon_stopped_t{}, std::move(__sched), (_Sender&&) __sndr, (_Fun&&) __fun);
       }
       template <sender _Sender, __movable_value _Fun>
         requires (!__tag_invocable_with_completion_scheduler<upon_stopped_t, set_stopped_t, _Sender, _Fun>) &&
-          tag_invocable<upon_stopped_t, _Sender, _Fun>
+          tag_invocable<upon_stopped_t, _Sender, _Fun> && invocable<_Fun> && __callable<_Fun>
```
ericniebler, line 2352:
```suggestion
          tag_invocable<upon_stopped_t, _Sender, _Fun> && __callable<_Fun>
```

include/execution.hpp
```diff
@@ -2139,22 +2243,126 @@ namespace std::execution {
...
       sender auto operator()(_Sender&& __sndr, _Fun __fun) const
         noexcept(nothrow_tag_invocable<upon_stopped_t, _Sender, _Fun>) {
         return tag_invoke(upon_stopped_t{}, (_Sender&&) __sndr, (_Fun&&) __fun);
       }
+      template <sender _Sender, __movable_value _Fun>
+        requires (!__tag_invocable_with_completion_scheduler<upon_stopped_t, set_stopped_t, _Sender, _Fun>) &&
+          (!tag_invocable<upon_stopped_t, _Sender, _Fun>) && __callable<_Fun>
+          && sender<__sender<_Sender, _Fun>>
```
ericniebler, line 2360:
```suggestion
          (!tag_invocable<upon_stopped_t, _Sender, _Fun>) && __callable<_Fun> &&
          sender<__sender<_Sender, _Fun>>
```

include/execution.hpp
```diff
@@ -2139,22 +2243,126 @@ namespace std::execution {
...
+      __sender<_Sender, _Fun> operator()(_Sender&& __sndr, _Fun __fun) const {
+        return __sender<_Sender, _Fun>{(_Sender&&) __sndr, (_Fun&&) __fun};
+      }
       template <class _Fun>
+        requires __callable<_Fun>
```
ericniebler, line 2365:
```suggestion
      template <__callable _Fun>
```

include/execution.hpp
```diff
@@ -2168,8 +2376,8 @@ namespace std::execution {
   namespace __bulk {
     struct bulk_t {
       template <sender _Sender, integral _Shape, __movable_value _Fun>
-        requires __tag_invocable_with_completion_scheduler<bulk_t, set_value_t, _Sender, _Shape, _Fun>
-      sender auto operator()(_Sender&& __sndr, _Shape __shape, _Fun __fun) const
+        requires __tag_invocable_with_completion_scheduler<bulk_t, set_value_t, _Sender, _Shape, _Fun> 
+        sender auto operator()(_Sender&& __sndr, _Shape __shape, _Fun __fun) const
```
ericniebler, line 2380:
```suggestion
      sender auto operator()(_Sender&& __sndr, _Shape __shape, _Fun __fun) const
```

ericniebler (approved):
Looks great. 👍

include/execution.hpp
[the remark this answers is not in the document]
ericniebler: Yeah I get it, pardon the brain fart.

PR #526, opened 2022-07-08
Minor CMakeLists.txt enhancements

msimberg:
The two primary goals for this PR are:
1. Allowing building without Conan.
2. Allowing installation of the library (headers and CMake configuration files).

For 1. I've added a CMake option `P2300_ENABLE_CONAN` which is on by default to keep the setup working as before for those of you who are using Conan. However, as I don't use Conan myself I'd be grateful if someone who does use it could check that I'm not breaking anything. At least the CI still works, but I don't know how comprehensively it tests the Conan stuff. @paulbendixen I'm pinging you since you've recently made changes to the Conan integration.

The background for this is that I'd like to add a P2300 package to spack (https://github.com/spack/spack), and the Conan integration is unnecessary there.

Minor related changes in this PR:
- Adds CMake options to disable tests and examples. They are on by default.
- Removes the `-dev` suffix from the project name. I'm happy to revert this if you prefer to have it there, but the CMake target namespace should correspond to the project name, and the exported target should in that case be `P2300-dev::P2300`.
- I'm now always setting the version based on the revision in `std_execution.bs` similarly to how it's done in `conanfile.py`, i.e. R6 becomes 0.6.0. This allows the version to be set correctly even when the Conan integration is disabled.
- Removes an unnecessary (I think) target alias. Or is this required for the Conan setup?
- Moves executable-specific flags to a separate `INTERFACE` target. Flags like `-Wall`, `-Wno-non-template-friend`, `-ftemplate-backtrace-limit=0`, and `-fcolor-diagnostics` should not be forced on consumers of the library, but only be set for tests and examples.
- Fix `message(ERROR ...)` to `message(FATAL_ERROR ...)`. The former just prints a literal `ERROR`.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -101,27 +101,44 @@ target_link_libraries(p2300 INTERFACE Threads::Threads)
 # Use C++20 standard
 target_compile_features(p2300 INTERFACE cxx_std_20)
 
-# Turn all warnings
+# Enable coroutines for GCC
 target_compile_options(p2300 INTERFACE
+                       $<$<CXX_COMPILER_ID:GNU>:-fcoroutines>
+                       )
+
+add_library(P2300::p2300 ALIAS p2300)
+
+# Don't require building everything when installing
+set(CMAKE_SKIP_INSTALL_ALL_DEPENDENCY ON)
+
+# Support target for examples and tests
+add_library(p2300_executable_flags INTERFACE)
+
+# Enable warnings
+target_compile_options(p2300_executable_flags INTERFACE
                        $<$<OR:$<CXX_COMPILER_ID:Clang>,$<CXX_COMPILER_ID:AppleClang>,$<CXX_COMPILER_ID:GNU>>:
                        -Wall>
                        $<$<CXX_COMPILER_ID:MSVC>:
                        /W4>)
-# template backtrace limit
-target_compile_options(p2300 INTERFACE
+
+
+# Silence warnings with GCC
+target_compile_options(p2300_executable_flags INTERFACE
+                       $<$<CXX_COMPILER_ID:GNU>:-Wno-non-template-friend>
+                       )
+
+# Template backtrace limit
+target_compile_options(p2300_executable_flags INTERFACE
                        $<$<OR:$<CXX_COMPILER_ID:Clang>,$<CXX_COMPILER_ID:AppleClang>,$<CXX_COMPILER_ID:GNU>>:
                        -ftemplate-backtrace-limit=0>
                        )
 
-if ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU")
-    add_compile_options(-fdiagnostics-color=always)
-    add_compile_options(-Wno-non-template-friend)
-    add_compile_options(-fcoroutines)
-elseif ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang")
-    add_compile_options(-fcolor-diagnostics)
-endif ()
-
-add_library(P2300::p2300 ALIAS p2300)
+# Always enable colored output
+target_compile_options(p2300_executable_flags INTERFACE
+                       $<$<OR:$<CXX_COMPILER_ID:Clang>,$<CXX_COMPILER_ID:AppleClang>>:
+                       -fcolor-diagnostics>
+                       $<$<CXX_COMPILER_ID:GNU>:-fdiagnostics-color=always>
+                       )
 
 # Now, set up test executable
 enable_testing()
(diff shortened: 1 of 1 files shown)


msimberg:
I was too slow to comment on #590 but really happy to see that it was merged. I've rebased this on main and left changes that I think would still be useful to have in the CMake setup. In no particular order:

- Parse the version number from `std_execution.bs` as is done in the `conanfile.py`
- Don't set `CMAKE_CXX_STANDARD`. This is unnecessary in my understanding and already taken care of by `cxx_std_20`. The user is still free to set `CMAKE_CXX_STANDARD` to 20 or higher if they wish, but it doesn't _need_ to be set.
- Use `CMAKE_INSTALL_INCLUDEDIR` in the `INSTALL_INTERFACE` to be consistent with the install command later on.
- Move "convenience" flags to only be applied on the tests/executables. They should not be applied to depending projects. The only flag that has to be in the target interface is `-fcoroutines`.

@trxcllnt I'd be grateful for your comments on this since you made #590.

trxcllnt:
@msimberg thanks for rebasing! I saw this PR after the other one was merged, and I liked all the things you added. I'll review it today.

bustercopley:
@msimberg, the `p2300_executable_flags`, especially `-Wno-non-template-friends`, would be nice to have for compiling the examples, as well as the tests. Also, `$<BOOL:$<CXX_COMPILER_ID:GNU>>` could be expressed more simply as `$<CXX_COMPILER_ID:GNU>`, which is the same boolean value.

CMakeLists.txt
```diff
@@ -97,35 +93,55 @@ set_target_properties(p2300 PROPERTIES
...
 target_link_libraries(p2300 INTERFACE Threads::Threads)
 
 # Use C++20 standard
 target_compile_features(p2300 INTERFACE cxx_std_20)
 
-# Turn all warnings
+# Enable coroutines for GCC
 target_compile_options(p2300 INTERFACE
+                       $<$<BOOL:$<CXX_COMPILER_ID:GNU>>:
+                       -fcoroutines>
```
trxcllnt, line 107:
```suggestion
                       $<$<CXX_COMPILER_ID:GNU>:-fcoroutines>
```

CMakeLists.txt
```diff
@@ -97,35 +93,55 @@ set_target_properties(p2300 PROPERTIES
...
-    add_compile_options(-fcolor-diagnostics)
-endif ()
-
-add_library(P2300::p2300 ALIAS p2300)
+# Always enable colored output
+target_compile_options(p2300_executable_flags INTERFACE
+                       $<$<OR:$<CXX_COMPILER_ID:Clang>,$<CXX_COMPILER_ID:AppleClang>>:
+                       -fcolor-diagnostics>
+                       $<$<BOOL:$<CXX_COMPILER_ID:GNU>>:
+                       -fdiagnostics-color=always>
```
trxcllnt, line 143:
```cmake
$<$<CXX_COMPILER_ID:GNU>:-fdiagnostics-color=always>
```

CMakeLists.txt
```diff
@@ -97,35 +93,55 @@ set_target_properties(p2300 PROPERTIES
 # Declare the public include directories
 target_include_directories(p2300 INTERFACE
                            $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include/>
-                           $<INSTALL_INTERFACE:include>
+                           $<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}>
```
trxcllnt, line 96: :+1: woops, forgot to update this :sweat_smile:

CMakeLists.txt
```diff
@@ -97,35 +93,55 @@ set_target_properties(p2300 PROPERTIES
...
                        $<$<CXX_COMPILER_ID:MSVC>:
                        /W4>)
-# template backtrace limit
-target_compile_options(p2300 INTERFACE
+
+
+# Silence warnings with GCC
+target_compile_options(p2300_executable_flags INTERFACE
+                       $<$<BOOL:$<CXX_COMPILER_ID:GNU>>:
+                       -Wno-non-template-friend>
```
trxcllnt, line 129:
```cmake
$<$<CXX_COMPILER_ID:GNU>:-Wno-non-template-friend>
```

trxcllnt (approved):
Looks good to me :+1:. Made a few small suggestions on the lines @bustercopley mentioned.

bustercopley (changes requested):
I suggest this extra change as part of the "Set only required compiler flags on main p2300 CMake target, the rest only on tests and examples" commit.
```diff
@@ -211,7 +205,7 @@ catch_discover_tests(test.P2300)
 # Set up examples
 function(def_example target sourceFile)
     add_executable(${target} ${sourceFile})
-    target_link_libraries(${target} PRIVATE P2300::p2300)
+    target_link_libraries(${target} PRIVATE P2300::p2300 p2300_executable_flags)
 endfunction()
 
 def_example(example.hello_world "examples/hello_world.cpp")
```

Thank you, @msimberg and @trxcllnt. It's great to have a nice install target.

msimberg:
Thanks @trxcllnt and @bustercopley for having a look. Very good catches! They should be fixed now.

The flags were intended to be on examples, but I forgot... and the use of `$<BOOL` was just me being ignorant.

ericniebler (approved):

PR #527, opened 2022-07-08
RFC Sequence prototype

kirkshoop:
This branch contains the prototype that I have been working on to represent sequences of async values.

there is one new cpo:
  `_ValueSenderOut set_next(_SequenceReceiver, _ValueSenderIn)`
  
The `set_next` function is called for each value, and passes a `_ValueSenderIn` for that value. The `set_next` function can compose that `_ValueSenderIn` and return a `_ValueSenderOut`. `_ValueSenderOut` in a SequenceAdaptor is often 

  `return set_next(_SequenceReceiver, compose(_ValueSenderIn));`

This onions the `set_next`s in a SequenceSender expression. 

`_ValueSenderOut` is given to the producer, which then connects and starts `_ValueSenderOut`. When `_ValueSenderOut` completes the producer knows that the value has been processed. This can be used for back-pressure and lock-step.

SequenceSender's do not have to be serial. `set_next` can be called simultaneously with different `_ValueSenderIn` to allow bulk/batch/parallel processing of values.

A `fork()` algo would have a fixed number of `_ValueOp` stored in its operation state and would make sure that each was replaced with a new op when the previous op completed. By applying `on(scheduler, _ValueSenderOut)` to each value, or by having a single `on(scheduler, repeat(next_value()))` that uses one schedule to loop through all the values, the scheduler will have control (up to the fixed number of `_ValueOp`) of the concurrency applied to the work.

There are design decisions that need review. 

The sender-per-value design, might be a perf trap. if construct/connect/start is expensive for a composed `_ValueSenderIn`, that cost will be per value. construction overhead could be removed if we allow an l-value connect `_ValueSenderOut` to be connected for each value with the `_ValueSenderIn` cooperating by sending a different value each time it is connected/started.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -234,6 +234,8 @@ set(test_sourceFiles
     test/exec/async_scope/test_spawn_future.cpp
     test/exec/async_scope/test_empty.cpp
     test/exec/async_scope/test_stop.cpp
+    test/exec/test_ignore_all.cpp
+    test/exec/test_iotas.cpp
     )
 
 add_executable(test.stdexec ${test_sourceFiles})
@@ -257,6 +259,7 @@ def_example(clangd.helper "examples/_clangd_helper_file.cpp")
 def_example(example.hello_world "examples/hello_world.cpp")
 def_example(example.hello_coro "examples/hello_coro.cpp")
 def_example(example.scope "examples/scope.cpp")
+def_example(example.sequence "examples/sequence.cpp")
 def_example(example.retry "examples/retry.cpp")
 def_example(example.then "examples/then.cpp")
 def_example(example.server_theme.let_value "examples/server_theme/let_value.cpp")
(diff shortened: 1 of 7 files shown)


kirkshoop:
> would we be better off going further and having a stream apply algorithm that you wrap any other operation in?

apply() could be a useful algo, but there are many algorithms that use state shared across values. apply() is not the right way to represent this and let_each_value() is more useful than apply but still insufficient.

Algos like pairwise(), window(), group_by(), throttle(), debounce(), etc.. all need to be their own bespoke implementations over the sequence.

prlw1:
Low brow by comparison: I just grabbed 4fc4190b8, applied this PR, and tried to compile examples/sequence.cpp. clang 13.0.1 complained, e.g.,
```
 sequence.cc:41:18: error: 'make_env' does not name a template but is followed by template arguments; did you mean 'make_env_t'?
```
I thought clang was the favoured compiler? (and looking forward to a repeat_until_effect to go with it...)

ericniebler:
closing this and awaiting a new and improved stream design from @kirkshoop and @lewissbaker. :-)

PR #532, opened 2022-07-15
fix undefined behavior in the server_theme.on_transfer example

ericniebler:


examples/server_theme/on_transfer.cpp
```diff
@@ -40,14 +40,32 @@ int main() { return 0; }
...
+  std::ostream& sout_;
+  std::unique_lock<std::mutex> lock_{s_mtx_};
+
+  template <class T>
+  friend sync&& operator<<(sync&& self, const T& value) {
+    self.sout_ << value;
+    return std::move(self);
+  }
+};
+std::mutex sync::s_mtx_{};
```
ericniebler, line 67:
```suggestion
struct sync_stream {
private:
  static std::mutex s_mtx_;

public:
  std::ostream& sout_;
  std::unique_lock<std::mutex> lock_{s_mtx_};

  template <class T>
  friend sync_stream&& operator<<(sync_stream&& self, const T& value) {
    self.sout_ << value;
    return std::move(self);
  }
  friend sync_stream&& operator<<(sync_stream&& self, std::ostream& (*manip)(std::ostream&)) {
    self.sout_ << manip;
    return std::move(self);
  }
};
std::mutex sync_stream::s_mtx_{};
```

examples/server_theme/on_transfer.cpp
```diff
@@ -57,7 +75,7 @@ size_t legacy_read_from_socket(int sock, char* buffer, size_t buffer_len) {
 }
 
 void process_read_data(const char* read_data, size_t read_len) {
-  std::cout << "Processing '" << std::string_view{read_data, read_len} << "'\n";
+  sync{std::cout} << "Processing '" << std::string_view{read_data, read_len} << "'\n";
```
ericniebler, line 78:
```suggestion
  sync_stream{std::cout} << "Processing '" << std::string_view{read_data, read_len} << "'\n";
```

PR #536, opened 2022-07-19
Fix the receiver concept to check that all completion signals are supported

### LINKED ISSUE #447: Does has_completions only check the last completion signature?
https://github.com/brycelelbach/wg21_p2300_std_execution/blob/8fe31c65711ae8dc08847305017e4ead25155156/include/execution.hpp#L291-L293

Can anyone explain why has_completions only check the last completion signature? Is that by design?

@ericniebler

ericniebler:
Fixes #447

### DIFF
--- a/include/execution.hpp
+++ b/include/execution.hpp
@@ -298,6 +298,11 @@ namespace std::execution {
     struct _MISSING_COMPLETION_SIGNAL_<_Tag(_Args...)> {
       template <class _Receiver>
         struct _WITH_RECEIVER_ : false_type {};
+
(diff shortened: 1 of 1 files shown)


villevoutilainen (commented):
Hideous. Is it really so heavy to use a conditional_t with a folded variadic same_as that you have to resort to a hack like this?

ericniebler:
This is going to get used _a lot_, and S&R is heavy enough. I'm willing to go quite a bit farther than this in the interest of compile times. If you think this is a hideous hack, avert your eyes.

Issue #540, opened 2022-07-19
propose a `write` sender adaptor


ericniebler:
This is the analogue of the `read` sender factory. `write` would adapt a sender by adding a set of [cpo, value] pairs to an environment.

**Proposed Usage:**

* `sndr | write(with(cpo1, value1), with(cpo2, value2)...)`

See also #538, #539

PR #546, opened 2022-07-20
add a write sender adaptor

ericniebler:
**Usage:**

```c++
auto sndr = write(read(cpo), with(cpo, value));
auto [result] = sync_wait(sndr).value();
assert(result == value);
```

refs #540


gonidelis:
`write` is not in the current version of P2300, right?

ericniebler:
Correct. I need to write a paper to propose a bunch of the environment manipulators.

PR #556, opened 2022-07-30
fix a bug in split with the handling of an external stop request

ericniebler:


include/execution.hpp
```diff
@@ -2569,6 +2569,28 @@ namespace std::execution {
...
                 __on_stop_requested{__shared_state->__stop_source_});
+            if (__shared_state->__stop_source_.stop_requested()) {
+              // If we are starting the first split sender, then __old
+              // is nullptr and the inner sender has not been started yet.
+              // There is no need to start it; put the shared state into
+              // the "stopped" state, notify the receiver, and return.
+              if (__old == nullptr) {
+                // No need to exchange here; the inner sender isn't running
+                // yet, so there is no race.
+                __head.store(__completion_state, memory_order_release);
```
senior-zero, line 2580: Current wording states:
> - (If receiver contract hasn't been satisfied...), it emplace constructs the stop callback optional with the arguments `execution::get_stop_token(get_env(out_r'))` and `stop-callback-fn{stop-src}`, where `stop-src` refers to the stop source of `sh_state`. 
> - Then, it checks to see if `stop-src.stop_requested()` is true. If so, it calls `execution::set_stopped(out_r')`.

My understanding is that only decayed-copied `out_r` (which maps to `__self.__recvr_`) should receive stop signal. I don't think the  wording currently requires split to stop all connected receivers as well. If all the thread does when observes stop request is propagate stop signal, other thread might start inner sender:

```cpp
if (stop_requested()) { // thread 1 observes false, thread 2 observes true
    set_stopped(out_r_prime); // thread 2 propagates `set_error_t` 
}

if (!contract_satisfied()) {
    start(op_state2); // thread 1 starts inner sender and propagates `set_value_t`
}
```

It leads to cases where split can complete with different signals, which contradicts the contract. Should we update the wording to something along the following lines?

> Then, it checks to see if `stop-src.stop_requested()` is true. If so and `op_state2` hasn't been started yet, it notifies the operation states in `sh_state`'s list of operation states that the results are ready and the completion signal is `set_stopped_t`.

In this case, the inner sender won't be started by other threads and one completion signal would be propagated. 

Regarding the reference implementation, we can't leave the line where it is to achieve the behavior above. Different thread might have won `__head.compare_exchange_weak` after current thread observed `__head.load(memory_order_acquire) == nullptr`. This other thread might be in process of propagating a completion signal, say `set_value_t`. Thread that observed `__head.load(memory_order_acquire) == nullptr` is resetting head to completion state. Some threads observe completion state  change and propagate `set_stopped_t` signal and the program behaves as if split completed with different completion signatures. 

Instead, we might have something like:

```cpp
if (__old != __completion_state) {
  // only insert callback here
  // no guarantee that the sender hasn't been started by different thread
  __self.__on_stop_.emplace(
      get_stop_token(get_env(__self.__recvr_)),
      __on_stop_requested{__shared_state->__stop_source_});
}

do {
  if (__old == __completion_state) {
    __self.__notify(&__self);
    return;
  }
  __self.__next_ = static_cast<__operation_base*>(__old);
} while (!__head.compare_exchange_weak(
    __old, static_cast<void *>(&__self),
    memory_order_release,
    memory_order_acquire));

if (__old == nullptr) {
  // inner sender isn't running
  if (__shared_state->__stop_source_.stop_requested()) {
    // 1. resets __head to completion state
    // 2. notifies waiting threads
    // 3. propagates sinal to `out_r'`
    __shared_state->__notify(); 
  } else {
    start(__shared_state->__op_state2_);
  }
}
```

senior-zero (changes requested):

include/execution.hpp
ericniebler: That makes sense to me. Another option would be to replace the `__head.store(...` I added with `__head.compare_exchange(__old, ...`, and only propagate the stop signal if we win the race. I kinda like your solution.

And yes, we should clarify the spec on this point.

ericniebler: > Then, it checks to see if <tt><i>stop-src</i>.stop_requested()</tt> is `true`. If so and `op_state2` hasn't been started yet, it notifies the operation states in `sh_state`'s list of operation states that the results are ready and the completion signal is `set_stopped_t`.

The problem here is that `op_state` hasn't been added to the list yet, so it would miss the notification. I'm noodling on the best way to patch this up.

ericniebler: @senior-zero How about:

> * Otherwise, it adds a pointer to `op_state` to the list of
>   operation states in `sh_state`. If `op_state` is the first such
>   state added to the list:
> 
>      * If <code><i>stop-src</i>.stop_requested()</code> is `true`,
>         all of the operation states in `sh_state`'s list of operation
>         states are notified as if `execution::set_stopped(r)` had
>         been called.
>
>      * Otherwise, `execution::start(op_state2)` is called.

std_execution.bs
```diff
@@ -5488,14 +5488,16 @@ template &lt;class E>  // arguments are not associated entities ([lib.tmpl-heads
                 <code><i>stop-src</i></code> refers to the stop source of
                 `sh_state`.
 
-            * Then, it checks to see if
-                <code><i>stop-src</i>.stop_requested()</code> is `true`. If so, it
-                calls `execution::set_stopped(out_r')`.
-
             * Otherwise, it adds a pointer to `op_state` to the list of
```
senior-zero, line 5491: Otherwise is dangling now. Should it be:
```suggestion
            * Then, it adds a pointer to `op_state` to the list of
```

senior-zero (approved):

std_execution.bs
ericniebler: That's not how the Committee specifies these things. They're very prescriptive about wording. Things are specified like:

* If `X` is `true`, then _<case 1>_.
* Otherwise, if `Y` is `true`, then _<case 2>_.
* Otherwise, _\<fallback case>_.

senior-zero: But that's now the case we are having. Currently, you have:
- If receiver contract has already been satisfied..
- Otherwise, it does X.
- Otherwise, it does Y. 

So, you seems to be describing the case if () else if () else. But it seems that the current wording is if () else () else (), which doesn't make sense to me.

include/execution.hpp
```diff
@@ -2633,7 +2633,15 @@ namespace std::execution {
...
-            start(__shared_state->__op_state2_);
+            // the inner sender isn't running
+            if (__shared_state->__stop_source_.stop_requested()) {
+              // 1. resets __head to completion state
+              // 2. notifies waiting threads
+              // 3. propagates "stopped" signal to `out_r'`
+              __shared_state->__notify();
+            } else {
+              start(__shared_state->__op_state2_);
+            }
```
lewissbaker, line 2644: What bug exactly does this fix?

If there was a race condition with the previous implementation then there would still be a race-condition with the new implementation (e.g. what if stop is requested on another thread immediately after `stop_requested()` returns).

This seems to be purely an optimisation that avoids starting the underlying op if stop has already been requested.

ericniebler: Without this change, the first of the new test cases I added will fail, and `split` won't conform to the spec.

PR #572, opened 2022-08-12
re-express `sender_of` to move optional `Env` param to the back

### LINKED ISSUE #486: `sender_of`'s execution environment parameter doesn't work as written and should be removed
I think the introduction of the notion of an execution environment for senders has inadvertently made the `sender_of` concept far less user friendly:

Before it was:

```
  template<class S, class... Ts>
    concept sender_of = see below;
```

Now it is:

```
  template<class S, class E = no_env, class... Ts>
    concept sender_of = see below;
```

The default doesn't seem correct - as written, the default will always be used, and any argument you pass after the second will be a part of Ts.

If we removed the default, then the interface would be worse:

Before:

```
sender_of<std::size_t> auto async_read(
    sender_of<std::span<std::byte>> auto buffer,
    auto handle);
```

After:

```
sender_of<std::size_t, no_env> auto async_read(
    sender_of<std::span<std::byte>, no_env> auto buffer,
    auto handle);
```

Given that one of the main reasons we added sender_of was because I wanted it for these examples, I'd rather not make them more complicated. 

I suggest we just hard wire `sender_of` to always use `no_env.

ericniebler:
fixes #486

### DIFF
--- a/test/concepts/test_concepts_sender.cpp
+++ b/test/concepts/test_concepts_sender.cpp
@@ -57,6 +57,8 @@ struct my_sender_int {
 TEST_CASE("my_sender_int is a sender & sender", "[concepts][sender]") {
   REQUIRE(ex::sender<my_sender_int>);
   REQUIRE(ex::sender<my_sender_int, empty_env>);
+  REQUIRE(ex::sender_of<my_sender_int, ex::set_value_t(int)>);
+  REQUIRE(ex::sender_of<my_sender_int, ex::set_value_t(int), empty_env>);
 }
 TEST_CASE("sender that accepts an int receiver models sender_to the given receiver",
     "[concepts][sender]") {
@@ -77,11 +79,13 @@ TEST_CASE("can query completion signatures for a typed sender that sends nothing
   check_val_types<type_array<type_array<>>>(my_sender0{});
   check_err_types<type_array<std::exception_ptr>>(my_sender0{});
   check_sends_stopped<true>(my_sender0{});
+  REQUIRE(ex::sender_of<my_sender0, ex::set_value_t()>);
 }
 TEST_CASE("can query completion signatures for a typed sender that sends int", "[concepts][sender]") {
   check_val_types<type_array<type_array<int>>>(my_sender_int{});
   check_err_types<type_array<std::exception_ptr>>(my_sender_int{});
   check_sends_stopped<true>(my_sender_int{});
+  REQUIRE(ex::sender_of<my_sender_int, ex::set_value_t(int)>);
 }
 
 struct multival_sender {
@@ -98,6 +102,7 @@ TEST_CASE("check completion signatures for sender that advertises multiple sets 
   check_val_types<type_array<type_array<int, double>, type_array<short, long>>>(multival_sender{});
   check_err_types<type_array<std::exception_ptr>>(multival_sender{});
   check_sends_stopped<false>(multival_sender{});
+  REQUIRE_FALSE(ex::sender_of<multival_sender, ex::set_value_t(int, double)>);
 }
 
 struct ec_sender {
@@ -113,4 +118,5 @@ TEST_CASE("check completion signatures for sender that also supports error codes
   check_val_types<type_array<type_array<>>>(ec_sender{});
   check_err_types<type_array<std::exception_ptr, int>>(ec_sender{});
   check_sends_stopped<false>(ec_sender{});
+  REQUIRE(ex::sender_of<ec_sender, ex::set_value_t()>);
 }
--- a/test/concepts/test_concepts_sender.cpp
+++ b/test/concepts/test_concepts_sender.cpp
@@ -39,6 +39,13 @@ struct my_sender0 {
 TEST_CASE("type w/ proper types, is a sender & sender", "[concepts][sender]") {
   REQUIRE(ex::sender<my_sender0>);
   REQUIRE(ex::sender<my_sender0, empty_env>);
+
+  REQUIRE(ex::sender_of<my_sender0, ex::set_value_t()>);
+  REQUIRE(ex::sender_of<my_sender0, ex::set_error_t(std::exception_ptr)>);
+  REQUIRE(ex::sender_of<my_sender0, ex::set_stopped_t()>);
+  REQUIRE(ex::sender_of<my_sender0, ex::set_value_t(), empty_env>);
+  REQUIRE(ex::sender_of<my_sender0, ex::set_error_t(std::exception_ptr), empty_env>);
+  REQUIRE(ex::sender_of<my_sender0, ex::set_stopped_t(), empty_env>);
 }
 TEST_CASE(
     "sender that accepts a void sender models sender_to the given sender", "[concepts][sender]") {
--- a/include/execution.hpp
+++ b/include/execution.hpp
@@ -1241,12 +1241,12 @@ namespace _P2300::execution {
         connect((_Sender&&) __sndr, (_Receiver&&) __rcvr);
       };
 
-  template<class _Sender, class _Env = no_env, class... _Ts>
+  template<class _Sender, class _SetValueSig, class _Env = no_env>
     concept sender_of =
       sender<_Sender, _Env> &&
       same_as<
-        __types<__types<_Ts...>>,
-        value_types_of_t<_Sender, _Env, __types, __types>>;
+        __types<_SetValueSig>,
+        __value_types_of_t<_Sender, _Env, __qf<set_value_t>, __q<__types>>>;
 
   /////////////////////////////////////////////////////////////////////////////
   // [exec.snd_queries], sender queries
--- a/include/execution.hpp
+++ b/include/execution.hpp
@@ -1241,12 +1241,22 @@ namespace _P2300::execution {
         connect((_Sender&&) __sndr, (_Receiver&&) __rcvr);
       };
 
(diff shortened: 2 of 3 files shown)


ericniebler:
Needs wording. Comments welcome.

kirkshoop (approved):
I like it!

ericniebler:
@kirkshoop Do you think I should support querying the error or stopped channel with `sender_of`? `sender_of<Sndr, set_error_t(error_code)>`? Or keep it strictly about the value channel?

kirkshoop:
I would say that set_error and set_stopped must be supported. If not here, then in a similar concept with a different name. If we add other concepts for them then sender_of would also get a name change I think. 

The simplest is to keep sender_of and allow set_error and set_stopped signatures. 

What do you think?

griwes:
The funny thing about this reformulation is that it makes me want to express multiple signatures with a single invocation of the concept, but either requires spelling out completion_signatures in such invocations, or lands on the same problem as the original where env can't be defaulted.

Maybe we want `sender_of` for checking a single signature, and then `sender_of_subset` (placeholder name) that takes a full completion signatures type and checks that the sender has no completions other than the ones specified? Currently we can check that by spelling out a receiver, but that's not particularly to algorithms that don't write their own receivers.

The other question is about this new `sender_of`. What exact semantics do we want here? "This is the only signature for the specified channel" seems reasonable, but I'm unsure if it's correct.

kirkshoop:
I want to be able to use multiple sender_of checks for the same channel. I would not find sender_of very useful if it only allowed one signature per signal. 

A sender_of that takes a completion_signatures is nice as a convenience, but I don't see it as necessary.

griwes:
What you want doesn't solve the need that sender_of was created for, i.e. checking that a sender sends exactly what you ask for. If you allow for other signals, then an algorithm also has to handle those other signals; if you require exactly just the provided one, then an algorithm knows it doesn't need to do anything else. I believe that all uses of the concept in P2300 examples are like this.

What you are describing also seems useful, but as an addition, not a replacement. One family of concepts that checks for "at most" the signals specified (that's the current sender_of), one that checks for "at least" (that's what you're asking for).

And the case of checking for multiple signatures of the same signal is why I suggested to have both the single signal and the completion signatures versions. I'm sure there's a way to spell all of this out in a useful manner to cover all these use cases with separate concepts.

ericniebler:
So, I agree with @griwes, and if you read the bug this is addressing (#486), you'll see that the semantics Michal describes is what Bryce is looking for also. It's somewhat cynical, but the purpose of this concept is to make examples look nicer, and that's pretty much it.

griwes:
I mean, it can also be used to neatly constrain algorithms in general. I consider it to be at least a couple of layers above slideware ;)

kirkshoop:
A concept that takes a completion_signatures and requires an exact match would need a different name. For instance sender_of_all_exact. 

A concept that only takes one signature and requires only that signature for that signal would also need a different name. For instance, sender_of_one_exact.

The name sender_of indicates that that it sends at least the signature(s) provided. 

I don't see how to build constraints for only one signal or one signature with an exact form. I would need to know every possible signature to constrain for any signature. 

I must be missing something.

ericniebler:
> I don't see how to build constraints for only one signal or one signature with an exact form. I would need to know every possible signature to constrain for any signature.

I don't understand what it is you're trying to do. Can you give an example?

test/concepts/test_concepts_sender.cpp
```diff
@@ -77,11 +86,13 @@ TEST_CASE("can query completion signatures for a typed sender that sends nothing
   check_val_types<type_array<type_array<>>>(my_sender0{});
   check_err_types<type_array<std::exception_ptr>>(my_sender0{});
   check_sends_stopped<true>(my_sender0{});
+  REQUIRE(ex::sender_of<my_sender0, ex::set_value_t()>);
 }
 TEST_CASE("can query completion signatures for a typed sender that sends int", "[concepts][sender]") {
   check_val_types<type_array<type_array<int>>>(my_sender_int{});
   check_err_types<type_array<std::exception_ptr>>(my_sender_int{});
   check_sends_stopped<true>(my_sender_int{});
+  REQUIRE(ex::sender_of<my_sender_int, ex::set_value_t(int)>);
```
williamspatrick, line 95: Is it worthwhile to include test cases of direct types?

```
REQUIRE(ex::sender_of<my_sender0>, int)
```

In most cases where I've wanted to use this has been something like:

`template <sender_of<int> S> auto foo(S s)`

Having the extra `set_value_t` doesn't seem very pleasing or documented.

ericniebler: You're right, it's not documented, which is why I haven't merged this yet. And `sender_of<int>` was how it used to work, but when P2300 got support for dependent sender types, it was no longer sufficient. Some types are only senders when used together with certain environment types, so `sender_of` needs to take an optional environment. But taking the environment last (which you would need to do if you want it to be optional) interferes with senders that send more than one value, like `sender_of< int, std::string >`. Hence, `sender_of<set_value_t(int, std::string) [, optional-env]>`.

We _could_ accept `sender_of<T [, optional-env]>` where `T` is not a function type, and interpret it as `sender_of<set_value_t(T) [, optional-env]>`. That feels a little dirty to me, but I understand its appeal. Do others have an opinion about this? @kirkshoop @griwes ?

kirkshoop: I would much rather have a different named concept for the single value type case. 

`sends_value<T[, Env]>`

williamspatrick: Thanks @ericniebler for the info.

I must admit I haven't fully absorbed the environment stuff yet and I've probably only absorbed about 40% of the rest, so take my opinion for whatever it's worth.  I'm mostly working in the coroutine model and it seems pretty handy to have some kind of concept to indicate "this type acts as if it did a `co_return T`".  That's what `sender_of<T>` did in libunifex and what seems like `sender_of<noenv, T>` does now here.  I don't have any opinion on the naming, but as a concept it seems fairly useful and a lot more consumable than with the extra `set_value_t`.  Again though, I haven't fully absorbed what it means for a sender to have multiple `set_value_t`s.

I actually arrived at this PR because I was trying to figure out why `sender_of<bool>` didn't currently work like I expected from libunifex and the P2300 examples, which lead to finding out about the environment parameter and this corresponding issue/PR.

ericniebler:
So, `receiver_of` takes a `completion_signatures<Signatures...>` type, and it checks that for each signature in `Signatures...`, the receiver can accept that completion. It is not exclusive; that is, it doesn't check that the receiver accepts those completions _and only those completions_.

That suggests to me that @kirkshoop is right. Calling this `sender_of` and giving it a radically different interface with radically different semantics now seems like a bad idea to me.

griwes:
I have a counterpoint to that. We design concepts to match algorithms. For receiver_of, the semantics of "handles at least these signatures" is algorithmically useful: you know a set of signatures and you want to check if a given receiver can handle them all.

The semantics of "this sends at most these signatures" for sender_of (and mind you, that doesn't mean it necessarily sends all of them) serve the same purpose as those of receiver_of: here's a list of signatures my algorithm handles, and I want to make sure a sender I get doesn't send anything else, which I _don't_ handle.

I don't understand the algorithmic purpose of "this sender sends _at least_ these signatures". I'm not saying that doesn't exist, just that I don't understand it, but it seems to me that it's fairly detached from receiver_of's algorithmic purpose.

For receivers, being able to handle more signals than specified doesn't matter for code that consumes them, as it will only pass the specified signatures in anyway. For senders, being able to send _less_ signals is the analogous thing that doesn't cause code to not compile; being able to send _more_ signals will, however, make code consuming a sender not handle all of them and, most probably, fail to compile.

Furthermore: if you have a list of signatures, and have an algorithm that accepts a sender_of those signatures (in the "it sends at most those signatures" sense) and a receiver_of those signatures (in the "it receives at least those signatures" sense), you are guaranteed that you can connect them together. If we, however, make the semantics of sender_of be "sends at least these signatures" instead, it will no longer be guaranteed that you can connect them together. This seems very, very wrong to me.

tl;dr: I don't consider "at least these signatures" for receiver_of and "at most these signatures" for sender_of to be "radically different semantics"; they compliment each other in a very natural way.

villevoutilainen:
> I don't understand the algorithmic purpose of "this sender sends _at least_ these signatures". I'm not saying that doesn't exist, just that I don't understand it, but it seems to me that it's fairly detached from receiver_of's algorithmic purpose.

Something like selective interceptor/filter algorithms come to mind. "I will intercept a std::expected on the value channel
and transfer it onto the error channel if the expected has an error in it", or "I will intercept a std::optional on the value channel
and transfer it onto the error channel if it's disengaged", perhaps? Such algorithms could well be transparent for other types, i.e. just let them through, so they would accept senders that send at least the signatures they want to intercept/filter, and wouldn't need to constrain the other signatures.

griwes:
Yeah that makes sense. But a constraint for that doesn't feel like the other side of the coin of receiver_of, so to speak.

villevoutilainen:
..but other than that, your rumination makes perfect sense to me, especially the "guaranteed to be able to connect" part.

ericniebler:
:thinking: OK, I can buy that argument.

ericniebler:
I've updated the wording to reflect the new design of `sender_of`. Reviews are welcome.

ericniebler:
I'm just merging this as-is. I have too many things in flight, and I need to land some of them or air traffic control will go on strike.

PR #576, opened 2022-08-21
Use int instead of bool to detect missing member functions

bustercopley:
This works around a bug in mingw gcc, in which 'bool(type::f)' evaluates
to true if f is the name of a member function declared in type.

With this change all tests pass on gcc 11.3 and 12.2 in MSYS2.

Without this change, the _MISSING_MEMBER overloads are always selected, and fail to compile because there is no base to fall back to.


ericniebler:
Thanks!

PR #577, opened 2022-08-23
Add standard bulk

senior-zero:
This PR adds reference `bulk` implementation and fixes a minor bug in the wording.

### DIFF
--- a/include/execution.hpp
+++ b/include/execution.hpp
@@ -2461,7 +2461,109 @@ namespace _P2300::execution {
   /////////////////////////////////////////////////////////////////////////////
   // [execution.senders.adaptors.bulk]
   namespace __bulk {
+    template <class _ReceiverId, integral _Shape, class _FunId>
+      class __receiver
+        : receiver_adaptor<__receiver<_ReceiverId, _Shape, _FunId>, __t<_ReceiverId>> {
+        using _Receiver = __t<_ReceiverId>;
+        using _Fun = __t<_FunId>;
+        friend receiver_adaptor<__receiver, _Receiver>;
+
+        [[no_unique_address]] _Shape __shape_;
+        [[no_unique_address]] _Fun __f_;
+
+        template <class... _As>
+        void set_value(_As&&... __as) && noexcept 
+          requires __nothrow_callable<_Fun, _Shape, _As...> {
+          for (_Shape __i{}; __i != __shape_; ++__i) {
+            __f_(__i, __as...);
+          }
+          execution::set_value(std::move(this->base()), (_As&&)__as...);
+        }
+
+        template <class... _As>
+        void set_value(_As&&... __as) && noexcept 
+          requires __callable<_Fun, _Shape, _As...> {
+          try {
+            for (_Shape __i{}; __i != __shape_; ++__i) {
+              __f_(__i, __as...);
+            }
+            execution::set_value(std::move(this->base()), (_As&&)__as...);
+          } catch(...) {
+            execution::set_error(std::move(this->base()), std::current_exception());
+          }
+        }
+
+       public:
+        explicit __receiver(_Receiver __rcvr, _Shape __shape, _Fun __fun)
+          : receiver_adaptor<__receiver, _Receiver>((_Receiver&&) __rcvr)
+          , __shape_(__shape)
+          , __f_((_Fun&&) __fun)
+        {}
+      };
+
+    template <class _SenderId, integral _Shape, class _FunId>
+      struct __sender {
+        using _Sender = __t<_SenderId>;
+        using _Fun = __t<_FunId>;
+        template <receiver _Receiver>
+          using __receiver = __receiver<__x<remove_cvref_t<_Receiver>>, _Shape, _FunId>;
+
+        [[no_unique_address]] _Sender __sndr_;
+        [[no_unique_address]] _Shape __shape_;
+        [[no_unique_address]] _Fun __fun_;
+
+        template <class _Tag, class _Fun, class _Sender, class _Env>
+          using __with_error_invoke_t =
+            __if_c<
+              __v<__value_types_of_t<
+                _Sender,
+                _Env,
+                __mbind_front_q<__non_throwing_, _Fun, _Shape>,
+                __q<__mand>>>,
+              completion_signatures<>,
+              __with_exception_ptr>;
+
+        template <class _Self, class _Env>
+          using __completion_signatures =
+            __make_completion_signatures<
+              __member_t<_Self, _Sender>,
+              _Env,
+              __with_error_invoke_t<set_value_t, _Fun, __member_t<_Self, _Sender>, _Env>>;
+
+        template <__decays_to<__sender> _Self, receiver _Receiver>
+          requires sender_to<__member_t<_Self, _Sender>, __receiver<_Receiver>>
+        friend auto tag_invoke(connect_t, _Self&& __self, _Receiver&& __rcvr)
+          noexcept(__nothrow_connectable<__member_t<_Self, _Sender>, __receiver<_Receiver>>)
+          -> connect_result_t<__member_t<_Self, _Sender>, __receiver<_Receiver>> {
+          return execution::connect(
+              ((_Self&&) __self).__sndr_,
+              __receiver<_Receiver>{
+                (_Receiver&&) __rcvr, 
+                __self.__shape_, 
+                ((_Self&&) __self).__fun_});
+        }
+
+        template <__decays_to<__sender> _Self, class _Env>
+        friend auto tag_invoke(get_completion_signatures_t, _Self&&, _Env)
+          -> dependent_completion_signatures<_Env>;
+
+        template <__decays_to<__sender> _Self, class _Env>
+        friend auto tag_invoke(get_completion_signatures_t, _Self&&, _Env)
+          -> __completion_signatures<_Self, _Env> requires true;
+
+        template <__sender_queries::__sender_query _Tag, class... _As>
+          requires __callable<_Tag, const _Sender&, _As...>
+        friend auto tag_invoke(_Tag __tag, const __sender& __self, _As&&... __as)
+          noexcept(__nothrow_callable<_Tag, const _Sender&, _As...>)
+          -> __call_result_if_t<__sender_queries::__sender_query<_Tag>, _Tag, const _Sender&, _As...> {
+          return ((_Tag&&) __tag)(__self.__sndr_, (_As&&) __as...);
+        }
+      };
+
     struct bulk_t {
+      template <sender _Sender, integral _Shape, class _Fun>
+        using __sender = __sender<__x<remove_cvref_t<_Sender>>, _Shape, __x<remove_cvref_t<_Fun>>>;
+
       template <sender _Sender, integral _Shape, __movable_value _Fun>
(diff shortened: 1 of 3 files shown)


include/execution.hpp
```diff
@@ -2461,7 +2461,110 @@ namespace _P2300::execution {
...
+        friend receiver_adaptor<__receiver, _Receiver>;
+
+        [[no_unique_address]] _Shape __shape_;
+        [[no_unique_address]] _Fun __f_;
+
+        template <class... _As>
+        void set_value(_As&&... __as) && noexcept 
+          requires __nothrow_invocable<_Fun, _Shape, _As...> {
+          for (_Shape __i{}; __i != __shape_; ++__i) {
+            std::invoke(__f_, __i, std::forward<_As>(__as)...);
```
ericniebler, line 2478: Since `_Shape` is currently required to model `integral`, `__f_` cannot be a member pointer, and `std::invoke` can be replaced with an ordinary function call. I avoid `std::invoke` when I know it's not needed for improved compile times and diagnostics. (Can also replace the constraint with `__nothrow_callable`.)

Also, it looks suspicious to me that we're `std::forward`-ing the values into the callable. We don't want to let the callable consume the value. That would be bugged.

include/execution.hpp
```diff
@@ -2461,7 +2461,110 @@ namespace _P2300::execution {
...
+          }
+          execution::set_value(std::move(this->base()), std::forward<_As&&>(__as)...);
+        }
+
+        template <class... _As>
+        void set_value(_As&&... __as) && noexcept {
```
ericniebler, line 2484: Needs a constraint.

include/execution.hpp
```diff
@@ -2461,7 +2461,110 @@ namespace _P2300::execution {
...
+
+        [[no_unique_address]] _Sender __sndr_;
+        [[no_unique_address]] _Shape __shape_;
+        [[no_unique_address]] _Fun __fun_;
+
+        template <class _Tag, class _Fun, class _Sender, class _Env>
+          using __with_error_invoke_t =
+            __if_c<
+              __v<__gather_sigs_t<
+                _Tag,
```
ericniebler, line 2519: `_Tag` can only be `set_value_t` here, so `__gather_sigs_t<_Tag,...` can be replaced with `__value_types_of_t<...`.

include/execution.hpp
```diff
@@ -2461,7 +2461,110 @@ namespace _P2300::execution {
...
+        [[no_unique_address]] _Shape __shape_;
+        [[no_unique_address]] _Fun __f_;
+
+        template <class... _As>
+        void set_value(_As&&... __as) && noexcept 
+          requires __nothrow_invocable<_Fun, _Shape, _As...> {
+          for (_Shape __i{}; __i != __shape_; ++__i) {
+            std::invoke(__f_, __i, std::forward<_As>(__as)...);
+          }
+          execution::set_value(std::move(this->base()), std::forward<_As&&>(__as)...);
```
ericniebler, line 2480: `std::forward`, blech. Just a waste of compile-time resources. `static_cast<_As&&>(__as)...` is fine. `(_As&&) __as...` would also work here and be consistent with the rest of the code. (Yes, I use C-style casts for forwarding. :man_shrugging:)

include/execution.hpp
```diff
@@ -2461,7 +2461,110 @@ namespace _P2300::execution {
...
+        template <class... _As>
+        void set_value(_As&&... __as) && noexcept {
+          try {
+            for (_Shape __i{}; __i != __shape_; ++__i) {
+              std::invoke(__f_, __i, std::forward<_As>(__as)...);
+            }
+            execution::set_value(std::move(this->base()), std::forward<_As&&>(__as)...);
+          } catch(...) {
+            set_error(std::move(this->base()), std::current_exception());
+            execution::set_error(std::move(this->base()), std::current_exception());
```
ericniebler, line 2492: This looks extremely suspicious to me. Why are we calling `set_error` twice?

ericniebler (changes requested):
Looks good overall, needs a few minor tweaks, I think.

include/execution.hpp
senior-zero: Thank you for noticing this! Happened to be a leftover.

prlw1:
Very happy to see this given

> All agents created by a single use of bulk execute with the same guarantee. This allows, for instance, a scheduler to execute all invocations of the function in parallel.

in the specification.

I had a quick try with
```
example::static_thread_pool ctx{8};
ex::scheduler auto sch = ctx.get_scheduler();
ex::sender auto go = ex::schedule(sch)
    | ex::bulk(4, [](int i){ fmt::print("hello {}\n", i); sleep_for(2s);} )
    ;
sync_wait(std::move(go));
```
but it took 8s, not the hoped for 2s.

According to `static_thread_pool.hpp`, I see `return std::execution::forward_progress_guarantee::parallel;`
which looks hopeful, and if I create a few senders scheduled on that `sch` and send them to `when_all`, I
do get parallel execution.

What am I missing?

senior-zero:
@prlw1 thank you for your comment! I believe there's an issue with the wording. Particularly, the user side part of the bulk description is out of sync with the [exec] one. 

> the number of agents to be created is known and provided to bulk via its shape parameter.

I don't think that the default implementation can create agents at all. Probably, it should complete with an error if the `n > 1`. This completion is complected as well. Current wording requires `bulk` to not add `exception_ptr` if the function is not throwing when called with predecessor sender value types. So for non-throwing cases, it's unclear which type to use to communicate the issue. The [exec] part simply states that:

> calls f(i, args...) for each i of type Shape from 0 to shape, then calls execution::set_value(out_r, args...). 

Since this PR doesn't provide customization for `static_thread_pool`, `shape` is covered by a single agent. 

Since current implementation is conforming to the description, I don't think this fact is blocking current PR. I'd like to address this separately.

include/execution.hpp
```diff
@@ -2461,7 +2461,109 @@ namespace _P2300::execution {
...
+        using _Receiver = __t<_ReceiverId>;
+        using _Fun = __t<_FunId>;
+        friend receiver_adaptor<__receiver, _Receiver>;
+
+        [[no_unique_address]] _Shape __shape_;
+        [[no_unique_address]] _Fun __f_;
+
+        template <class... _As>
+        void set_value(_As&&... __as) && noexcept 
+          requires __nothrow_callable<_Fun, _Shape, _As...> {
```
ericniebler, line 2476:
```suggestion
          requires __nothrow_callable<_Fun, _Shape, _As&...> {
```

include/execution.hpp
```diff
@@ -2461,7 +2461,109 @@ namespace _P2300::execution {
...
+          for (_Shape __i{}; __i != __shape_; ++__i) {
+            __f_(__i, __as...);
+          }
+          execution::set_value(std::move(this->base()), (_As&&)__as...);
+        }
+
+        template <class... _As>
+        void set_value(_As&&... __as) && noexcept 
+          requires __callable<_Fun, _Shape, _As...> {
```
ericniebler, line 2485:
```suggestion
          requires __callable<_Fun, _Shape, _As&...> {
```

include/execution.hpp
```diff
@@ -2461,7 +2461,109 @@ namespace _P2300::execution {
...
+        using _Sender = __t<_SenderId>;
+        using _Fun = __t<_FunId>;
+        template <receiver _Receiver>
+          using __receiver = __receiver<__x<remove_cvref_t<_Receiver>>, _Shape, _FunId>;
+
+        [[no_unique_address]] _Sender __sndr_;
+        [[no_unique_address]] _Shape __shape_;
+        [[no_unique_address]] _Fun __fun_;
+
+        template <class _Tag, class _Fun, class _Sender, class _Env>
```
ericniebler, line 2515:
```suggestion
        template <class _Fun, class _Sender, class _Env>
```

include/execution.hpp
```diff
@@ -2461,7 +2461,109 @@ namespace _P2300::execution {
...
+                __q<__mand>>>,
+              completion_signatures<>,
+              __with_exception_ptr>;
+
+        template <class _Self, class _Env>
+          using __completion_signatures =
+            __make_completion_signatures<
+              __member_t<_Self, _Sender>,
+              _Env,
+              __with_error_invoke_t<set_value_t, _Fun, __member_t<_Self, _Sender>, _Env>>;
```
ericniebler, line 2531:
```suggestion
              __with_error_invoke_t<_Fun, __member_t<_Self, _Sender>, _Env>>;
```

include/execution.hpp
```diff
@@ -2476,6 +2578,14 @@ namespace _P2300::execution {
         noexcept(nothrow_tag_invocable<bulk_t, _Sender, _Shape, _Fun>) {
         return tag_invoke(bulk_t{}, (_Sender&&) __sndr, (_Shape&&) __shape, (_Fun&&) __fun);
       }
+      template <sender _Sender, integral _Shape, __movable_value _Fun>
+        requires
+           (!__tag_invocable_with_completion_scheduler<bulk_t, set_value_t, _Sender, _Shape, _Fun>) &&
+           (!tag_invocable<bulk_t, _Sender, _Shape, _Fun>)
+      __sender<_Sender, _Shape, _Fun> operator()(_Sender&& __sndr, _Shape __shape, _Fun __fun) const {
+        return __sender<_Sender, _Shape, _Fun>{
+          (_Sender&&)__sndr, __shape, __fun};
```
ericniebler, line 2587:
```suggestion
        return __sender<_Sender, _Shape, _Fun>{
          (_Sender&&) __sndr, __shape, (_Fun&&) __fun};
```

ericniebler (approved):
I have a few small tweaks only; otherwise, this is good to go. And let's fix the `bulk` spec and implementation ASAP so `static_thread_pool` can actually _use_ the concurrency it has.

include/execution.hpp
```diff
@@ -2461,7 +2461,109 @@ namespace _P2300::execution {
...
+
+        template <__decays_to<__sender> _Self, class _Env>
+        friend auto tag_invoke(get_completion_signatures_t, _Self&&, _Env)
+          -> dependent_completion_signatures<_Env>;
+
+        template <__decays_to<__sender> _Self, class _Env>
+        friend auto tag_invoke(get_completion_signatures_t, _Self&&, _Env)
+          -> __completion_signatures<_Self, _Env> requires true;
+
+        template <__sender_queries::__sender_query _Tag, class... _As>
```
williamspatrick, line 2554: Seems like something might be missing here?  I don't see __sender_query having ever been defined.

```
$ git grep __sender_query
execution.hpp:    template <__sender_queries::__sender_query _Tag, class... _As>
execution.hpp:            -> __call_result_if_t<__sender_queries::__sender_query<_Tag>, _Tag,
```

On GCC I'm seeing:
```
../include/sdbusplus/async/std_execution/execution.hpp:3001:15: error: ‘_P2300::execution::__sender_queries::__sender_query’ has not been declared
 3001 |     template <__sender_queries::__sender_query _Tag, class... _As>
```

senior-zero: @williamspatrick I believe it was removed in the following [PR](https://github.com/brycelelbach/wg21_p2300_std_execution/pull/581). It should be `tag_category<forwarding_sender_query>` now. I fixed that locally and will create a PR soon.

prlw1:
Re: my comment above, since https://github.com/brycelelbach/wg21_p2300_std_execution/commit/15207389006529448a00e9eaaafdd49a23e5bfc8, the example runs in 2s not 8s. Thanks!

PR #578, opened 2022-08-23
use std::size_t over size_t

williamspatrick:
Since the implementation is defined in the _P2300 namespace, it does
not have direct access to `std::size_t` through its own namespace.
Depending on the include order you can end up with compile errors.

```cpp
 #include "include/__utility.hpp"

 int main() { return 0; }
```

```
In file included from foo.cpp:1:
include/__utility.hpp:52:13: error: ‘size_t’ has not been declared
   52 |   template <size_t _N>
      |             ^~~~~~
include/__utility.hpp:53:44: error: ‘size_t’ was not declared in this scope; did you mean ‘std::size_t’?
   53 |     using __index = std::integral_constant<size_t, _N>;
      |                                            ^~~~~~
      |                                            std::size_t
...
```

Rather than trying to use the global namespace typedef, explicitly
select `std::size_t` whenever used.

### DIFF
--- a/include/__utility.hpp
+++ b/include/__utility.hpp
@@ -49,8 +49,8 @@ namespace _P2300 {
   template <bool _B>
     using __bool = std::bool_constant<_B>;
 
-  template <size_t _N>
-    using __index = std::integral_constant<size_t, _N>;
+  template <std::size_t _N>
+    using __index = std::integral_constant<std::size_t, _N>;
 
   // Some utilities for manipulating lists of types at compile time
   template <class...>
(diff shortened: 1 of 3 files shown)


williamspatrick:
I attached a sample "recreate" here that directly includes `__utility.hpp` which I realized is probably never intended.  Ultimately, I was trying to import this code into another project, while we wait for this to show up in the standard, and some automated clang-format stuff in that project caused include order to be readjusted, which turned into compile failures.  The direct include of `__utility.hpp` shows the kinds of failures I was seeing as a result of this, but I've switched it through the whole codebase.

ericniebler:
Thank you very much!

PR #580, opened 2022-08-24
Conan: Fix test package

### LINKED ISSUE #579: Conan test_package broken after move of sync_wait to _P2300 namespace
Because the conan package was not updated, building the package now breaks.

Possible solution:
Use `_P2300` instead of `std` in `test_package/test.cpp`

paulbendixen:
Fix after changing sync_wait from `std` to `_P2300`

Closes #579

### DIFF
--- a/include/execution.hpp
+++ b/include/execution.hpp
@@ -4785,6 +4785,9 @@ namespace std {
   namespace execution {
     using namespace _P2300::execution;
   }
+  namespace this_thread {
+    using namespace _P2300::this_thread;
+  }
 }
 
 _PRAGMA_POP()
--- a/test_package/test.cpp
+++ b/test_package/test.cpp
@@ -4,6 +4,6 @@ int main()
 {
 	auto x = std::execution::just(42);
 
-	auto [a] = std::this_thread::sync_wait(std::move(x)).value();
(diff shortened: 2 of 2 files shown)


ericniebler:
A better fix would be to edit this block at the bottom of execution.hpp: https://github.com/brycelelbach/wg21_p2300_std_execution/blob/main/include/execution.hpp#L4783-L4788

... to add:

```c++
  namespace this_thread {
    using namespace _P2300::this_thread;
  }
```

paulbendixen:
@ericniebler That does look like a better solution.
Wouldn't that undo the change that was done in #574 ?

ericniebler:
It doesn't _completely_ undo the change in #574. Consider the following code:

```c++
namespace std {
  struct __detail {};
  __detail f() { return {}; }
}
namespace _P2300 {
  struct __detail {};
  __detail g() { return {}; }
}

namespace std {
  using namespace _P2300;
}

int main() {
  std::f();
  std::g();
}
```

Within the namespace `_P2300` above, I use the name `__detail` for some implementation detail, and that just happens to collide with an entity of the same name in `std`. But it's ok! The code above compiles just fine.

Had I instead defined everything in `std`, it would have been a compile error.

This is the problem that #574 is addressing.

ericniebler:
Thank you. :pray:

PR #587, opened 2022-08-29
Fix bulk

senior-zero:
This PR provides `bulk` customization for `static_thread_pool`. It also harmonizes the "[Design - user side]" part of the proposal with the default implementation described in "Execution control library [exec]". This is done by allowing bulk to assign available execution resources to invocations of the `function` in an unspecified way. The proposal used to require `bulk` to launch `shape` execution agents, which is troublesome at the moment.

### DIFF
--- a/examples/schedulers/static_thread_pool.hpp
+++ b/examples/schedulers/static_thread_pool.hpp
@@ -18,6 +18,7 @@
 
 #include <execution.hpp>
 #include "../detail/intrusive_queue.hpp"
+#include "__utility.hpp"
 
 #include <atomic>
 #include <condition_variable>
@@ -30,7 +31,7 @@
 namespace example {
   struct task_base {
     task_base* next;
-    void (*__execute)(task_base*) noexcept;
+    void (*__execute)(task_base*, std::uint32_t tid) noexcept;
   };
 
   template <typename ReceiverID>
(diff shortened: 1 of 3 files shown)


examples/schedulers/static_thread_pool.hpp
```diff
@@ -88,11 +88,229 @@ namespace example {
...
+          friend void tag_invoke(std::execution::set_value_t, bulk_receiver&& self, As&&... as) noexcept {
+            shared_state& state = self.shared_state_;
+
+            if (state.shape_) {
+              try {
+                using tuple_t = std::execution::__decayed_tuple<As...>;
+                state.data_.template emplace<tuple_t>((As &&) as...);
+                state.pool_.bulk_enqueue(&state, state.num_agents_required());
+              } catch (...) {
+                std::execution::set_error((Receiver&&)state.receiver_, std::current_exception());
```
ericniebler, line 202: Likewise here, we can't instantiate the error path if we know that emplacing the values into the tuple cannot possibly throw, and that also needs to be factored into the `completion_signatures` type computation (which I see is _not_ being done below).

examples/schedulers/static_thread_pool.hpp
```diff
@@ -88,11 +88,229 @@ namespace example {
...
+                Env,
+                with_error_invoke_t<Fun, std::__member_t<Self, Sender>, Env>>;
+
+          template <class Receiver>
+            using bulk_op_state_t = bulk_op_state<SenderId, std::__x<std::remove_cvref_t<Receiver>>, Shape, Fun>;
+
+          template <std::__decays_to<bulk_sender> Self, std::execution::receiver Receiver>
+          friend bulk_op_state_t<Receiver> tag_invoke(std::execution::connect_t, Self&& self, Receiver&& rcvr)
+            noexcept(std::is_nothrow_constructible_v<bulk_op_state_t<Receiver>, static_thread_pool&, Shape, Fun, Sender, Receiver>) {
+            return bulk_op_state_t<Receiver>{
```
ericniebler, line 278: This isn't quite right. First, it's not constrained. Second, although you take the `bulk_sender` by forwarding reference, you aren't forwarding the `sndr_` member into the operation state, and the operation state isn't taking the cv- and ref-qualifiers of the sender into consideration.

ericniebler (changes requested):

examples/schedulers/static_thread_pool.hpp
```diff
@@ -88,11 +88,229 @@ namespace example {
...
+
+              if (is_last_thread) {
+                if (self.exception_) {
+                  if (std::execution::tag_invocable<std::execution::set_error_t, Receiver&&, std::exception_ptr>) {
+                    std::execution::set_error((Receiver&&)self.receiver_, self.exception_);
+                  }
+                } else {
+                  std::visit([&](auto& tupl) noexcept -> void {
+                    std::apply([&](auto&... args) noexcept -> void {
+                        std::execution::set_value((Receiver&&)self.receiver_, (decltype(args)&&)args...);
```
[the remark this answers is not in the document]
senior-zero: Yeah.. I should've used `std::move`. Thank you for noticing this!

examples/schedulers/static_thread_pool.hpp
```diff
@@ -255,6 +478,12 @@ namespace example {
     threadStates_[startIndex].push(task);
   }
 
+  inline void static_thread_pool::bulk_enqueue(task_base* task, std::uint32_t n_threads) noexcept {
+    for (std::size_t i = 0; i < n_threads; ++i) {
+      threadStates_[i % available_parallelism()].push(task);
```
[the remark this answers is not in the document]
senior-zero: `vector::push_back` is potentially throwing indeed, but there are no `vector` usages in this function.

examples/schedulers/static_thread_pool.hpp
```diff
@@ -88,11 +89,275 @@ namespace example {
...
+          static std::pair<Shape, Shape> 
+          even_share(Shape n, std::uint32_t rank, std::uint32_t size) noexcept {
+            const auto avg_per_thread = n / size;
+            const auto n_big_share = avg_per_thread + 1;
+            const auto big_shares = n % size;
+            const auto is_big_share = rank < big_shares;
+            const auto begin = is_big_share ? n_big_share * rank
+                                            : n_big_share * big_shares +
+                                                (rank - big_shares) * avg_per_thread;
+            const auto end = begin + (is_big_share ? n_big_share : avg_per_thread);
```
ericniebler, line 134: This could really use a comment to say what is happening here.

examples/schedulers/static_thread_pool.hpp
```diff
@@ -88,11 +89,275 @@ namespace example {
...
+              if constexpr (MayThrow) {
+                try {
+                  std::visit([&](auto& tupl) -> void {
+                    std::apply([&](auto&... args) -> void {
+                      auto [begin, end] = even_share(self.shape_, tid, total_threads);
+                      for (Shape i = begin; i < end; ++i) {
+                        self.fn_(i, args...);
+                      }
+                    }, tupl);
+                  }, self.data_);
```
ericniebler, line 165: If you have a member function that accepts the inner lambda here as an argument, then you can reuse it 4 times. (The lambda can always take by lvalue reference, and the `set_value` ones can `std::move` the tuple elements.)

Then, 2 of the 4 invocations would be passing the same lambda. Create two named lambdas and reuse them in the 4 invocations of the member function you add. This function gets a lot smaller and easier to maintain.

examples/schedulers/static_thread_pool.hpp
```diff
@@ -88,11 +89,275 @@ namespace example {
...
+                  std::__q<std::__mand>>>,
+                std::execution::completion_signatures<>,
+                std::execution::__with_exception_ptr>;
+
+          template <class Self, class Env>
+            using completion_signatures =
+              std::execution::__make_completion_signatures<
+                std::__member_t<Self, Sender>,
+                Env,
+                with_error_invoke_t<Fun, std::__member_t<Self, Sender>, Env>>;
```
ericniebler, line 313: You'll also need to transform the values with `decay_t<Value>`.

examples/schedulers/static_thread_pool.hpp
```diff
@@ -88,11 +89,275 @@ namespace example {
...
+                } catch (...) {
+                  std::execution::set_error((Receiver&&)state.receiver_, std::current_exception());
+                }
+              } else {
+                using tuple_t = std::execution::__decayed_tuple<As...>;
+                state.data_.template emplace<tuple_t>((As &&) as...);
+                state.pool_.bulk_enqueue(&state, state.num_agents_required());
+              }
+            } else {
+              std::execution::set_value((Receiver&&)state.receiver_, (As&&)as...);
```
ericniebler, line 242: This is slightly off. In the `bulk_enqueue` case, what you'll be sending are xvalues referring to decay-copied values. Here, `As...` could contain lvalues. Once you fix up the completion_signatures as I suggest below, then you will not be declaring that you `bulk` might send lvalue. Instead, I think even in this code path, you need to decay-copy the values into the tuple and then move the elements into the `set_value` call.

Kinda sucks, I know, but them's the breaks.

examples/schedulers/static_thread_pool.hpp
```diff
@@ -88,11 +89,275 @@ namespace example {
...
+                  std::__mbind_front_q<bulk_non_throwing, Fun, Shape>,
+                  std::__q<std::__mand>>>;
+
+          using bulk_receiver = bulk_receiver<SenderId, ReceiverId, Shape, Fun, may_throw>;
+          using shared_state = bulk_shared_state<SenderId, ReceiverId, Shape, Fun, may_throw>;
+          using inner_op_state = std::execution::connect_result_t<Sender, bulk_receiver>;
+
+          shared_state shared_state_;
+
+          inner_op_state inner_op;
```
ericniebler, line 275:
```suggestion
          inner_op_state inner_op_;
```

examples/schedulers/static_thread_pool.hpp
```diff
@@ -88,11 +89,275 @@ namespace example {
...
+                with_error_invoke_t<Fun, std::__member_t<Self, Sender>, Env>>;
+
+          template <class Self, class Receiver>
+            using bulk_op_state_t = 
+              bulk_op_state<
+                std::__x<std::__member_t<Self, Sender>>, 
+                std::__x<std::remove_cvref_t<Receiver>>, Shape, Fun>;
+
+          template <std::__decays_to<bulk_sender> Self, std::execution::receiver Receiver>
+            requires std::execution::sender_to<std::__member_t<Self, Sender>, Receiver>
```
ericniebler, line 322: This constraint also doesn't factor in the fact that the `bulk_sender` decay-copies values. Maybe this instead (once you make the above change to `completion_signatures`):

```suggestion
            requires std::execution::receiver_of<Receiver, completion_signatures<Self, env_of_t<Receiver>>
```

ericniebler (changes requested):
Looking good, needs just a few more tweaks.

examples/schedulers/static_thread_pool.hpp
```diff
@@ -88,11 +89,292 @@ namespace example {
...
+            }
+
+            if (state.shape_) {
+              state.pool_.bulk_enqueue(&state, state.num_agents_required());
+            } else {
+              std::visit([&](auto& tupl) -> void {
+                std::apply([&](auto&... args) -> void {
+                  std::execution::set_value((Receiver&&)state.receiver_, std::move(args)...);
+                }, tupl);
+              }, state.data_);
```
ericniebler, line 253: This could also use `bulk_shared_state::apply`, right?

ericniebler (approved):
Just that last question about using `shared_state::apply` from the receiver, then I'll merge.

examples/schedulers/static_thread_pool.hpp
senior-zero: Good point, done!

ericniebler:
Thanks!

PR #592, opened 2022-08-30
Restore conan packaging ability

### LINKED ISSUE #591: Conan packages can no longer be built
After moving the reterival of Catch from conan to CPM, the ability to build the library as a package was removed.

The conanfile.py and files in test_package should be restored, this should in no way hinder the usage of CPM

(possibly the generator should be changed to be cmake_find_package)

paulbendixen:
Restore the parts that enable building a conan package, without requiring the use of conan to retrieve Catch

Closes #591

### DIFF
--- a/test_package/CMakeLists.txt
+++ b/test_package/CMakeLists.txt
@@ -1,11 +1,9 @@
 cmake_minimum_required( VERSION 3.17 )
 project(PackageTest CXX)
 
-# Integrate with LLVM/clang tooling
-set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
-
-find_package(p2300 REQUIRED)
 find_package(Threads REQUIRED)
+include(${CMAKE_BINARY_DIR}/conanbuildinfo.cmake)
+conan_basic_setup()
 
 add_executable(test_p2300 test.cpp)
-target_link_libraries(test_p2300 PUBLIC P2300::p2300)
+target_link_libraries(test_p2300 ${CONAN_LIBS} Threads::Threads)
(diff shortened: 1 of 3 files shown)


paulbendixen:
@trxcllnt I hope this is agreeable with what you wanted to achieve, building should be possible with conan but it is no longer required (please double check) and creating the conan package of P2300 is still possible.

trxcllnt:
Sure, having conan is fine with me. I removed it because @ericniebler mentioned it's a common point of failure and we weren't really using it. If others do in fact find it helpful, we should keep it around!

test_package/CMakeLists.txt
```diff
@@ -1,11 +1,9 @@
 cmake_minimum_required( VERSION 3.17 )
 project(PackageTest CXX)
 
-# Integrate with LLVM/clang tooling
-set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
```
trxcllnt, line 5: Any reason not to turn this on automatically?

trxcllnt (changes requested):
What's happening in `test_package/CMakeLists.txt`? I assumed it was a smoketest to validate `P2300` can be consumed by other CMake projects, but it doesn't seem like it can be built without conan. What am I missing?

trxcllnt (approved):
Approving after a discussion with @ericniebler.

ericniebler:
I tried this out locally and (modulo a cmake version check) everything configured, built, and ran just fine without needing to jump through any Conan hoops. :tada:

paulbendixen:
> What's happening in `test_package/CMakeLists.txt`? I assumed it was a smoketest to validate `P2300` can be consumed by other CMake projects, but it doesn't seem like it can be built without conan. What am I missing?

Almost correct `test_package` is a smoke test for validating the conan package and is generated along with the boiler plate when creating a new project, which is why it doesn't really need to be nice for consumption in CMake.

PR #594, opened 2022-09-02
add `.clangd` configuration and `clangd.helper` to provide intellisense

trxcllnt:
Adds a `.clangd` configuration and `clangd.helper` executable target to help [`clangd`](https://clangd.llvm.org/) more consistently provide good intellisense results.

It's unfortunate we need to do hack `clangd` a bit by adding a dummy `.cpp` file -- if anyone knows a better solution, I'm all ears. I've tested this approach also works for subdirectories that define additional `CMakeLists.txt` with different compile flags than the root.

From the clangd logs. Line 2 indicates from which file `clangd` infers the `execution.hpp` intellisense compile flags:
```
I[09:18:35.323] Loaded compilation database from std_execution/./compile_commands.json
I[09:18:35.324] ASTWorker building file std_execution/include/execution.hpp version 1 with command inferred from std_execution/examples/_clangd_helper_file.cpp
[std_execution/build]
/usr/lib/llvm-16/bin/c++ --driver-mode=g++ -Istd_execution/include -w -O3 -DNDEBUG -fcolor-diagnostics -Wall -ftemplate-backtrace-limit=0 -c -std=gnu++20 -ferror-limit=0 -ferror-limit=0 -resource-dir=/usr/lib/llvm-16/lib/clang/16.0.0 -- std_execution/include/execution.hpp
```

### DIFF
--- a/.clangd
+++ b/.clangd
@@ -1,0 +1,49 @@
+# https://clangd.llvm.org/config
+
+# Apply a config conditionally to all C files
+If:
+  PathMatch: .*\.(c|h)$
+
+---
+
+# Apply a config conditionally to all C++ files
+If:
+  PathMatch: .*\.(c|h)pp
+
+---
+
+# Apply a config conditionally to all CUDA files
+If:
+  PathMatch: .*\.cuh?
+CompileFlags:
+  Add:
+    # Allow variadic CUDA functions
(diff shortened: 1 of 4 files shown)


ajschmidt8:
Ignore this comment. Just ensuring that webhooks are working correctly for this repository.

PR #599, opened 2022-09-07
Add `ensure_started` implementation

trxcllnt:
This PR adds [`ensure_started`](https://brycelelbach.github.io/wg21_p2300_std_execution/std_execution.html#spec-execution.senders.adapt.ensure_started). Need to add more tests, but figured I'd get it up for review.

### DIFF
--- a/include/__memory.hpp
+++ b/include/__memory.hpp
@@ -1,0 +1,159 @@
+/*
+ * Copyright (c) NVIDIA
+ *
+ * Licensed under the Apache License Version 2.0 with LLVM Exceptions
+ * (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ *   https://llvm.org/LICENSE.txt
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#pragma once
+
+#include <__utility.hpp>
+
+#include <atomic>
+#include <memory>
+#include <new>
+
+namespace _P2300 {
+  namespace __ptr {
+    template <class _Ty>
+      struct __make_intrusive_ptr_t;
+
+    template <class _Ty>
+      struct __enable_intrusive_from_this;
+
+    template <class _Ty>
+      struct __control_block {
+        alignas(_Ty) unsigned char __value_[sizeof(_Ty)];
+        std::atomic<unsigned long> __refcount_;
+
+        template <class... _Us>
+          explicit __control_block(_Us&&... __us)
+            noexcept(noexcept(_Ty{__declval<_Us>()...}))
+            : __refcount_(1u) {
+            // Construct the value *after* the initialization of the
+            // atomic in case the constructor of _Ty calls
+            // __intrusive_from_this() (which increments the atomic):
+            ::new ((void*) __value_) _Ty{(_Us&&) __us...};
+          }
+        ~__control_block() {
+          __value().~_Ty();
+        }
+
+        _Ty& __value() const noexcept {
+          return *(_Ty*) __value_;
+        }
+      };
+
+    template <class _Ty>
+      class __intrusive_ptr {
+        using _UncvTy = std::remove_cv_t<_Ty>;
+        friend struct __make_intrusive_ptr_t<_Ty>;
+        friend struct __enable_intrusive_from_this<_UncvTy>;
+
+        __control_block<_UncvTy>* __data_{nullptr};
+
+        explicit __intrusive_ptr(__control_block<_UncvTy>* __data) noexcept
+          : __data_(__data) {}
+
+        void __addref_() noexcept {
+          if (__data_) {
+            __data_->__refcount_.fetch_add(1, std::memory_order_relaxed);
+          }
+        }
+
+        void __release_() noexcept {
+          if (__data_ && (1u == __data_->__refcount_.fetch_sub(1, std::memory_order_release))) {
+            std::atomic_thread_fence(std::memory_order_acquire);
+            delete __data_;
+          }
+        }
+
+       public:
+        __intrusive_ptr() = default;
+        bool operator==(const __intrusive_ptr&) const = default;
+
+        __intrusive_ptr(__intrusive_ptr&& __that) noexcept
+          : __data_(std::exchange(__that.__data_, nullptr)) {}
+
+        __intrusive_ptr(const __intrusive_ptr&& __that) noexcept
+          : __data_(__that.__data_) {
+          __addref_();
+        }
+
+        __intrusive_ptr& operator=(__intrusive_ptr&& __that) noexcept {
+          __release_();
+          __data_ = std::exchange(__that.__data_, nullptr);
+        }
+
+        __intrusive_ptr& operator=(const __intrusive_ptr&& __that) noexcept {
+          __release_();
+          __data_ = __that.__data_;
+          __addref_();
+        }
+
+        ~__intrusive_ptr() {
+          __release_();
+        }
+
+        void reset() noexcept {
+          __release_();
+          __data_ = nullptr;
+        }
+
+        _Ty* get() const noexcept {
+          return &__data_->__value();
+        }
+
+        _Ty* operator->() const noexcept {
+          return &__data_->__value();
+        }
+
+        _Ty& operator*() const noexcept {
+          return __data_->__value();
+        }
+      };
+
+    template <class _Ty>
+      struct __enable_intrusive_from_this {
+        __intrusive_ptr<_Ty> __intrusive_from_this() noexcept {
+          static_assert(0 == offsetof(__control_block<_Ty>, __value_));
+          _Ty* __this = static_cast<_Ty*>(this);
+          __intrusive_ptr<_Ty> __p{(__control_block<_Ty>*) __this};
+          __p.__addref_();
+          return __p;
+        }
+
+        __intrusive_ptr<const _Ty> __intrusive_from_this() const noexcept {
+          static_assert(0 == offsetof(__control_block<_Ty>, __value_));
+          const _Ty* __this = static_cast<const _Ty*>(this);
+          __intrusive_ptr<const _Ty> __p{(__control_block<_Ty>*) __this};
+          __p.__addref_();
+          return __p;
+        }
+      };
+
+    template <class _Ty>
+      struct __make_intrusive_ptr_t {
+        template <class... _Us>
+            requires std::constructible_from<_Ty, _Us...>
+          __intrusive_ptr<_Ty> operator()(_Us&&... __us) const {
+            using _UncvTy = std::remove_cv_t<_Ty>;
+            return __intrusive_ptr<_Ty>{::new __control_block<_UncvTy>{(_Us&&) __us...}};
+          }
+      };
+  } // namespace __ptr
+
+  using __ptr::__intrusive_ptr;
+  using __ptr::__enable_intrusive_from_this;
+  template <class _Ty>
+    inline constexpr __ptr::__make_intrusive_ptr_t<_Ty> __make_intrusive_ptr {};
+
+} // namespace _P2300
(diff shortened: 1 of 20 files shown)


include/execution.hpp
```diff
@@ -2852,7 +2852,217 @@ namespace _P2300::execution {
...
+
+        __variant_t __data_;
+        in_place_stop_source __stop_source_{};
+
+        std::atomic<void*> __op_state1_;
+        connect_result_t<_Sender, __receiver> __op_state2_;
+
+        explicit __sh_state(_Sender& __sndr)
+          : __op_state1_{nullptr}
+          , __op_state2_(connect((_Sender&&) __sndr, __receiver{*this}))
```
msimberg, line 2928: Ignoring the question of whether `ensure_started` should exist in the first place, `__receiver` also needs to increment the reference count of the shared state to be able to support the case where the `ensure_started` sender is destroyed without ever connecting it. Otherwise the sender holds the only (reference counted) reference to the shared state, and it'll likely be destroyed before the receiver is signalled. This requirement definitely convolutes the implementation and I'm not quite sure how to do it with a `shared_ptr`, but in HPX and pika we've done it relatively cleanly with an intrusive pointer.

include/execution.hpp
```diff
@@ -2852,7 +2852,217 @@ namespace _P2300::execution {
...
+
+        static void __notify(__operation_base* __self) noexcept {
+          __operation *__op = static_cast<__operation*>(__self);
+          __op->__on_stop_.reset();
+
+          std::visit([&](const auto& __tupl) noexcept -> void {
+            std::apply([&](auto __tag, const auto&... __args) noexcept -> void {
+              __tag((_Receiver&&) __op->__recvr_, __args...);
+            }, __tupl);
+          }, __op->__shared_state_->__data_);
```
ericniebler, line 2983: The intention is for `ensure_started` to return a single-shot sender (a sender that can only ever be connected to a receiver once). Because of that, the `ensure_started` spec requires that these values be moved in:

```suggestion
          std::visit([&](auto& __tupl) noexcept -> void {
            std::apply([&](auto __tag, auto&... __args) noexcept -> void {
              __tag((_Receiver&&) __op->__recvr_, std::move(__args)...);
            }, __tupl);
          }, __op->__shared_state_->__data_);
```

include/execution.hpp
```diff
@@ -2852,7 +2852,217 @@ namespace _P2300::execution {
...
+        using _Sender = __t<_SenderId>;
+        using __sh_state_ = __sh_state<_SenderId>;
+        template <class _Receiver>
+          using __operation = __operation<_SenderId, __x<remove_cvref_t<_Receiver>>>;
+
+        _Sender __sndr_;
+        std::shared_ptr<__sh_state_> __shared_state_;
+
+      public:
+        template <__decays_to<__sender> _Self, receiver _Receiver>
```
ericniebler, line 3023: For a single-shot sender, we must accept only rvalue instances of `__sender`.

```suggestion
        template <same_as<__sender> _Self, receiver _Receiver>
```

include/execution.hpp
```diff
@@ -2852,7 +2852,217 @@ namespace _P2300::execution {
...
+            requires receiver_of<_Receiver, completion_signatures_of_t<_Self, __empty_env>>
+          friend auto tag_invoke(connect_t, _Self&& __self, _Receiver&& __recvr)
+            noexcept(std::is_nothrow_constructible_v<decay_t<_Receiver>, _Receiver>)
+            -> __operation<_Receiver> {
+            return __operation<_Receiver>{(_Receiver &&) __recvr,
+                                          __self.__shared_state_};
```
ericniebler, line 3029: We can only `connect` once, so transfer ownership here instead of sharing it.

```suggestion
                                          std::move(__self.__shared_state_)};
```

include/execution.hpp
```diff
@@ -2852,7 +2852,217 @@ namespace _P2300::execution {
...
+            noexcept(__nothrow_callable<_Tag, const _Sender&, _As...>)
+            -> __call_result_if_t<tag_category<_Tag, forwarding_sender_query>, _Tag, const _Sender&, _As...> {
+            return ((_Tag&&) __tag)(__self.__sndr_, (_As&&) __as...);
+          }
+
+        template <class... _Tys>
+        using __set_value_t = completion_signatures<set_value_t(const decay_t<_Tys>&...)>;
+
+        template <class _Ty>
+        using __set_error_t = completion_signatures<set_error_t(const decay_t<_Ty>&)>;
```
ericniebler, line 3045:
```suggestion
        template <class... _Tys>
        using __set_value_t = completion_signatures<set_value_t(decay_t<_Tys>&&...)>;

        template <class _Ty>
        using __set_error_t = completion_signatures<set_error_t(decay_t<_Ty>&&)>;
```

include/execution.hpp
```diff
@@ -2852,7 +2852,217 @@ namespace _P2300::execution {
...
+
+        template <__decays_to<__sender> _Self, class _Env>
```
ericniebler, line 3047:
```suggestion
        template <same_as<__sender> _Self, class _Env>
```

include/execution.hpp
```diff
@@ -2852,7 +2852,217 @@ namespace _P2300::execution {
...
+          friend auto tag_invoke(get_completion_signatures_t, _Self&&, _Env) ->
+            make_completion_signatures<
+              _Sender,
+              make_env_t<with_t<get_stop_token_t, in_place_stop_token>>,
+              completion_signatures<set_error_t(const std::exception_ptr&)>,
```
ericniebler, line 3052:
```suggestion
              completion_signatures<set_error_t(std::exception_ptr&&)>,
```

include/execution.hpp
```diff
@@ -2867,6 +3077,15 @@ namespace _P2300::execution {
         noexcept(nothrow_tag_invocable<ensure_started_t, _Sender>) {
         return tag_invoke(ensure_started_t{}, (_Sender&&) __sndr);
       }
+      template <sender _Sender>
+        requires (!__tag_invocable_with_completion_scheduler<ensure_started_t, set_value_t, _Sender>) &&
+          (!tag_invocable<ensure_started_t, _Sender>)
```
ericniebler, line 3082:
```suggestion
        requires (!__tag_invocable_with_completion_scheduler<ensure_started_t, set_value_t, _Sender>) &&
          (!tag_invocable<ensure_started_t, _Sender>) &&
          sender<_Sender, make_env_t<with_t<get_stop_token_t, in_place_stop_token>>>
```

We repeat the type of the `ensure_started` environment in a number of places. Probably best to create an alias.

ericniebler (changes requested):
This is really nice. I suggested a few changes including some due to the spec being confused about whether `ensure_started` should return a single-shot sender or not. I think it should since it's more efficient, and we can always pass the `ensure_started` sender to `split` to get the other behavior.

include/execution.hpp
```diff
@@ -2877,7 +2878,246 @@ namespace _P2300::execution {
...
+
+        template <class _Ty>
+        using __set_error_t = completion_signatures<set_error_t(decay_t<_Ty>&&)>;
+
+        template <same_as<__sender> _Self, class _Env>
+          friend auto tag_invoke(get_completion_signatures_t, _Self&&, _Env) ->
+            make_completion_signatures<
+              _Sender,
+              __env,
+              completion_signatures<set_error_t(std::exception_ptr&&)>,
```
msimberg, line 3098:
```suggestion
              completion_signatures<set_error_t(std::exception_ptr&&)
                                    set_stopped_t()>,
```

include/execution.hpp
```diff
@@ -2892,6 +3132,15 @@ namespace _P2300::execution {
         noexcept(nothrow_tag_invocable<ensure_started_t, _Sender>) {
         return tag_invoke(ensure_started_t{}, (_Sender&&) __sndr);
       }
+      template <sender _Sender>
+        requires (!__tag_invocable_with_completion_scheduler<ensure_started_t, set_value_t, _Sender>) &&
+          (!tag_invocable<ensure_started_t, _Sender>)
+      __sender<_Sender> operator()(_Sender&& __sndr) const {
```
msimberg, line 3138: There's also a minor optimization opportunity with making `ensure_started` the identity function when the given sender is already an `ensure_started` sender.

include/__memory.hpp
```diff
@@ -0,0 +1,172 @@
...
+          }
+        }
+
+       public:
+        __intrusive_ptr() = default;
+
+        __intrusive_ptr(__intrusive_ptr&& __that) noexcept
+          : __data_(std::exchange(__that.__data_, nullptr)) {}
+
+        __intrusive_ptr(const __intrusive_ptr&& __that) noexcept
```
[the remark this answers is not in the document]
ericniebler: Copy pasta. Thx.

include/execution.hpp
```diff
@@ -2877,7 +2878,246 @@ namespace _P2300::execution {
...
+        }
+
+        void __notify() noexcept {
+          void* __old{nullptr};
+          void* const __completion_state = static_cast<void*>(this);
+          bool const changed = __op_state1_.compare_exchange_weak(
+            __old, __completion_state,
+            std::memory_order_release,
+            std::memory_order_acquire);
+          if (changed && __old != nullptr) {
```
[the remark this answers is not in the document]
ericniebler: `ensure_started` kicks off the operation and returns a sender. Two things will then happen independently, possibly on different threads: (a) the operation finishes, and (b) the returned sender is connected and started. If (a) happens first, then this code will see `__old` as `nullptr`. If (b) happens first, then `__old` points to the operation state that was started in (b).

include/execution.hpp
ericniebler: Indeed, thx.

include/__memory.hpp
```diff
@@ -0,0 +1,172 @@
...
+          : __data_(__that.__data_) {
+          __addref_();
+        }
+
+        __intrusive_ptr& operator=(__intrusive_ptr&& __that) noexcept {
+          __release_();
+          __data_ = std::exchange(__that.__data_, nullptr);
+        }
+
+        __intrusive_ptr& operator=(const __intrusive_ptr&& __that) noexcept {
```
ericniebler, line 96:
```suggestion
        __intrusive_ptr& operator=(const __intrusive_ptr& __that) noexcept {
```

include/execution.hpp
```diff
@@ -2877,7 +2878,247 @@ namespace _P2300::execution {
...
+
+        template <class _Ty>
+        using __set_error_t = completion_signatures<set_error_t(decay_t<_Ty>&&)>;
+
+        template <same_as<__sender> _Self, class _Env>
+          friend auto tag_invoke(get_completion_signatures_t, _Self&&, _Env) ->
+            make_completion_signatures<
+              _Sender,
+              __env,
+              completion_signatures<set_error_t(std::exception_ptr&&)
```
ericniebler, line 3098:
```suggestion
              completion_signatures<set_error_t(std::exception_ptr&&),
```

include/execution.hpp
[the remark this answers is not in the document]
msimberg: 100% agreed. However:

If (a) happens first the `compare_exchange_weak` will return `true` since `__op_state1` contained the expected value (i.e. `nullptr` in `__old). `changed && __old != nullptr` is `false` in this case. This is ok. The notify should not be called.

If (b) happens first the `compare_exchange_weak` will return `false` since `__op_state1` does _not_ contain the expected value (again `nullptr`). `changed && __old != nullptr` is again `false`. In this second case the notify should, however, be called.

I think it would be enough to make the predicate only `if (__old != nullptr)`. Still slightly unsure if `__op_state1` needs to be updated/checked again after that before notifying...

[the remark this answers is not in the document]
ericniebler: OK yup, you're right. Thanks.

include/execution.hpp
```diff
@@ -2877,7 +2878,247 @@ namespace _P2300::execution {
...
+        }
+
+        void __notify() noexcept {
+          void* __old{nullptr};
+          void* const __completion_state = static_cast<void*>(this);
+          bool const changed = __op_state1_.compare_exchange_weak(
+            __old, __completion_state,
+            std::memory_order_release,
+            std::memory_order_acquire);
+          if (changed && __old != nullptr) {
```
ericniebler, line 2967: I think this should be:
```suggestion
          void* const __completion_state = static_cast<void*>(this);
          void* const __old =
            __op_state1_.exchange(__completion_state, std::memory_order_acq_rel);
          if (__old != nullptr) {
```

@msimberg ?

msimberg: Yeah, that looks good to me!

PR #602, opened 2022-09-07
[awaitables] fix: get_completion_signatures to test for await-ability with a promise type.

SAtacker:
fixes: https://github.com/brycelelbach/wg21_p2300_std_execution/issues/598

Assumptions: 
1. Used Promise type and Environment analogously  

Questions:
1. If the above assumptions are false as there may or may not be a state associated with the promise type, what should be a better approach?

TODO:
1. Change std_execution.bs

Signed-off-by: Shreyas Atre <shreyasatre16@gmail.com>

### DIFF
--- a/include/execution.hpp
+++ b/include/execution.hpp
@@ -470,8 +470,6 @@ namespace _P2300::execution {
   // sender's metadata. If sender<S> and sender<S, Ctx> are both true, then they
   // had better report the same metadata. This completion signatures wrapper
   // enforces that at compile time.
-  // However when env is a promise type, the evaluation will lead to different
-  // results and hence not required.
   template <class _Sender, class _Env>
     struct __checked_completion_signatures {
      private:
--- a/include/execution.hpp
+++ b/include/execution.hpp
@@ -456,8 +456,7 @@ namespace _P2300::execution {
         __one_of<
           _WithoutEnv,
           _WithEnv,
-          dependent_completion_signatures<no_env>> ||
-          __awaitable<_Sender,__env_or_void<_Env>>);
+          dependent_completion_signatures<no_env>>);
      public:
       using __t = _WithEnv;
     };
--- a/include/execution.hpp
+++ b/include/execution.hpp
@@ -365,7 +365,7 @@ namespace _P2300::execution {
             __required_completions) } ->
           __receiver_concepts::__is_valid_completion<remove_cvref_t<_Receiver>>;
       };
-  
+
   /////////////////////////////////////////////////////////////////////////////
   // [execution.sndtraits]
   namespace __get_completion_signatures {
@@ -427,9 +427,8 @@ namespace _P2300::execution {
 
   template <class _Sender, class _Env = no_env>
     concept sender =
-      ((__sender<_Sender, _Env> &&
-      __sender<_Sender, no_env> ) ||
-      __awaitable<_Sender, _Env>) &&
+      __sender<_Sender, _Env> &&
+      __sender<_Sender, no_env> &&
       move_constructible<remove_cvref_t<_Sender>> &&
       constructible_from<remove_cvref_t<_Sender>, _Sender>;
 
--- a/include/execution.hpp
+++ b/include/execution.hpp
@@ -254,7 +254,12 @@ namespace _P2300::execution {
     template <class, class = void>
       __types<> __test(...);
 
-    struct __dependent {};
+    // BUGBUG not to spec!
+    struct __dependent {
+      bool await_ready();
+      void await_suspend(const __coro::coroutine_handle<no_env>&);
+      __dependent await_resume();
+    };
   } // namespace __completion_signatures
 
   template <same_as<no_env>>
@@ -361,9 +366,6 @@ namespace _P2300::execution {
           __receiver_concepts::__is_valid_completion<remove_cvref_t<_Receiver>>;
       };
   
-  template<class _Env>
-    using __env_or_void = __if_c<std::is_same_v<_Env, no_env>, void, _Env>;
-
   /////////////////////////////////////////////////////////////////////////////
   // [execution.sndtraits]
   namespace __get_completion_signatures {
@@ -381,7 +383,7 @@ namespace _P2300::execution {
       template <class _Sender, class _Env = no_env>
         requires (__with_tag_invoke<_Sender, _Env> ||
                   __with_member_alias<_Sender> ||
-                  __awaitable<_Sender,__env_or_void<_Env>>)
+                  __awaitable<_Sender, _Env>)
       constexpr auto operator()(_Sender&&, const _Env& = {}) const noexcept {
         static_assert(sizeof(_Sender), "Incomplete type used with get_completion_signatures");
         static_assert(sizeof(_Env), "Incomplete type used with get_completion_signatures");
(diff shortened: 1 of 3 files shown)


include/execution.hpp
```diff
@@ -420,8 +424,9 @@ namespace _P2300::execution {
 
   template <class _Sender, class _Env = no_env>
     concept sender =
-      __sender<_Sender, _Env> &&
-      __sender<_Sender, no_env> &&
+      ((__sender<_Sender, _Env> &&
+      __sender<_Sender, no_env> ) ||
+      __awaitable<_Sender,__env_or_void<_Env>>) &&
```
ericniebler, line 429: I think we want awaitables to satisfy the `__sender` concept checks here. We shouldn't need an extra test for awaitables if `get_completion_signatures` is doing the right thing.

include/execution.hpp
```diff
@@ -360,6 +360,9 @@ namespace _P2300::execution {
             __required_completions) } ->
           __receiver_concepts::__is_valid_completion<remove_cvref_t<_Receiver>>;
       };
+  
+  template<class _Env>
+    using __env_or_void = __if_c<std::is_same_v<_Env, no_env>, void, _Env>;
```
ericniebler, line 365: I'm curious why this is needed. What happens if you pass `no_env` to the `__awaitable` concept?

ericniebler (changes requested):

include/execution.hpp
SAtacker: Because `__sender` concepts evaluates the `completion_signatures` using the following 
https://github.com/brycelelbach/wg21_p2300_std_execution/blob/2aaf6eb131a1d04c0b3e0dc00f2ce3b960477da1/include/execution.hpp#L298-L304
We can see that for all the awaitables `no_env` and a specific environment would give different completion_signatures`

include/execution.hpp
SAtacker: I guessed that `__awaitable<Sender, Env>` would be specific to that Env, if it is no_env then the promise type is void. If a `no_env` is used for this concept, then it wouldn't satisfy as `await_transform` wouldn't exist.
The following overload will be considered instead of `void*` one
https://github.com/brycelelbach/wg21_p2300_std_execution/blob/2aaf6eb131a1d04c0b3e0dc00f2ce3b960477da1/include/coroutine.hpp#L63-L73

Please correct me if I am wrong.

ericniebler: > If a `no_env` is used for this concept, then it wouldn't satisfy as `await_transform` wouldn't exist.

It would, actually! The concept is checking whether a call to `__get_awaiter` with an argument of type `_Promise*` is well-formed. Clearly _this_ overload would not be selected when `_Promise` is `no_env` because `no_env` doesn't have an `await_transform` member.

But! There is another overload of `__get_awaiter` that doesn't require that, and it accepts a `void*` instead of a `_Promise*`. And (here's the key bit) a `_Promise*` pointer is implicitly convertible to a `void*` one.

So the `__awaitable` concept doesn't require you to map `no_env` into a `void` first. You can just pass `no_env` to it.

test/concepts/test_awaitables.cpp
```diff
@@ -0,0 +1,162 @@
...
+      ex::__q1<ex::__id>, ex::__q<ex::__variant>>;
+
+  static_assert(std::is_same_v<error_types_of, error_types>);
+}
+
+template <typename Signatures>
+void test_awaitable_sender2(Signatures) {
+  // is_sender_v relies on get_completion_signatures and is not true
+  // even if a sender is an awaitable if it's promise type is not
+  // used to evaluate the concept awaitable
```
ericniebler, line 89: This is the real problem with your patch. The `sender` concept is used without an environment to test whether a type has "sender-ness" even when it is only really a sender when used with a certain environment type.

For instance, imagine `Sndr` is the type of the expression `execution::read(execution::get_scheduler)` (which is a sender that reads a scheduler out of the environment and passes it to the receiver's `set_value). The type `Sndr` is only really a sender -- i.e., can only be connected to a receiver -- if that receiver's environment has a scheduler to give it. And yet `sender<Sndr>` would evaluate to true.

That's because the `read` sender hooks `get_completion_signatures` to return `dependent_completion_signatures<no_env>` when passed a `no_env`. That is basically like the sender responding to "are you a sender?" with "reply hazy. ask again later."

This is used to constrain sender adaptors to only accept things that look sender-ish, even though there isn't yet an environment available. Without that, we would have no way to reject obvious nonsense like `execution::then( emphatically-not-a-sender, fn )`.

With your patch, you won't be able to use just about _any_ sender adaptor with a type that is awaitable in only certain coroutine types.

The real solution would look a little like the current one with `dependent_completion_signatures`. I can think of an awaitable type implementing a dummy `operator co_await` that returns a known awaitable type that the library knows to mean "ask again later". Let's call it `dependent_awaitable`, and it can look like this:

```c++
struct dependent_awaitable {
  bool await_ready();
  dependent_awaitable await_resume();
};
```

Then the `__awaitable< Awaitable, no_env >` test will succeed (yay!) and we can have `get_completion_signatures` return `dependent_completion_signatures<no_env>` for such cases.

Make sense?

Think I'll code this up to be sure I'm not smoking dope.

ericniebler (changes requested):

include/execution.hpp
SAtacker: Got it, (and tried it), how silly of me.

test/concepts/test_awaitables.cpp
SAtacker: > This is the real problem with your patch. The `sender` concept is used without an environment to test whether a type has "sender-ness" even when it is only really a sender when used with a certain environment type.
> 
> For instance, imagine `Sndr` is the type of the expression `execution::read(execution::get_scheduler)` (which is a sender that reads a scheduler out of the environment and passes it to the receiver's `set_value`). The type `Sndr` is only really a sender -- i.e., can only be connected to a receiver -- if that receiver's environment has a scheduler to give it. And yet `sender<Sndr>` would evaluate to true.
> 
> That's because the `read` sender hooks `get_completion_signatures` to return `dependent_completion_signatures<no_env>` when passed a `no_env`. That is basically like the sender responding to "are you a sender?" with "reply hazy. ask again later" instead of "yes" or "no".
> 
> This is used to constrain sender adaptors to only accept things that look sender-ish, even though there isn't yet an environment available. Without that, we would have no way to reject obvious nonsense like `execution::then( emphatically-not-a-sender, fn )`.
> 
> With your patch, you won't be able to use just about _any_ sender adaptor with a type that is awaitable in only certain coroutine types.
> 
> The real solution would look a little like the current one with `dependent_completion_signatures`. I can think of an awaitable type implementing a dummy `operator co_await` that returns a known awaitable type that the library knows to mean "ask again later". Let's call it `dependent_awaitable`, and it can look like this:
> 
> ```c++
> struct dependent_awaitable {
>   bool await_ready();
>   void await_suspend(std::coroutine_handle<no_env>);
>   dependent_awaitable await_resume();
> };
> ```
> 
> Then the `__awaitable< Awaitable, no_env >` test will succeed (yay!) and we can have `get_completion_signatures` return `dependent_completion_signatures<no_env>` for such cases.
> 
> Make sense?
> 
> Think I'll code this up to be sure I'm not smoking dope.

Yes, it makes sense. That's really good, I couldn't think that far. Thanks for explaining in detail, appreciate it.

include/execution.hpp
```diff
@@ -463,6 +470,8 @@ namespace _P2300::execution {
   // sender's metadata. If sender<S> and sender<S, Ctx> are both true, then they
   // had better report the same metadata. This completion signatures wrapper
   // enforces that at compile time.
+  // However when env is a promise type, the evaluation will lead to different
+  // results and hence not required.
```
ericniebler, line 474:
```suggestion
```

PR #613, opened 2022-09-23
task: fix leak upon stops

williamspatrick:
When a task is awaiting a Sender which stops, the task_awaitable is
never await_resume'd (because that would resume the task).  In that
case the task_awaitable is destructed and the coroutine_handle is
leaked.  Fix this by destroying up any residual coroutine_handle in
the task_awaitable destructor.

### DIFF
--- a/examples/task.hpp
+++ b/examples/task.hpp
@@ -268,6 +268,12 @@ private:
     __coro::coroutine_handle<_promise> coro_;
     std::optional<awaiter_context_t<_promise, ParentPromise>> context_{};
 
+    ~_task_awaitable() {
+        if (_coro) {
+            _coro.destroy();
+        }
+    }
+
     static std::false_type await_ready() noexcept {
       return {};
     }
--- a/examples/task.hpp
+++ b/examples/task.hpp
@@ -269,9 +269,8 @@ private:
     std::optional<awaiter_context_t<_promise, ParentPromise>> context_{};
 
     ~_task_awaitable() {
-        if (_coro) {
-            _coro.destroy();
-        }
+      if (coro_)
+        coro_.destroy();
     }
 
     static std::false_type await_ready() noexcept {


examples/task.hpp
```diff
@@ -268,6 +268,12 @@ class basic_task {
     __coro::coroutine_handle<_promise> coro_;
     std::optional<awaiter_context_t<_promise, ParentPromise>> context_{};
 
+    ~_task_awaitable() {
+        if (_coro) {
+            _coro.destroy();
```
miscco, line 273: Thanks a lot for the fix. Here the member is named `coro_`, so we need to change that to
 ```suggestion
        if (coro_) {
            coro_.destroy();
```

File: examples/task.hpp [hunk 2]
```diff
@@ -268,6 +268,12 @@ class basic_task {
...
+        }
+    }
```
ericniebler, line 275: member name and formatting.

```suggestion
    ~_task_awaitable() {
      if (coro_)
        coro_.destroy();
    }
```

ericniebler:
A million thanks for this. I completely missed it.

Also, I'm looking for feedback about this sample `task` implementation, if you have any to give. Have you used it in production by any chance? It might be what `std::task` ends up looking like.

examples/task.hpp
williamspatrick: > Thanks a lot for the fix. Here the member is named `coro_`, so we need to change that to

Sorry.  That's what I get for having an epiphany at 1am and then, when it solved the issue, excitedly trying to transpose it to this task implementation.

williamspatrick:
> A million thanks for this. I completely missed it.

I wanted to ask, do we need a move-ctor that exchanges the handle too?  I don't know if there are any cases where the awaitable is moved.  It doesn't seem like it as best I can tell.  I'm slightly worried that we could end up with two task_awaitables each holding the same co-routine handle and the first one to destruct destroys the (still used) handle.

> Also, I'm looking for feedback about this sample `task` implementation, if you have any to give. Have you used it in production by any chance? It might be what `std::task` ends up looking like.

I couldn't get it to work in GCC, which is primarily what our project uses, so I gave up on trying to pull it in.  I did use it as a reference and wrote a simplified version of it (I use the promise as the env, which may or may not be the right thing to do, and punted on most of the stop-token stuff).  I think from a functionality stand-point it would meet my needs if it compiled in GCC.

> Have you used it in production by any chance?

What I'm currently playing with is using an import of everything under `include` to then build an async library around for DBus.  I've been working on it for a few weeks and haven't put any of it in production yet, but so far in testing it is going well.  Like I said above I've had to write my own `task` but almost all the DBus operations I've been able to model as Sender/Receiver.

Our project does most of our development in a Gerrit instance, which is mirrored to Github later.  I can either point you to the pending Gerrit reviews or let you know when it gets merged and mirrored.  I'll probably be merging it next week sometime.

PR #617, opened 2022-09-30
Update github pages link

ccotter:
github forwards some links after a repository moves locations, but doesn't forward the github pages link

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 `std::execution`, the proposed C++ framework for asynchronous and parallel programming.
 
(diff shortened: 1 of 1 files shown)


ericniebler:
Thanks. I'm trying to figure out what needs to happen on our end to cover our bases from a legal POV when accepting outside contributions. We _may_ need a Developer Certificate of Origin. Obvs, seems like overkill for a trivial change like this, but when corporate legal is concerned, I've learned to step carefully.

ccotter:
Ha! I understand! Since this is trivial, feel free to proceed however you'd like. I just happened to notice when I clicked the link yesterday.

PR #620, opened 2022-10-03
Support for the nvc++ compiler and a simple cuda stream scheduler

ericniebler:


CMakeLists.txt
```diff
@@ -191,8 +200,8 @@ set(test_sourceFiles
     test/queries/test_get_forward_progress_guarantee.cpp
     test/async_scope/test_dtor.cpp
     test/async_scope/test_spawn.cpp
-    test/async_scope/test_spawn_future.cpp
-    test/async_scope/test_empty.cpp
+    # test/async_scope/test_spawn_future.cpp
```
miscco, line 203: Ditto rather work with `#ifdef` in the tests than removing them completely

examples/cuda/examples/bulk.cpp
```diff
@@ -0,0 +1,36 @@
...
+namespace ex = std::execution;
+namespace stream = example::cuda::stream;
+
+int main() {
+  using example::cuda::is_on_gpu;
+
+  stream::context_t stream_context{};
+  ex::scheduler auto sch = stream_context.get_scheduler();
+
+  auto bf = [](int lbl) {
```
miscco, line 15: Given that these are example functions I would appreciate some more descriptive names.

`bulk_function` and `then_function` are not great but better than `bf` and `tf`

examples/cuda/examples/maxwell/cpp.cuh
```diff
@@ -0,0 +1,54 @@
...
+             std::string_view method) {
+  fields_accessor accessor = grid.accessor();
+
+  auto initializer = grid_initializer(dt, accessor);
+  for (std::size_t i = 0; i < accessor.cells; i++) {
+    initializer(i);
+  }
+
+  report_performance(grid.cells, n_inner_iterations * n_outer_iterations, method, 
+                     [&]() {
```
miscco, line 31: Can we move the lambda out of the function call so that the indentation is less horrible

examples/cuda/examples/maxwell/snr.cuh
```diff
@@ -0,0 +1,173 @@
...
+    friend void tag_invoke(ex::set_error_t, sink_receiver &&, auto&&) noexcept {}
+    friend void tag_invoke(ex::set_stopped_t, sink_receiver &&) noexcept {}
+    friend ex::__empty_env tag_invoke(ex::get_env_t, sink_receiver) noexcept {
+      return {};
+    }
+  };
+
+  template <class SenderId, class ReceiverId>
+    struct operation_state_t {
+      using Sender = std::__t<SenderId>;
```
miscco, line 42: Given that ADL-isolation is an important tool for S&R, we should provide some standard facility that is not __ugly

When we were trying to get another reference implementation going I used this hacked up thingy:

https://github.com/dietmarkuehl/P2300/blob/main/include/p2300/non_associated.hpp

No change requested, just food for thought

examples/cuda/schedulers/detail/common.cuh
```diff
@@ -0,0 +1,400 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#pragma once
+
+#include <execution.hpp>
+
+#include <type_traits>
```
miscco, line 20: Should those become `#include <cuda/std/type_traits>`

test/test_common/receivers.hpp
```diff
@@ -395,9 +396,11 @@ template <typename F>
 struct fun_receiver {
   F f_;
 
-  template <typename... Ts>
+  template <typename... Ts _NVCXX_CAPTURE_PACK(Ts)>
```
miscco, line 399: 😿

ericniebler: Tell me about it.

examples/cuda/schedulers/detail/common.cuh
ericniebler: This is a question for @senior-zero.

examples/cuda/examples/maxwell/snr.cuh
ericniebler: You mean, standardize it as part of P2300?

CMakeLists.txt
ericniebler: This test no longer needs to be excluded.

CMakeLists.txt
```diff
@@ -159,7 +168,7 @@ set(test_sourceFiles
     test/concepts/test_concept_operation_state.cpp
     test/concepts/test_concepts_sender.cpp
     test/concepts/test_awaitables.cpp
-    test/types/test_type_async_scope.cpp
+    # test/types/test_type_async_scope.cpp
```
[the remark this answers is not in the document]
ericniebler: This test is no longer commented out.

examples/cuda/examples/maxwell/snr.cuh
miscco: Yes, a user should be able to write senders without having to use implementation reserved keywords or rebuild standard facilities by hand.

Also it would help to actually constraint based on a given standard facility so that it is less often forgotten

PR #622, opened 2022-10-05
The Great Renaming

ericniebler:
Reorganize the library and assign meaningful namespaces.

This library contains three kinds of things of interest to developers:

1. Things approved by the C++ Committee for future inclusion in C++,
2. Generally useful things that are not (yet) approved by the Committee, or that will not be proposed, and
3. Vendor-specific utilities and customizations.

This commit puts (1) in namespace `::stdexec`, (2) in namespace `::exec`, and (3) in `::<vendor>exec` (e.g., `::nvexec` for NVIDIA).

In addition, the headers have been reorganized to follow suit. (1) is in `stdexec/`, (2) is in `exec/`, and (3) is in `<vendor>exec/`. 

Finally, the entities that appear in the latest draft revision of [P2300](https://brycelelbach.github.io/wg21_p2300_execution/execution.html) are imported into their corresponding `std::` namespace with using declarations.


miscco:
I am not too happy with `__detail`. We should try and follow the libc++ modularization efforts and use `__senders` or something less generic.

That would help a lot once we want to bring it back to libcu++

examples/cuda/schedulers/detail/bulk.cuh
```diff
@@ -98,7 +98,7 @@ template <class SenderId, std::integral Shape, class FunId>
 
     template <class Self, class Env>
       using completion_signatures =
-        std::execution::__make_completion_signatures<
+        _P2300::execution::__make_completion_signatures<
```
miscco: Do we want to add a shorthand for `_P2300::execution`?

like `_EXEC` or similar?

examples/cuda/schedulers/detail/reduce.cuh
```diff
@@ -73,7 +73,7 @@ namespace reduce_ {
 
         static constexpr std::size_t value =
           _P2300::__v<
-            std::execution::__gather_sigs_t<
+            _P2300::execution::__gather_sigs_t<
               std::execution::set_value_t,
```
miscco, line 77: I believe that should also be `_P2300::execution::set_value_t`

ericniebler: I don't think you're commenting on the most recent commit. "`_P2300`" isn't a thing after this PR.

examples/cuda/schedulers/detail/bulk.cuh
ericniebler: You're looking at the wrong commit. This PR changes it to `stdexec`. Take a look at the qualification rationale I just added to the PR description and see if you agree.

PR #625, opened 2022-10-10
enable tasks on gcc

williamspatrick:
- task: awaiter_context: remove unused parameter
- task: fix compile on gcc
- hello_coro: re-enable on gcc

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,6 +13,17 @@ jobs:
           prebuild_command: |
             wget -O - https://apt.kitware.com/keys/kitware-archive-latest.asc 2>/dev/null | apt-key add -;
             apt update && apt install -y --no-install-recommends git cmake;
+  build-gcc-12:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@master
+      - uses: docker://lucteo/action-cxx-toolkit.gcc12:latest
+        with:
+          cc: gcc-12
+          checks: build test
+          prebuild_command: |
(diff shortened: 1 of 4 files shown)


williamspatrick:
I seem to not have trouble locally with GCC-12. Is there a desire to keep GCC pinned back at 11 or can I also attempt to upgrade the CI to 11?

ericniebler:
Thanks for this! Maybe enable `task` only on gcc >= 12 and add gcc-12 to the test matrix.

williamspatrick:
I need to play with this a bit in order to figure out enabling GCC-12 in CI. Going to close it to reduce the noise upstream here and then I’ll reopen later.

ericniebler:
I don't think the CMake/CPM problems in CI were necessarily caused by your change. I wonder if @trxcllnt could weigh in on the use of the RAPIDS cmake utilities.

williamspatrick:
> I don't think the CMake/CPM problems in CI were necessarily caused by your change. I wonder if @trxcllnt could weigh in on the use of the RAPIDS cmake utilities.

I thought I'd be able to enable gcc-12 in the CI just by specifying it but it turns out that the underlying github action being used doesn't support gcc-12 yet.  I looked into modifying that but it is all based on Ubuntu 20.04, in order to get some latest CMake debs, but Ubuntu 20.04 doesn't have gcc-12 packages and Ubuntu 22.04 isn't supported by the Kitware apt repo.  I'll probably just fix this commit sequence up to attempt to resolve the problems I'm seeing such that this runs in gcc-12 but we won't be able to test it in CI for the time being.

PR #628, opened 2022-10-11
add a scoped exec::on algorithm and exec::complete_on

ericniebler:


test/exec/test_on.cpp
```diff
@@ -0,0 +1,341 @@
...
+}
+TEST_CASE("exec::on can be called with ref scheduler", "[adaptors][exec::on]") {
+  inline_scheduler sched;
+  auto snd = exec::on(sched, ex::just(13));
+  auto op = ex::connect(std::move(snd), expect_value_receiver{13});
+  ex::start(op);
+  // The receiver checks if we receive the right value
+}
+
+TEST_CASE("exec::on forwards set_error calls", "[adaptors][exec::on]") {
```
LeeHowes, line 126: You don't test an incoming error, only one that the scheduler generates. You also don't test the scheduler transition behaviour if the incoming sender completes with an error. Does it still complete on the receiver scheduler in the case?

test/exec/test_on.cpp
```diff
@@ -0,0 +1,341 @@
...
+  inline_scheduler sched1{};
+  error_scheduler sched2{};
+  stopped_scheduler sched3{};
+
+  check_sends_stopped<false>(exec::on(sched1, ex::just(1)));
+  check_sends_stopped<true>(exec::on(sched2, ex::just(2)));
+  check_sends_stopped<true>(exec::on(sched3, ex::just(3)));
+}
+
+TEST_CASE("exec::on transitions back to the receiver's scheduler", "[adaptors][exec::on]") {
```
LeeHowes, line 175: Without a scheduler on the receiver, does `on(sched, closure)` complete on sched? The test above doesn't fail to compile in that situation, so I assume it does without any other alternative.

ericniebler: Yes

test/exec/test_on.cpp
ericniebler: I've added tests to cover those cases. In short: it still completes on the receiver's scheduler, even when the input sender completes with an error.

ericniebler:
Still todo: replace `inline_scheduler` hack with explicitly provided environments to all eager sender algorithms.

kirkshoop:
Lgtm!

senior-zero (approved):

ccotter:
I think this change might have broke things for clang on Mac. It looks like the version of sys/cdefs.h on Mac defines `__single` to nothing:

```
#if __has_include(<ptrcheck.h>)
...
#else
...
#define __single
```

Do you have a suggestion for a fix here (aside from reverting the name, which doesn't appear to match the convention for other names in __meta.hpp)? Thanks!

PR #629, opened 2022-10-11
task: enable on gcc

williamspatrick:
- task: awaiter_context: remove unused parameter
- task: fix compile on gcc
- hello_coro: re-enable on gcc-12 or higher

### DIFF
--- a/examples/hello_coro.cpp
+++ b/examples/hello_coro.cpp
@@ -13,7 +13,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-#if defined(__GNUC__) && !defined(__clang__)
+#if defined(__GNUC__) && !defined(__clang__) && (__GNUC__ < 12)
 int main() { return 0; }
 #else
 
--- a/include/exec/task.hpp
+++ b/include/exec/task.hpp
@@ -143,7 +143,7 @@ namespace exec {
       struct __default_task_context_impl::__awaiter_context<void> {
         template <class _Ty>
           explicit __awaiter_context(
-            __default_task_context_impl& __self, _Ty&) noexcept
+            __default_task_context_impl&, _Ty&) noexcept
           {}
 
         template <__indirect_stop_token_provider _ParentPromise>
--- a/include/exec/task.hpp
+++ b/include/exec/task.hpp
@@ -202,6 +202,22 @@ namespace exec {
         std::variant<std::monostate, __void, std::exception_ptr> __data_{};
       };
 
+    // Specify basic_task's completion signatures
+    //   This is only necessary when basic_task is not generally awaitable
+    //   owing to constraints imposed by its _Context parameter.
+    template <class... _Ts>
+      using __task_traits_t =
+        std::execution::completion_signatures<
+          std::execution::set_value_t(_Ts...),
+          std::execution::set_error_t(std::exception_ptr),
+          std::execution::set_stopped_t()>;
+
+    template <class _Ty> requires (!std::is_same_v<void, _Ty>)
+    auto __task_completion_sigs() -> __task_traits_t<_Ty>;
+
+    template <class _Ty> requires (std::is_same_v<void, _Ty>)
+    auto __task_completion_sigs() -> __task_traits_t<>;
+
     ////////////////////////////////////////////////////////////////////////////////
     // basic_task
     template <class _Ty, class _Context = default_task_context<_Ty>>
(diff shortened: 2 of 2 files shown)


williamspatrick:
This is the same content as with #625, but with the example code only re-enabled on GCC-12+.  Currently, lucteo/action-cxx-toolkit doesn't support GCC-12.  Opened lucteo/action-cxx-toolkit#2 to start discussion on how to enable GCC-12.

include/exec/task.hpp
```diff
@@ -306,18 +322,8 @@ namespace exec {
...
-        //   owing to constraints imposed by its _Context parameter.
-        template <class... _Ts>
-          using __task_traits_t =
-            std::execution::completion_signatures<
-              std::execution::set_value_t(_Ts...),
-              std::execution::set_error_t(std::exception_ptr),
-              std::execution::set_stopped_t()>;
-
         friend auto tag_invoke(std::execution::get_completion_signatures_t, const basic_task&, auto)
-          -> std::conditional_t<std::is_void_v<_Ty>, __task_traits_t<>, __task_traits_t<_Ty>>;
```
ericniebler, line 320: I'm curious why you had to change the return type here.

examples/hello_coro.cpp
```diff
@@ -13,7 +13,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-#if defined(__GNUC__) && !defined(__clang__)
```
ericniebler, line 16: If we have determined that coroutines are unusable on gcc-11 and earlier, then we should probably just set `_STD_NO_COROUTINES` to 1 in `stdexec/coroutine.hpp` and remove this check here. Line 25 below checks `_STD_NO_COROUTINES`.

ericniebler (changes requested):

examples/hello_coro.cpp
williamspatrick: Will do.

include/exec/task.hpp
williamspatrick: I tried to add the details to the commit message of the commit where I did this.

The original failure on GCC was:
```
task.hpp:370:17: error: invalid parameter type ‘void’
  370 |     friend auto tag_invoke(std::execution::get_completion_signatures_t,
      |                 ^~~~~~~~~~
task.hpp:370:17: error: in declaration ‘std::conditional_t<is_void_v<_Ty>, stdexec::completion_signatures<stdexec::__receivers::set_value_t(), stdexec::__receivers::set_error_t(std::__exception_ptr::exception_ptr), stdexec::__receivers::set_stopped_t()>, stdexec::completion_signatures<stdexec::__receivers::set_value_t(_Ty), stdexec::__receivers::set_error_t(std::__exception_ptr::exception_ptr), stdexec::__receivers::set_stopped_t()> > exec::__task::tag_invoke(stdexec::__get_completion_signatures::get_completion_signatures_t, const basic_task<_Ty, _Context>&, auto:82)’
```

GCC is unhappy about the evaluation of `__task_traits_t<void>` even though it is inside the `std::conditional_t`.  I promoted the return type to a template specialization so that the code never attempts to evaluate `__task_traits_t<void>` and only does `__task_traits_t<>` (when _Ty is void).

examples/hello_coro.cpp
williamspatrick: Actually, I don't think it is that co-routines are unusable on gcc-11 and earlier.  With gcc-11, we get the following compile failure with something in task.hpp:

```
In file included from /github/workspace/examples/hello_coro.cpp:26:
/github/workspace/include/exec/task.hpp:121:43: error: redefinition of ‘struct exec::__task::__default_task_context_impl::__awaiter_context<_ParentPromise>’
  121 |       struct __default_task_context_impl::__awaiter_context<_ParentPromise> {
      |                                           ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/github/workspace/include/exec/task.hpp:93:43: note: previous definition of ‘struct exec::__task::__default_task_context_impl::__awaiter_context<_ParentPromise>’
   93 |       struct __default_task_context_impl::__awaiter_context<_ParentPromise> {
      |                                           ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/github/workspace/include/exec/task.hpp:134:43: error: redefinition of ‘struct exec::__task::__default_task_context_impl::__awaiter_context<_ParentPromise>’
  134 |       struct __default_task_context_impl::__awaiter_context<_ParentPromise> {
      |                                           ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/github/workspace/include/exec/task.hpp:93:43: note: previous definition of ‘struct exec::__task::__default_task_context_impl::__awaiter_context<_ParentPromise>’
   93 |       struct __default_task_context_impl::__awaiter_context<_ParentPromise> {
      |                                           ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

I don't think setting `_STD_NO_COROUTINES` on gcc-11 is entirely appropriate.  Any ideas on how we might resolve this issue?  It isn't obvious to me what gcc is unhappy about.

ericniebler: You're right. This is a gcc bug with concepts. I can work around it, and I've confirmed locally that gcc-11 can handle coroutines just fine. I'll make a PR and merge, then this PR is not needed.

ericniebler:
Superseded by #631

PR #634, opened 2022-10-13
fix `when_all` and `async_scope::spawn_future` with correct initialization order

zyctree:
Yet it will forever wait on `sync_wait(std::move(allDone));`

I'm not sure these `on(sch, ...)` are needed, or it is a bug inside `when_all`. 
A more simple case, `sync_wait(when_all(scope.spawn_future(...)));` will never return.

### DIFF
--- a/include/exec/async_scope.hpp
+++ b/include/exec/async_scope.hpp
@@ -458,13 +458,12 @@ namespace exec {
         }
 
         std::optional<in_place_stop_callback<__forward_stopped>> __forward_scope_;
-        connect_result_t<_Sender, __future_rcvr<__x<_Sender>>> __op_;
-
         std::mutex __mutex_;
         __future_step __step_ = __future_step::__created;
         std::unique_ptr<__future_state> __no_future_;
         std::variant<std::monostate, __future_result_t<_Sender>, std::tuple<set_stopped_t>> __data_;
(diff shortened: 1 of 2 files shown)


ericniebler:
Interesting. When I build with clang, the scope example runs fine for me. But when compiled with gcc, I see the hand you're talking about. @kirkshoop can you weigh in on why this hangs?

kirkshoop:
it hangs in `in_place_stop_source::__try_lock_unless_stop_requested_`

the inner loop never sees an `__old_state_ == 0`

Not too familiar with this code, it could be a lifetime issue or a bug in the inplace_stop_source or a bug in the godegen..

zyctree:
with release build, it hangs; with debug build, it runs fine
i'm using 'Apple clang version 14.0.0'

lewissbaker:
Generally, stop-callbacks on the receiver's stop token shouldn't be registered until start() is called on the op state, although I don't think we have actually codified this guideline anywhere.

But good find! We should do a more general audit of algorithms to make sure child op states appear ss the last data members in all algorithms.

kirkshoop:
Maybe inplace_stop_source can add some validity checks for the state to report rather than hang?

It would be nice if private constructors for inplace_stop_source and a make function would force users to add explicit construction of the members that would show that it was being used before it was constructed.

PR #635, opened 2022-10-14
refactor async_scope to be less demanding on the compiler

ericniebler:
This change also eliminates the use of `void*` and `const_cast` and corrects an error where incoming senders were tested against `__empty_env` instead of the real environment the `async_scope` uses (which has a stop token).


kirkshoop (approved):
Looks great!

ericniebler:
Thanks for looking! 🙂

PR #642, opened 2022-10-19
reorder write/transfer again for `exec::on`

ericniebler:


senior-zero (approved):

griwes:
Doesn't the write before first transfer also need to be moved?

senior-zero:
> Doesn't the write before first transfer also need to be moved?

Not really, the intent was to have a non-interrupted pipeline segment `transfer(new) | sender | transfer(old)`. Having `write` after transfer introduced uncustomizable sender in the pipeline which is a bit of an issue for stream scheduler. The order will change when we have `connect_transform`, since environment will affect `transfer` customization. For now there should be no difference.

kirkshoop:
@ericniebler why didn't __start_fn::__call need the same change?

PR #646, opened 2022-10-21
Use GPU-enabled action runners

trxcllnt:
This PR adds a CI job that uses the new GPU-accelerated action runners to build with `nvc++` and run the `nvexec` CUDA tests.


ericniebler (approved):
LGTM!

PR #652, opened 2022-10-22
Make `clangd` use NVHPC macros in nvexec files, temporaril…

trxcllnt:
* Makes `clangd` use the `nvc++` variants of macros when viewing files under `nvexec` dirs
  * Fixes errors (like [this](https://gist.github.com/trxcllnt/2a8c669a9dcdfb7feb1f64e02aec60be)) I was seeing using `clang++` for intellisense on test files under  `test/nvexec`.
* Temporarily suppress shadow template parameter diagnostic
  * The `nvc++` version of [`_NVCXX_EXPAND_PACK`](https://github.com/NVIDIA/stdexec/blob/4c67901667e038749ffdf00a923499da402f52e2/include/stdexec/execution.hpp#L50-L55) causes a shadowed-template-parameter error. `nvc++` doesn't seem to mind, but probably worth fixing.

### DIFF
--- a/.clangd
+++ b/.clangd
@@ -30,6 +30,7 @@ CompileFlags:
   Add:
     - -x
     - cuda
+    - -D__NVCOMPILER
 
 ---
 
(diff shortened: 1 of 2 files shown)


ericniebler:
I'm typing those macros out once nvhpc 22.11 ships in a few weeks, so we'll need to deal with this then.

PR #670, opened 2022-11-01
New technique for erasing associated namespaces of class template parameters

ericniebler:
* instantiates fewer class templates overall
* significantly simplifies and shortens type names

Also, refactor the default `then` algorithm for better compile times.

Before the change, the expression:

```c++
just() | then(fn{}) | then(fn{}) | then(fn{})
```

has type:

```
stdexec::__then::__sender<stdexec::_X<stdexec::__then::__sender<stdexec::_X<stdexec::__then::
__sender<stdexec::_X<stdexec::__just::__sender<stdexec::__receivers::set_value_t>>::__t_, std
exec::_X<fn>::__t_>>::__t_, stdexec::_X<fn>::__t_>>::__t_, stdexec::_X<fn>::__t_
```

After, it has type:

```
stdexec::__then::__sender<stdexec::__then::__sender<stdexec::__then::__sender<stdexec::__just
::__sender<>, fn>, fn>, fn>::__t
```

So far, I have only brought the `just` senders and `then` into this new scheme. Some grunt work is necessary to apply it to all algorithms.

### DIFF
--- a/MAINTAINERS.md
+++ b/MAINTAINERS.md
@@ -1,0 +1,93 @@
+# Maintainers' Guide
+
+This is a place to put design rationale and code idioms that maintainers of
+stdexec should follow.
+
+## ADL isolation
+
+Stdexec makes use of the [`tag_invoke`](http://wg21.link/p1895) mechanism to
+support end-user customization of algorithms. For the sake of compile-time
+performance, this mechanism requires care to ensure that overload sets stay
+small.
+
+### Hidden friend functions
+
+Define `tag_invoke` overloads for class types _in-situ_ as hidden friend
+functions. Keep in mind that when calling a customizable funtion that
+all the associated entities (classes and namespaces) of all the function's
+arguments are searched for `tag_invoke` overloads. That means that for
+a function call like:
+
+```c++
+connect(sender, receiver);
+```
+
+... the hidden friend functions of the _receiver_ will be considered in
+addition to those of the sender. This can sometimes lead to surprising and
+confusing compiler errors.
+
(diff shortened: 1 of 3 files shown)


kirkshoop:
Wow. The symbols.

MAINTAINERS.md
```diff
@@ -0,0 +1,93 @@
...
+    stdexec::__t<my_sender_id<stdexec::__id<Sender>, Arg>>;
+```
+
+Note that we use `stdexec::__id` to "encode" a type into an identifier
+before passing it as a template argument. And we use `stdexec::__t`
+to get back the original type from the identifier.
+
+> **Note:**
+>
+> We only really need to encode the type of the template arguments that
```
ericniebler, line 82:
```suggestion
> **Note**
>
> **Note:** We only really need to encode the type of the template arguments that
```

MAINTAINERS.md
```diff
@@ -0,0 +1,93 @@
...
+    stdexec::__t<my_sender_id<stdexec::__id<Sender>, Arg>>;
+```
+
+Note that we use `stdexec::__id` to "encode" a type into an identifier
+before passing it as a template argument. And we use `stdexec::__t`
+to get back the original type from the identifier.
+
+> **Note**
+>
+> **Note:** We only really need to encode the type of the template arguments that
```
ericniebler, line 82:
```suggestion
> We only really need to encode the type of the template arguments that
```

ericniebler:
> Wow. The symbols.

Way better, huh?

PR #680, opened 2022-11-06
[WIP] implement a lightweight form of connect_transform

ericniebler:


include/stdexec/execution.hpp
```diff
@@ -2737,13 +2869,40 @@ namespace stdexec {
                   ((_Self&&) __self).__fun_});
           }
 
+          template <class _Self, class _Scheduler>
+              requires tag_invocable<bulk_t, _Scheduler, __member_t<_Self, _Sender>, _Shape, __member_t<_Self, _Fun>>
+            using __connect_transform_sender_t =
+              tag_invoke_result_t<bulk_t, _Scheduler, __member_t<_Self, _Sender>, _Shape, __member_t<_Self, _Fun>>;
+
+          // NOT TO SPEC
+          template <scheduler _Scheduler, __decays_to<__sender> _Self, receiver _Receiver, class _Tag = bulk_t>
```
senior-zero, line 2878: This customization is never found (related to all other customizations as well). I believe it should be:

```suggestion
          template <scheduler _Scheduler, __decays_to<__t> _Self, receiver _Receiver, class _Tag = bulk_t>
```
With this change `stdexec::on(sch, snd)` actually customize `snd`.

senior-zero (changes requested):

include/stdexec/execution.hpp
ericniebler: Thanks lmk if you support this problem anyplace else. I have yet to add tests for all this stuff.

ericniebler:
This turns out to be a dead end. Waiting until we can fix this issue properly.

PR #716, opened 2022-12-12
Fix move-only test for `when_all`

msimberg:
Apologies, I only checked that the new test in #714 compiles (and doesn't compile before the change), but didn't run it. The test that I added had a superfluous copy-paste `connect` in it... This fixes that.


ericniebler:
thanks

PR #721, opened 2022-12-18
proposal for async_scope

kirkshoop:
Proposal to separate the immovable async_scope with on_empty and stop_source and a 'handle' with nest, spawn, and spawn_future.

Thoughts?

### DIFF
--- a/examples/scope.cpp
+++ b/examples/scope.cpp
@@ -44,7 +44,8 @@ class noop_receiver : receiver_adaptor<noop_receiver> {
 
 int main() {
   exec::static_thread_pool ctx{1};
-  exec::async_scope scope;
+  exec::async_scope context;
+  auto scope = context.get_nester();
 
   scheduler auto sch = ctx.get_scheduler();                               // 1
 
@@ -53,7 +54,7 @@ int main() {
   sender auto printVoid = then(begin,
     []()noexcept { printf("void\n"); });                                  // 3
 
-  sender auto printEmpty = then(on(sch, scope.on_empty()),
+  sender auto printEmpty = then(on(sch, context.on_empty()),
     []()noexcept{ printf("scope is empty\n"); });                         // 4
 
   printf("\n"
@@ -87,18 +88,18 @@ int main() {
     sender auto nest = scope.nest(begin);
     (void)nest;
   }
-  sync_wait(scope.on_empty());
+  sync_wait(context.on_empty());
 
   {
     sender auto nest = scope.nest(begin);
     auto op = connect(std::move(nest), noop_receiver{});
     (void)op;
   }
-  sync_wait(scope.on_empty());
+  sync_wait(context.on_empty());
 
   {
     sender auto nest = scope.nest(begin);
     sync_wait(std::move(nest));
   }
-  sync_wait(scope.on_empty());
+  sync_wait(context.on_empty());
 }
--- a/examples/server_theme/on_transfer.cpp
+++ b/examples/server_theme/on_transfer.cpp
@@ -89,7 +89,8 @@ int main() {
 
   std::array<std::byte, 16*1024> buffer;
 
-  exec::async_scope scope;
+  exec::async_scope context;
+  auto scope = context.get_nester();
 
   // Fake a couple of requests
   for (int i = 0; i < 10; i++) {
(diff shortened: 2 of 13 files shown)


ericniebler:
What is the benefit of this split? If we make changes here, I would like see a factorization of (a) necessary and sufficient basis operations, and (b) generic algorithms. Is this a step in that direction?

kirkshoop:
@ericniebler the purpose is to follow the stop_source and execution context pattern of having an immovable object that holds state and provides the operations that the owner needs and a handle object that that is freely copyable and only provides the operations that non-owners need. 

The basis-operations question is not answered here yet. 

I would prefer for the basis operations to take the handle type added here instead of a raw ref to the existing async_scope object.

kirkshoop:
The owner/non-owner is also mutable/const. The methods on each also follows the stop_source and execution context pattern.

`request_stop()` is a mutable operation, `spawn()` etc.. are const methods. 

`when_empty()` & `on_empty()` are const methods. They are on the mutable object because they are used to define the lifetime (when is it safe to destruct the immovable object?). 

The `jthread` object added `get_stop_token()` and `request_stop()` as a convenience - (I would prefer not to put convenience on the object. IMO convenience should be external to the object). I added them to `async_scope` to follow the `jthread` precedent.

ericniebler:
That all makes sense, thanks. The only thing that is giving me pause here is that the number 1 reason folks have an `async_scope` is to spawn work. This change makes it awkward to do that. Instead of:

```c++
scope.spawn(work);
```

folks need to:

```c++
scope.get_nester().spawn(work);
```

Of course, we could add `spawn(scope, work)`, but that seems unsatisfying. Maybe I'm just too used to being able to spawn work without creating a nester first.

(Aside: I don't like the name "nester". Even something nondescript like "handle" would be an improvement, IMO.)

kirkshoop:
Yes. Nester is an awful name. It was derived from scheduler. 

schedule/scheduler
nest/nester

The nester/get_nester allows an environment to provide a scope. 

If spawn(nester, sender) is overloaded to work with any object that satisfies nester_provider then the scope or the environment etc can be used. 

I still think that nest is the only basis operation. spawn() and spawn_future() are not. They would be cpos that take an environment and a nester.

LeeHowes:
> folks need to:
> 
> ```c++
> scope.get_nester().spawn(work);
> ```

I think that's true only for slideware. The real difference here is in giving a moveable handle isn't it? So real code would be more aligned with the way schedulers are used. So we might instead do:

```c++
get_nester(receiver).spawn(s, get_env(receiver));
```

but if we're doing that get_nester is part of the environment, which does suggest that a free function spawn that grabs the nester from the environment works, and we only actually store a nester if we want one in a local object for some reason. Either way, we aren't touching the scope at all any more than we touch the execution context.

Which means, our names are very bad. How about `async_scope` being the local name rather than `nester` and `lifetime_context` or something like that for the immovable `execution_context` analogue?

ericniebler:
> Can we reasonably nest in a scope?
>
> What else do we do with scopes? We open them. We construct variables in them. We inject into them?

This has been suggested before, but `emplace`?

LeeHowes:
I think emplacing into a scope has a reasonable ring to it, so that combo would work.

Then, a followup question is whether the scope should be able to carry an environment, given that this is the copyable locally constructed entity anyway. So maybe `get_scope()` and `get_scope_with_environment()`? We'd need an `async_scope_with_environment` type/concept though with a simpler `spawn` variant.

kirkshoop:
`emplace` would be overloaded. Perhaps it is okay?

I would not like to bind an env to the scope. We only need an env when doing eager spawns. That is where the env is added. 

If an env is bound to the scope, then nest needs to describe how the scope env and the env supplied by the receiver are merged.

ericniebler:
> How about `async_scope` being the local name rather than `nester` and `lifetime_context` or something like that for the immovable `execution_context` analogue?

I suspect once we know the basis operations, and hence the concepts, naming will be easier. but...

`async_scope` ==> `async_scope_[context|source]`
`nester`      ==> `async_scope`

?

LeeHowes:
That's certainly the obvious mapping. It doesn't have to be, I was thinking more `scheduler`->`execution_context`

LeeHowes:
Thinking further... splitting this objecting into two risks causing confusion about lifetimes. The handle can go out of scope, but that's what people will want to store. So we're going to end up with a lot of designs where we do just have these two objects tied close to the use site. That reduces the value a lot compared with the `context`/`scheduler` split.

ericniebler:
awaiting an improved design from @kirkshoop

PR #728, opened 2023-01-16
Initial Clang CUDA support for stream scheduler

senior-zero:
This PR adds initial support for Clang CUDA. To compile with Clang CUDA you need to configure stdexec as follows:
```bash
cmake -DSTDEXEC_ENABLE_CUDA=ON \
      -DCMAKE_CUDA_ARCHITECTURES=86 \
      -DCMAKE_CXX_COMPILER=clang++-16 \
      -DCMAKE_CUDA_COMPILER=clang++-16 \
      ..
```
Clang revealed some issues with `let_*` implementation which this PR addresses as well. Currently, all stream scheduler tests pass when compiled with Clang CUDA. Note, that only minimal function annotations were made. Extending tests (especially `let_*` ones) might require annotating more functions in `execution.hpp` header. 

One disabled example is reduce. There's no support for `__reduce_{add,min,max}_sync` on Clang CUDA side. These intrinsics are required by CUB. Later, we might workaround this by reducing non-integral types (like `double`), which do not rely on this intrinsics. 

CI for Clang CUDA will be supported in a separate PR.


test/nvexec/variant.cpp
```diff
@@ -70,26 +70,27 @@ __global__ void kernel(V* v, T alt) {
 
 TEST_CASE("variant emplaces alternative from GPU", "[cuda][stream][containers][variant]") {
   using variant_t = variant_t<int, double>;
-  variant_t v{};
-  THROW_ON_CUDA_ERROR(cudaHostRegister(&v, sizeof(variant_t), cudaHostRegisterDefault));
+  variant_t *v{};
+  THROW_ON_CUDA_ERROR(cudaMallocManaged(&v, sizeof(variant_t)));
+  new (v) variant_t();
```
trxcllnt, line 75: Does clangd not report the `clang(need_header_before_placement_new)` error here for you @senior-zero?

senior-zero: It does, but inclusion of `new` doesn't fix the issue for me.

include/nvexec/detail/queue.cuh
```diff
@@ -101,7 +101,8 @@ namespace nvexec::STDEXEC_STREAM_DETAIL_NS {
     struct producer_t {
       task_base_t** tail_;
 
-      void operator()(task_base_t* task) {
+      STDEXEC_DETAIL_CUDACC_HOST_DEVICE //
+      inline void operator()(task_base_t* task) {
```
ericniebler, line 105: I wonder why `inline` is necessary here. Isn't it implicit for member functions defined _in situ_?

include/nvexec/stream/common.cuh
```diff
@@ -496,7 +498,7 @@ namespace nvexec {
               : operation_state_base_t<OuterReceiverId>((outer_receiver_t&&)out_receiver, context_state, true)
               , storage_(queue::make_host<variant_t>(this->status_, context_state.pinned_resource_))
               , task_(queue::make_host<task_t>(this->status_, context_state.pinned_resource_, receiver_provider(*this), storage_.get(), this->get_stream(), context_state.pinned_resource_).release())
-              , started_(ATOMIC_FLAG_INIT)
+              , started_ ATOMIC_FLAG_INIT
```
ericniebler, line 501: Is this even well-formed?! From cppreference.com, it appears that this flag isn't needed anymore since C++20. https://en.cppreference.com/w/cpp/atomic/ATOMIC_FLAG_INIT

And from [atomic.flags]:

> The macro can be used in the form:
> ```
> atomic_flag guard = ATOMIC_FLAG_INIT;
> ```
> It is unspecified whether the macro can be used in other initialization contexts.

include/nvexec/stream/let_xxx.cuh
```diff
@@ -104,7 +104,7 @@ namespace nvexec::STDEXEC_STREAM_DETAIL_NS {
                 stdexec::sender_to<stdexec::__minvoke<__result_sender<_Fun>, _As...>, _Receiver>
             friend void tag_invoke(_Tag, __t&& __self, _As&&... __as) noexcept {
               using result_sender_t = stdexec::__minvoke<__result_sender<_Fun>, _As...>;
-              using __tuple_t = stdexec::__decayed_tuple<_As...>;
+              // using __tuple_t = stdexec::__decayed_tuple<_As...>;
```
ericniebler, line 107: If this alias isn't used, just delete the line.

include/nvexec/stream/split.cuh
```diff
@@ -225,7 +225,7 @@ namespace nvexec::STDEXEC_STREAM_DETAIL_NS {
               noexcept(std::is_nothrow_move_constructible_v<Receiver>)
             : operation_base_t{nullptr, notify}
             , operation_state_base_t<ReceiverId>((Receiver&&)rcvr, shared_state->context_state_, false)
-            , shared_state_(move(shared_state)) {
+            , shared_state_(std::move(shared_state)) {
```
ericniebler, line 228: Why `std::move` here but `::cuda::std::move` elsewhere?

include/stdexec/execution.hpp
```diff
@@ -253,8 +253,10 @@ namespace stdexec {
       template <class _Fn, class... _Args>
         using __f = __minvoke<_Fn, _Args...>;
 
+      // TODO Extract annotation
```
ericniebler, line 256: Is this a TODO for this commit or a later one?

test/nvexec/variant.cpp
```diff
@@ -70,26 +70,27 @@ __global__ void kernel(V* v, T alt) {
...
+  kernel<<<1, 1>>>(v, 42);
   THROW_ON_CUDA_ERROR(cudaDeviceSynchronize());
 
   visit([](auto alt) {
       REQUIRE(alt == 42);
-  }, v);
+  }, *v);
 
-  THROW_ON_CUDA_ERROR(cudaHostUnregister(&v));
+  THROW_ON_CUDA_ERROR(cudaFree(v));
```
ericniebler, line 93: This gets leaked if any of the above throws. We should have an RAII wrapper, or else use the `scope_guard` in `exec/scope.hpp`.

ericniebler (changes requested):
This is amazing, thank you!

I'm a little surprised at how few annotations you had to add to `execution.hpp`. How did you find those, trial and error? I wonder if there are other places that also need the annotation that are simply not exposed by our tests. How can we be sure?

trxcllnt (approved):

include/nvexec/stream/common.cuh
trxcllnt: This is defined as `#define ATOMIC_FLAG_INIT {false}` in `cuda/std/detail/libcxx/include/atomic`. Wrapping `{false}` in parens triggers a clang warning.

ericniebler: Yes, but that syntax is not portable, and cppreference says we don't even need it because default initialization should have the same effect.

include/nvexec/stream/split.cuh
senior-zero: I'm trying to use ::cuda::std::move in device code.

ericniebler (approved):

PR #730, opened 2023-01-16
Introduce get_attrs

ccotter:
Add get_attrs and some supporting machinery.
Add get_attrs support to many of the sender adaptors since they'll need it eventually.

### DIFF
--- a/include/stdexec/execution.hpp
+++ b/include/stdexec/execution.hpp
@@ -66,6 +66,28 @@ namespace stdexec {
 
   using std::remove_cvref_t;
 
+  // [exec.queries.queryable]
+  template <class T>
+  concept queryable = std::destructible<T>;
+
+  // [exec.fwd_env]
+  namespace __forwarding_query {
+    struct forwarding_query_t {
+      template <class _Query>
+      constexpr bool operator()(_Query __query) const noexcept {
+        if constexpr (tag_invocable<forwarding_query_t, _Query>) {
+          return tag_invoke(*this, (_Query&&) __query);
+        } else if constexpr (std::derived_from<_Query, forwarding_query_t>) {
+          return true;
+        } else {
+          return false;
+        }
+      }
+    };
+  }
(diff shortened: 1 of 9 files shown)


ccotter:
I've got a much bigger and messier branch that updates the `sender`, `sender_of` etc concepts according to what's in https://github.com/brycelelbach/wg21_p2300_execution/pull/17. This PR introduces some of the changes incrementally first.

ccotter:
@ericniebler let me know if this is roughly the direction you were thinking re #519. For now, I've introduced `get_attrs` with some initial integration points in addition to existing facilities like `forwarding_sender_query_t` . This would allow the attrs and non-attrs code to live side by side for a small amount of time. I'm comfortable with larger changesets too, depending on feedback/direction.

include/stdexec/execution.hpp
```diff
@@ -101,6 +123,31 @@ namespace stdexec {
...
+
+    struct get_attrs_t {
+      template <class _Sender>
+        requires tag_invocable<get_attrs_t, const _Sender&>
+        constexpr auto operator()(const _Sender& __sender) const
+          noexcept(nothrow_tag_invocable<get_attrs_t, const _Sender&>)
+          -> tag_invoke_result_t<get_attrs_t, const _Sender&> {
+          static_assert(queryable<tag_invoke_result_t<get_attrs_t, const _Sender&>>);
+          return tag_invoke(*this, __sender);
+        }
```
ericniebler, line 143: I'm making a change to the spec of `get_attrs` so that if <code><i>is-awaitable</i>&lt;S&gt;</code> is true, it returns <code><i>empty-attrs</i></code>. Would be good to make an analogous change here.

ericniebler (changes requested):
Otherwise looks great, thanks.

ericniebler (approved):

ccotter:
Thanks!

I'll work on advancing to the next set of incremental changes unless you have any concerns.

ericniebler (approved):

PR #732, opened 2023-01-17
Fix data race in static thread pool specialization of bulk

### LINKED ISSUE #729: Possible data race in static_thread_pool::bulk_enqueue
Hi, is it possible, that there is a data race in `exec::static_thread_pool::bulk_enqueue` ?

The problematic lines are

```cpp
  inline void static_thread_pool::bulk_enqueue(task_base* task, std::uint32_t n_threads) noexcept {
    for (std::size_t i = 0; i < n_threads; ++i) {
      threadStates_[i % available_parallelism()].push(task);
    }
  }
```

Each thread state has its own mutex which gets locked. But the same task is being modified in different thread states. `task_base` and the `__intrusive queue` are not using any synchronisation. Am I correct or do I miss something?

Thread Sanitizer catches this.

Edit: I used that in code such as

```cpp
  ex::sender auto sched = ex::schedule(scheduler);
  ex::sender auto left = sched | ex::bulk( /* ... */);
  ex::sender auto right = sched | ex::bulk(/* ... */);
  return ex::when_all(left, right);
```

senior-zero:
It is possible for data race to happen when concurrent thread enqueue a task into static thread pool while bulk receiver is running. The race happens on writing `_Next` and is addressed by separating bulk shared state and bulk tasks. I also took the liberty of fixing remeaning data races in our tests so that we can start using tsan in our CI.  

Closes #729

### DIFF
--- a/include/exec/static_thread_pool.hpp
+++ b/include/exec/static_thread_pool.hpp
@@ -117,59 +117,59 @@ namespace exec {
           using Receiver = stdexec::__t<ReceiverId>;
 
           struct bulk_task : task_base {
-              bulk_shared_state *sh_state_;
+            bulk_shared_state *sh_state_;
 
-              bulk_task(bulk_shared_state *sh_state)
-                : sh_state_(sh_state) {
-                this->__execute = [](task_base* t, std::uint32_t tid) noexcept {
-                  auto& sh_state = *static_cast<bulk_task*>(t)->sh_state_;
-                  auto total_threads = sh_state.num_agents_required();
+            bulk_task(bulk_shared_state *sh_state)
+              : sh_state_(sh_state) {
+              this->__execute = [](task_base* t, std::uint32_t tid) noexcept {
+                auto& sh_state = *static_cast<bulk_task*>(t)->sh_state_;
+                auto total_threads = sh_state.num_agents_required();
 
-                  auto computation = [&](auto&... args) {
-                    auto [begin, end] = even_share(sh_state.shape_, tid, total_threads);
-                    for (Shape i = begin; i < end; ++i) {
-                      sh_state.fn_(i, args...);
-                    }
-                  };
+                auto computation = [&](auto&... args) {
+                  auto [begin, end] = even_share(sh_state.shape_, tid, total_threads);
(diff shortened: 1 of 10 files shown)


include/exec/static_thread_pool.hpp
```diff
@@ -112,10 +112,66 @@ namespace exec {
           >;
 
       template <class SenderId, class ReceiverId, class Shape, class Fun, bool MayThrow>
-        struct bulk_shared_state : task_base {
+        struct bulk_shared_state {
           using Sender = stdexec::__t<SenderId>;
           using Receiver = stdexec::__t<ReceiverId>;
 
+          struct bulk_task : task_base {
+              bulk_shared_state *sh_state_;
```
ericniebler, line 120: prefer two-space indents. I can has .clang-format now? :-)

test/stdexec/algos/other/test_execute.cpp
```diff
@@ -49,16 +50,17 @@ TEST_CASE("execute works with schedulers that need to be triggered manually", "[
 
 TEST_CASE("execute works on a thread pool", "[other][execute]") {
   exec::static_thread_pool pool{2};
-  bool called{false};
+  std::atomic<bool> called{false};
   {
     // launch some work on the thread pool
-    ex::execute(pool.get_scheduler(), [&] { called = true; });
+    ex::execute(pool.get_scheduler(), [&] { called.store(true, std::memory_order_relaxed); });
```
ericniebler, line 56: Why store and load with relaxed semantics here, but sequential consistency elsewhere?

ericniebler (commented):
Looks good. One Q and one formatting quibble.

test/stdexec/algos/other/test_execute.cpp
senior-zero: I've started with care, then noticed that other places might requrie ordering and switched to sequential consistency not to spend much time :)

include/exec/static_thread_pool.hpp
senior-zero: Not yet :) I've reformatted the code, thank you for noticing that!

PR #738, opened 2023-01-19
Refactor get_completion_scheduler to work with attrs

ccotter:
One open question,

 - How should we handle backwards compatibility for any users outside of this repository? Assuming I can get the nvexec/test.CUDA code to compile cleanly, the changes I have in this branch change the `stdexec::scheduler` concept to require that `get_completion_scheduler(get_attrs(schedule(scheduler)))` is valid. Types which satisfied the scheduler concept will no longer satisfy without updating the type. If we need backwards compatibility, I could keep the concept(s) forwards and backwards compatible, e.g., by checking whether the completion scheduler comes from the sender itself, or the senders attributes.
 - ~I'm not sure how I can build or run test.CUDA. I managed to get the code to compile with the [gcc11-cuda11.8-nvhpc22.11-ubuntu22.04](https://github.com/NVIDIA/stdexec/blob/main/.github/workflows/ci.yml#L54) container, although I'm not always able to compile successfully (in one case, the compilation appears to succeed, but nvc++ bails with the error pasted below. In any case, I'm not sure if have a sufficient hardware setup to run test.CUDA (I'm on an intel macbook pro, or a redhat VM). I've never worked with GPUs before. Is github code spaces a viable option (mirroring what CI is doing)?~

### DIFF
--- a/include/stdexec/execution.hpp
+++ b/include/stdexec/execution.hpp
@@ -469,37 +469,6 @@ namespace stdexec {
             (_Variant*) nullptr))>;
   } // namespace __compl_sigs
 
-  /////////////////////////////////////////////////////////////////////////////
-  // get_attrs
-  namespace __attrs {
-    struct __empty_attrs {
-      using __t = __empty_attrs;
-      using __id = __empty_attrs;
-    };
-
-    struct get_attrs_t {
-      template <class _Sender>
-        requires tag_invocable<get_attrs_t, const _Sender&>
-        constexpr auto operator()(const _Sender& __sender) const
-          noexcept(nothrow_tag_invocable<get_attrs_t, const _Sender&>)
-          -> tag_invoke_result_t<get_attrs_t, const _Sender&> {
-          static_assert(queryable<tag_invoke_result_t<get_attrs_t, const _Sender&>>);
-          return tag_invoke(*this, __sender);
-        }
-
-      template <class _Sender>
-        requires (!tag_invocable<get_attrs_t, const _Sender&>) &&
-          __awaitable<_Sender, no_env_promise>
-        constexpr auto operator()(const _Sender& __sender) const
-          noexcept -> __empty_attrs {
-          return {};
-        }
-    };
-  } // namespace __attrs
-  inline constexpr __attrs::get_attrs_t get_attrs{};
-  using __attrs::get_attrs_t;
-  using __attrs::__empty_attrs;
-
   template <class _Ty>
     concept __is_completion_signatures =
       __is_instance_of<_Ty, completion_signatures>;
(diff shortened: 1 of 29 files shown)


ericniebler:
> How should we handle backwards compatibility for any users outside of this repository?

I suggest we leave the `no_env` and `dependent_completion_signatures` machinery in place for now (deprecating those types), and then we define `get_attrs` to first look for the `tag_invoke` overload, then to look for conformance with the current `__sender<S, no_env>` concept. If that concept is satisfied, then `get_attrs` would just return a reference to the sender itself.

ccotter:
That works. I don't think that changes much in this PR (this will be useful when we get around to changing the sender concept).

Do we want to additionally keep `get_completion_scheduler<CPO>(snd)` as a valid expression? With my changes here, that is no longer a valid expression. It could remain a valid expression by changing the CPO to search for and invoke `tag_invoke(get_completion_scheduler_t<CPO>, Sender)` if `get_attrs(snd)` is not valid.

ccotter:
Ah, never mind. I think I got it now :)

include/stdexec/execution.hpp
```diff
@@ -5589,7 +5535,15 @@ namespace stdexec {
...
           tag_invocable<sync_wait_t, _Sender>
-      tag_invoke_result_t<sync_wait_t, _Sender>
+      // TODO - for reasons I cannot figure out, the move_only_t test in
+      // test/nvexec/then.cpp appears to instantiate this version of
+      // operator(), even though the complation scheduler overload above is
+      // what actually is called. This leads the compilation to fail when
+      // computing tag_invoke_result_t in the return type.
+      // Using decltype(auto) avoids the tag_invoke_result_t expansion.
+      // Why does the compiler not respect the requires clause and avoid
+      // expanding the tag_invoke_result_t?
```
ericniebler, line 5545: Well, compilers are buggy. They really _shouldn't_ be trying to instantiate the return type when the requirements aren't met, but that wasn't the case until the resolution of a core issue (whose number I can't remember).

But. You should never get a hard error from a use of `tag_invoke_result_t` in a SFINAE context. If that's happening to you, then there is a bug elsewhere, and using `decltype(auto)` here is just hiding it.

ericniebler (changes requested):
This is starting to look really good. A few small issues.

include/stdexec/execution.hpp
ccotter: It might be CWG 2369, which I don't think clang implements yet.

ericniebler: Yes, that's the one.

include/stdexec/execution.hpp
```diff
@@ -668,6 +637,51 @@ namespace stdexec {
...
+          return tag_invoke(*this, __sender);
+        }
+
+      template <class _Sender>
+        requires (!tag_invocable<get_attrs_t, const _Sender&>) &&
+          __awaitable<_Sender, no_env_promise>
+        constexpr auto operator()(const _Sender& __sender) const
+          noexcept -> __empty_attrs {
+          return {};
+        }
```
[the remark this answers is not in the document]
ccotter: Isn't the `__awaitable` constrained overload to spec (the latest merged changes at least), and the only thing which is not to spec is the `__sender` overload (which would go away after a period of backwards compatibility support)? I can extend the comment I added to the `__sender` overload (and here as well), I just want to confirm what's not to spec in my version of the `__awaitable` overload here?

ccotter:
> Also, with this change, I think you can also change the sender concept to what is currently in the spec; that is, checking that get_attrs(sndr) is well-formed and returns a queryable.
 
 Will work on fleshing this out more fully, along with the `tag_invoke_result_t` failure.

include/stdexec/execution.hpp
ccotter: Ah well, as luck would have it, removing the `same_as<get_attrs_t> _Tag` bit from then.cuh ended up fixing this. Still need to dig deeper as to why though..

include/stdexec/execution.hpp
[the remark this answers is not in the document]
ericniebler: Correct, the `__sender` overload will go away eventually. Until then, you can comment out the `__awaitable` one since it will never be selected.

include/stdexec/execution.hpp
```diff
@@ -668,8 +637,59 @@ namespace stdexec {
...
+
+  /////////////////////////////////////////////////////////////////////////////
+  // [execution.senders]
   template <class _Sender, class _Env = no_env>
     concept sender =
+      // NOT TO SPEC:
+      // The sender related concepts are temporarily "in flight" being
+      // upgraded from P2300R5 to the get_attrs aware version of P2300.
+      requires (const remove_cvref_t<_Sender>& __sender) {
+        { get_attrs(__sender) } -> queryable;
```
ericniebler, line 691: Since `__sender` is already a concept here...

```suggestion
      requires (const remove_cvref_t<_Sender>& __sndr) {
        { get_attrs(__sndr) } -> queryable;
```

ericniebler (approved):
This is a significant contribution, thank you very much. :pray:

ccotter:
> I suggest we leave the `no_env` and `dependent_completion_signatures` machinery in place for now (deprecating those types)

With this change now merged, are you suggesting to leave these types in place for a (possibly indeterminate) amount of time, or that we can do away with these as long as the code builds within the repository (keeping them around and/or using transitory types and concepts to assist code within the repository to continue building)? I'm trying to gauge how much we care about users of this code outside of the repository.

ericniebler:
I mean leave them around (deprecated) for a fixed length of time so folks have a chance to migrate their own code to conform to the new `sender` concept.

PR #741, opened 2023-01-20
Fix bug in `intrusive_queue::push_back()`

trxcllnt:
This PR fixes the deadlock in `maxwell_cpu_mt --run-thread-pool-scheduler`.

### DIFF
--- a/include/stdexec/__detail/__intrusive_queue.hpp
+++ b/include/stdexec/__detail/__intrusive_queue.hpp
@@ -69,6 +69,8 @@ namespace stdexec {
         [[nodiscard]] _Item* pop_front() noexcept {
           STDEXEC_ASSERT(!empty());
           _Item* __item = std::exchange(__head_, __head_->*_Next);
+          // This should test if __head_ == nullptr, but due to a bug in
+          // nvc++'s optimization, `__head_` isn't assigned until later.
           if (__item->*_Next == nullptr) {
             __tail_ = nullptr;
           }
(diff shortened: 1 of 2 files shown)


include/stdexec/__detail/__intrusive_queue.hpp
```diff
@@ -87,7 +87,7 @@ namespace stdexec {
         void push_back(_Item* __item) noexcept {
           STDEXEC_ASSERT(__item != nullptr);
           __item->*_Next = nullptr;
-          if (__tail_ == nullptr) {
+          if (empty()) {
```
ericniebler, line 90: `empty()` checks whether `__head_` is `nullptr`. I haven't spotted the bug yet. It looks like when `__head_` is null, `__tail_` is also null. How do we get into a state where only one is null?

trxcllnt: I traced the issue to the thread enqueueing a task that was never consumed. I intuited the behavior of `push_back()` only setting `__head_` when `__tail_` is null could be a bug, especially since `__tail_` is set unconditionally at the bottom.

But as we've discussed offline, I never suspected the real culprit could be an `nvc++` bug in the `std::exchange()` call in `pop_front()`.

include/stdexec/__detail/__intrusive_queue.hpp
```diff
@@ -69,7 +69,9 @@ namespace stdexec {
         [[nodiscard]] _Item* pop_front() noexcept {
           STDEXEC_ASSERT(!empty());
           _Item* __item = std::exchange(__head_, __head_->*_Next);
-          if (__head_ == nullptr) {
+          // This should test if __head_ == nullptr, but due to a bug in
+          // nvc++'s optimization, `__head_` isn't assigned until later.
```
ericniebler, line 73:
```suggestion
          // nvc++'s optimization, `__head_` isn't assigned until later.
          // Filed as NVBug#3952534.
```

PR #742, opened 2023-01-20
Configure tests and examples in separate `CMakeLists.txt`

trxcllnt:
This PR moves test and example configuration into separate `CMakeLists.txt` files.

It also moves tests and examples behind flags (which default to `ON`). This allows other CMake projects to add stdexec as a subdirectory and skip building our tests and examples.

### DIFF
--- a/.clangd
+++ b/.clangd
@@ -30,6 +30,7 @@ CompileFlags:
   Add:
     - "-x"
     - "cuda"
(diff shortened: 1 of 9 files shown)


ericniebler (approved):

ericniebler:
Very nice, thanks.

PR #743, opened 2023-01-20
Add get_attrs to most senders

ccotter:


ccotter:
After this, we'll be able to add the requirement of `get_attrs` to the sender concept with this and #738

ericniebler:
Thanks!

PR #744, opened 2023-01-23
Make repeated `ensure_started` calls compile

### LINKED ISSUE #715: Repeated `ensure_started` calls don't compile after #673
The following no longer compiles after #673:
```
ex::sync_wait(ex::ensure_started(ex::ensure_started(ex::just())));
```
I think it's meant to pick this overload: https://github.com/NVIDIA/stdexec/blob/2a7b8c0ee8950101b95be2fc0931c79b42effb26/include/stdexec/execution.hpp#L3553-L3556, but no longer does because it can't deduce the template arguments (GCC says `note:   couldn’t deduce template parameter ‘_SenderId’`). I guess this is because of the ADL isolation wrapper which makes that a context in which it simply can't be deduced. I'm not quite sure if there's a simple fix to the ADL isolation wrapper or to the call site... The only thing I can think of is introducing e.g. some tag type into the `ensure_started` sender which can be used to identify it. Do you have a preferred way of fixing this?

msimberg:
Constrains the `ensure_started` overload supposed to take an existing `ensure_started` sender with a custom concept to allow deduction to happen. `__test_ensure_started_sender` is not an inline lambda because of missing support in earlier clang versions.

Ideally I'd like a test to ensure that this overload is really picked, and that a new shared state isn't actually created. Would something like a public `__get_shared_state` member function on the `ensure_started` `__sender` be fair game to add for testing purposes?

Fixes #715.

### DIFF
--- a/include/stdexec/execution.hpp
+++ b/include/stdexec/execution.hpp
@@ -3601,6 +3601,14 @@ namespace stdexec {
       using __dispatcher_for =
         __make_dispatcher<__cust_sigs, __mconstructor_for<__sender_t>, _Sender, _Env>;
 
+    template <class _SenderId, class _EnvId>
+    void __test_ensure_started_sender(__sender<_SenderId, _EnvId> const& __sndr2){};
+
+    template <class _Sender>
+    concept __ensure_started_sender = requires(typename _Sender::__id __sndr1) {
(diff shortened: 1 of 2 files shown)


ericniebler (approved):

ericniebler:
Thank you, @msimberg!

PR #746, opened 2023-01-24
Add test for sync-wait-type's definition in the spec

ericniebler:


ericniebler:
@ccotter This seems to work ok now. I think it's because my little metaprogramming library jumps through some hoops to sidestep the "pack expansion into non-variadic parameter" issue. Maybe the bug is in the spec, and the fix is to specify those hoops.

ccotter:
Ah, it looks like https://github.com/NVIDIA/stdexec/pull/665 fixed the test I wrote back in Sept. I'm not sure if there's anything wrong with the spec, but I'll take a closer look again.

ericniebler:
Right, `__meval` is the secret sauce that makes this work.

ericniebler:
I have fixed this in P2300 in this commit: brycelelbach/wg21_p2300_execution@4af029a04b4f0b6fd6973496792c018fe6dcf249.

PR #747, opened 2023-01-24
Fix use-after-move in test_split

ccotter:


ericniebler:
Nice catch.

ccotter:
clang-tidy didn't find this one, though I'm not sure why yet.

ccotter:
Ah, I think clang-tidy doesn't flag this as a use-after-move violation since `connect()` takes an lvalue reference to `sndr`.

PR #748, opened 2023-01-26
Implement R6 version of split

ccotter:


include/stdexec/execution.hpp
```diff
@@ -688,6 +691,9 @@ namespace stdexec {
 
   /////////////////////////////////////////////////////////////////////////////
   // [execution.senders]
+  template <class _T>
+    using attrs_of_t = decltype(get_attrs(__declval<_T>()));
```
ccotter, line 695: I'll keep the attrs stuff in place until the R7 changes are "finalized." attrs->env seems to be a relatively "uninteresting" change.

ericniebler: I think I'd rather make the changes in brycelelbach/wg21_p2300_execution#20 as soon as possible. I don't want folks to start writing to the `get_attrs` API, and I don't want to have to support it indefinitely.

ccotter: I can put up a attrs->env PR now if you'd like based on what is currently on that PR.

include/stdexec/execution.hpp
```diff
@@ -3297,9 +3325,8 @@ namespace stdexec {
     struct split_t {
       template <sender _Sender, class _Env = __empty_env>
           requires
-            (copy_constructible<remove_cvref_t<_Sender>> &&
-             sender_to<_Sender&, __receiver_t<_Sender, _Env>>) ||
-            __is_split_customized<_Sender, _Env>
+             (sender_in<_Sender, _Env> && constructible_from<decay_t<attrs_of_t<_Sender>>, attrs_of_t<_Sender>>) ||
```
ericniebler, line 3328: > `constructible_from<decay_t<attrs_of_t<_Sender>>, attrs_of_t<_Sender>>`

We should have a `__decay_copable` concept that simplifies this. It would look like:

```c++
template <class _Ty>
  concept __decay_copyable =
    constructible_from<decay_t<_Ty>, _Ty>;
```

... and it would live along side (and be subsumed by) `__nothrow_decay_copyable` in concepts.hpp.

include/stdexec/execution.hpp
ericniebler: Yes, along with the `enable_sender` / `enable_receiver` change. That change must be made together with the `get_attrs`->`get_env` change or else senders will start satisfying the `receiver` concept. :grimacing:

include/stdexec/execution.hpp
```diff
@@ -3273,8 +3301,8 @@ namespace stdexec {
     using _Env = __1;
     using __cust_sigs =
       __types<
-        tag_invoke_t(split_t, get_completion_scheduler_t<set_value_t>(get_attrs_t(_Sender&)), _Sender),
-        tag_invoke_t(split_t, get_completion_scheduler_t<set_value_t>(get_attrs_t(_Sender&)), _Sender, _Env),
+        tag_invoke_t(split_t, get_completion_scheduler_t<set_value_t>(get_attrs_t(const _Sender&)), _Sender),
+        tag_invoke_t(split_t, get_completion_scheduler_t<set_value_t>(get_attrs_t(const _Sender&)), _Sender, _Env),
```
ccotter, line 3305: question on the customizations taking env - it wasn't clear to me from the spec that split or ensure_started provide the env customized versions. can you help point that out. or are they non-standard?

ericniebler: They're non-standard. I intend to propose them, though. Any algorithm that must eagerly `connect` a sender before the final receiver is available to optionally accept an environment.

ericniebler:
@ccotter Heads up, I rebased this PR to fix the test failure with nvc++.

include/stdexec/execution.hpp
```diff
@@ -3233,14 +3248,23 @@ namespace stdexec {
...
+              // NOT TO SPEC:
+              // The spec says this should be copy_cvref, but the
+              // 'lvalue split move only sender' test fails if copy_cvref
+              // is used. E.g., in
+              //
+              //   auto snd = just(move_only_type{});
+              //   auto snd2 = split(snd);
+              //
+              // the lvalue 'snd2' does not satisfy the 'sender' concept since
+              // just(move_only_type{}) is move only.
```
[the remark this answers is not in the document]
ericniebler: Now that I think about it, I don't think this test case _should_ work. `snd2` should end up storing (a copy of) `snd` by value. If it's stored by reference, then the split sender will become invalid as soon as the input sender does. But if `snd` is an lvalue to a move-only type, there's no way for `snd2` to store a copy of it.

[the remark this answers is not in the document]
ccotter: Ah, you are right. I thought that was the simplified version of what I was actually having trouble with, which is getting the following to be valid (the "split into then" test in test_split.cpp)

```
ex::split(ex::just(move_only_type{0}))
    | ex::then([](const move_only_type&) { });
```

I'm not sure if the bug is in the make_completion_signatures alias or get_completion_signatures_t signature, but my current version of make_completion_signatures was *a way* for this expression to compile.

include/stdexec/concepts.hpp
```diff
@@ -198,6 +198,10 @@ namespace stdexec {
       constructible_from<_T, _As...> && std::is_nothrow_constructible_v<_T, _As...>;
 #endif
 
+  template <class _Ty>
+    concept __decay_copyable =
+      constructible_from<decay_t<_Ty>, _Ty>;
+
   template <class _Ty>
     concept __nothrow_decay_copyable =
```
ericniebler, line 206: For the sake of concept subsumption...

```suggestion
    concept __nothrow_decay_copyable =
      __decay_copyable<_Ty> &&
```

include/stdexec/execution.hpp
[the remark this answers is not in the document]
ericniebler: Actually, I take it all back. Don't listen to me. `snd2` doesn't need to store a copy of `snd` at all. It stores a `shared_ptr` to the `__sh_state`, in which the _operation state_ lives (and the sender's attrs). So that means that the type of the sender passed to `split` should be perfectly forwarded all the way into `__sh_state`, and _that_ type is used to compute the type of the inner operation state. If a `S&` is passed to `split`, then `__sh_state` should be parameterized on `S&`, and an `S&` is passed to `connect` in `__sh_state`'s c'tor. And that should work even if `S` is a move-only type.

Sorry for the confusion. LMK if I have now thoroughly confused you, and I can take a stab at it.

ericniebler (changes requested):
I keep getting this algorithm wrong. My apologies.

include/stdexec/concepts.hpp
ccotter: I tested my change with something like 

```
template <class T> requires __nothrow_decay_copyable<T>
int fn(T t) { return 0; } 
template <class T> requires __decay_copyable<T>
int fn(T t) { return 1; }
```

And confirmed `T=int` (which satisfies both concepts) selected the first `fn`. Perhaps my test was not completely valid to verify subsumption works?

ericniebler: Ah, sure. Because `__nothrow_constructible_from` subsumes `constructible_from`. Disregard.

include/stdexec/execution.hpp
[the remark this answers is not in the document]
ccotter: I don't think I'm thoroughly confused :) - except possibly by this:

> If a `S&` is passed to `split`, then `__sh_state` should be parameterized on `S&`, and an `S&` is passed to `connect` in `__sh_state`'s c'tor. And that should work even if `S` is a move-only type.

`split(s)` should not be a valid expression when `s=just(move_only_type{})` - correct? Only `split(move(s))` should be valid, in which case `__sh_state` is parameterized on `S`. Really, I'm interested in single-shot senders, which is the case for `s`.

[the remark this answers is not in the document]
ccotter: I *think* my latest commit has `split()` behaving as described in the spec after updating the `__sender_t` alias on line 3314 to no longer remove any references from the type . Now, if `S&` is passed to `split`, then the inner sender and shared state templates are parameterized on `S&`, and likewise for `S` being passed to `split`. The completion signatures still doesn't like `copy_cvref_t` however.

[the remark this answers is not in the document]
ccotter: Looking closer at the spec,

 > The name execution::split denotes a customization point object. For some subexpression s, let S be decltype((s)). If execution::sender_in<S, split-env> ... is false, execution::split is ill-formed. 

In `s=just(move_only_type{})`, `S` will be an lvalue reference, and `sender<S>` is not true since `s` cannot be copied. Is my understanding correct?

In split's completion signatures definition in terms of `make_completion_signatures<
  copy_cvref_t<S2, S>, ...` isn't `split` logically meant to decouple the input sender from the result (hence copy_cvref_t doesn't seem correct here)?

include/stdexec/execution.hpp
ccotter: > attrs->env seems to be a relatively "uninteresting" change.
 
 Turns out that was a wrong assessment :P

include/stdexec/execution.hpp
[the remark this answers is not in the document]
ericniebler: I think you're right.

include/stdexec/execution.hpp
```diff
@@ -3285,10 +3311,7 @@ namespace stdexec {
         __minvocable<__which<__cust_sigs>, _Sender, _Env>;
 
     template <class _Sender, class _Env>
-      using __sender_t = __t<__sender<stdexec::__id<remove_cvref_t<_Sender>>, stdexec::__id<remove_cvref_t<_Env>>>>;
-
-    template <class _Sender, class _Env>
-      using __receiver_t = __t<__receiver<stdexec::__id<remove_cvref_t<_Sender>>, stdexec::__id<remove_cvref_t<_Env>>>>;
+      using __sender_t = __t<__sender<stdexec::__id<_Sender>, stdexec::__id<remove_cvref_t<_Env>>>>;
```
ericniebler, line 3314: This isn't quite right. `_Sender` could be a reference type here. `__id` of a reference type isn't going to do anything useful. You'll need a bit of nastiness here to remove the cvref, apply `__id<>`, then slap the cvref onto the Id, and pass that as a template parameter to `__sender`.

Then within `__sender`, you'll have to undo all of that.

This dance gets done in a number of places. We should probably have some alias templates to make this simpler.

ccotter: Do we need all cvref, or can we remove cv from `_Sender`, then do the same dance but only for the ref-ness?

ericniebler (approved):

ericniebler:
@ccotter The other authors didn't like the `split` sender inheriting the input sender's environment. I've reverted that change and will merge as soon as tests finish.

Thanks for your work on this!

ccotter:
Thanks! Sounds good, I think that makes sense as well. Interestingly, the original desire for `split` to be able to forward sender queries was one of the motivations for introducing sender environments.

ccotter:
I can take a look at ensure_started next, although I wonder if it'll need a similar treatment with respect to the environment returned by the returned sender?

PR #749, opened 2023-01-26
Hack/tbb thread pool

BenFrantzDale:
@ericniebler this is a hack. Don't merge. But I think(?) the `tbbexec::tbb_thread_pool` class is about right.

Notes:
1. `fmt` isn't actually needed.
2. I'm sure the cmake is wonky. I hard-coded a mac version.
3. The test doesn't test anything/doesn't work.
4. I had crashes when I used the inline scheduler. Not sure why.

My goal here is to check over the implementation of `tbbexec::tbb_thread_pool` to see if it seems right. Then we can sort out the details.


include/tbbexec/tbb_thread_pool.hpp
```diff
@@ -0,0 +1,413 @@
...
+            template <typename Receiver>
+            operation<stdexec::__x<std::decay_t<Receiver>>> make_operation_(Receiver&& r) const {
+                return operation<stdexec::__x<std::decay_t<Receiver>>>{this->pool_, (Receiver &&) r};
+            }
+
+            template <class Receiver>
+            friend operation<stdexec::__x<std::decay_t<Receiver>>> tag_invoke(stdexec::connect_t,
+                                                                              sender s,
+                                                                              Receiver&& r) {
+                return s.make_operation_((Receiver &&) r); // std::move(r)?
```
BenFrantzDale, line 64: Are these moves or forwards? I think they are `std::forward<Receiver>(r)`?

include/tbbexec/tbb_thread_pool.hpp
```diff
@@ -0,0 +1,413 @@
...
+                const auto avg_per_thread = n / size;
+                const auto n_big_share = avg_per_thread + 1;
+                const auto big_shares = n % size;
+                const auto is_big_share = rank < big_shares;
+                const auto begin =
+                    is_big_share ? n_big_share * rank : n_big_share * big_shares + (rank - big_shares) * avg_per_thread;
+                const auto end = begin + (is_big_share ? n_big_share : avg_per_thread);
+
+                return std::make_pair(begin, end);
+            }
```
BenFrantzDale, line 128: This is very similar to the version in `static_thread_pool.hpp`. Ideally the logic would be shared...

include/tbbexec/tbb_thread_pool.hpp
```diff
@@ -0,0 +1,413 @@
...
+                return std::min(shape_, static_cast<Shape>(pool_.available_parallelism()));
+            }
+
+            template <class F>
+            void apply(F f) {
+                std::visit([&](auto& tupl) -> void { std::apply([&](auto&... args) -> void { f(args...); }, tupl); },
+                           data_);
+            }
+
+            bulk_shared_state(tbb_thread_pool& pool, Receiver receiver, Shape shape, Fun fn)
```
BenFrantzDale, line 140: This is basically all copy-paste from `static_thread_pool.hpp` :-/

include/tbbexec/tbb_thread_pool.hpp
```diff
@@ -0,0 +1,413 @@
...
+    }
+    void bulk_enqueue(task_base* task, std::uint32_t n_threads) noexcept {
+        for (std::size_t i = 0; i < n_threads; ++i) {
+            // FIXME: Is this right? i is the tid?
+            m_arena.enqueue(
+                [task, i] { tbb::this_task_arena::isolate([task, i] { task->__execute(task, /*tid=*/i); }); });
+        }
+    }
+
+    tbb::task_arena m_arena;
```
BenFrantzDale, line 387: This and the c'tor is basically all that's different from `static_thread_pool`.

include/tbbexec/tbb_thread_pool.hpp
```diff
@@ -0,0 +1,413 @@
...
+    scheduler get_scheduler() noexcept { return scheduler{*this}; }
+
+    void request_stop() noexcept { /* TODO? */
+    }
+
+    std::uint32_t available_parallelism() const { return m_arena.max_concurrency(); }
+
+private:
+    void enqueue(task_base* task) noexcept {
+        m_arena.enqueue([task] { tbb::this_task_arena::isolate([task] { task->__execute(task, /*tid=*/0); }); });
```
BenFrantzDale, line 377: I added `tbb::this_task_arena::isolate` to prevent [moonlighting](https://link.springer.com/chapter/10.1007/978-1-4842-4398-5_12). Not sure if it's necessary, but I feel like in the context of P2300, there's no benefit of moonlighting and plenty of footguns.

test/CMakeLists.txt
```diff
@@ -70,12 +70,13 @@ set(stdexec_test_sources
...
     exec/async_scope/test_empty.cpp
     exec/async_scope/test_stop.cpp
+    tbbexec/test_tbb_thread_pool.cpp
     )
 
 add_executable(test.stdexec ${stdexec_test_sources})
 
 target_include_directories(test.stdexec PRIVATE ${CMAKE_CURRENT_LIST_DIR})
-target_link_libraries(test.stdexec PUBLIC STDEXEC::stdexec stdexec_executable_flags Catch2::Catch2)
+target_link_libraries(test.stdexec PUBLIC STDEXEC::stdexec stdexec_executable_flags Catch2::Catch2 TBB::tbb fmt::fmt)
```
BenFrantzDale, line 79: How do you want to deal with the tbb dependency? I assume you don't want to depend on it entirely. I'm not a CMake expert, though...

include/tbbexec/tbb_thread_pool.hpp
BenFrantzDale: I don't quite understand what this stuff is trying to do enough to make a suggestion, but it feels like maybe a base class would cover it. So instead of
```c++
template <class SenderId, class ReceiverId, class Shape, class Fun, bool MayThrow>
struct bulk_shared_state : exec::task_base {
    ...
};
```
we'd have
```c++
template <class SenderId, class ReceiverId, class Shape, class Fun, bool MayThrow>
struct bulk_shared_state : exec::task_base_impl<tbb_thread_pool, SenderId, ReceiverId, Shape, Fun, MayThrow> {};
```
which would provide all the goodies.

include/tbbexec/tbb_thread_pool.hpp
msimberg: Depends on the context :upside_down_face: But in this case it's a forward.

test/CMakeLists.txt
msimberg: I've stared at enough CMake and happy to help here if needed. It's relatively easy to either put it behind an explicit flag (generally my preference) or to only attempt to look for TBB and disable support if it's not found (instead of erroring out).

include/tbbexec/tbb_thread_pool.hpp
ericniebler: Recently I've started taking the receiver by value in the `connect` `tag_invoke` overload. Most receivers are just a pointer to the operation state so copies are cheap, and it simplifies the code.

And yes, in the interest of compile times, I tend to avoid `std::move` and `std::forward`. The fewer templates instantiated, the better.

include/tbbexec/tbb_thread_pool.hpp
ericniebler: I see where you're coming from, but I would leave it for now until I have a better idea where else we would use such a thing. I'm wary of putting things in the public API until they are more well thought out.

include/tbbexec/tbb_thread_pool.hpp
BenFrantzDale: I'm a big fan of
```c++
//! Convenience macro in place of std::forward
//! See https://www.fluentcpp.com/2019/05/14/3-types-of-macros-that-improve-c-code/
//! and https://www.foonathan.net/2020/09/move-forward/
#define FWD(...) static_cast<decltype(__VA_ARGS__)&&>(__VA_ARGS__)
```
which is included basically everywhere in our codebase (next to our `operator""_uz` and `operator""_z` :-) )

include/tbbexec/tbb_thread_pool.hpp
BenFrantzDale: I'm happy to leave it for now. (I also could put that in `tbbexec::detail` just so it's factored out.) It does feel like if others are going to write other thread-pool wrappers, it would be more ergonomic for `::exec` to provide some reasonable defaults.

test/CMakeLists.txt
BenFrantzDale: @msimberg lmk what you'd like to see and how to spell it for cmake.

I plan to drop `fmt` before this is merged (I guess it's C++20 so I can just use `<format>` 🤦. I'm still at 17 for day-to-day work for a few more months.)

msimberg: @BenFrantzDale I think you might actually be fine mostly duplicating what's been done for CUDA support: https://github.com/NVIDIA/stdexec/blob/41ec3de5bb9bfb24ac18e3d736f2aaa2fd65d533/CMakeLists.txt#L193-L251. Replacing `CUDA` with with `TBB` and `nvexec` with `tbbexec` of course. However, I don't think you need any of the `target_compile_features`/`target_compile_options` calls. The stuff you want to keep is setting up the target: https://github.com/NVIDIA/stdexec/blob/41ec3de5bb9bfb24ac18e3d736f2aaa2fd65d533/CMakeLists.txt#L205-L208, linking to `stdexec`: https://github.com/NVIDIA/stdexec/blob/41ec3de5bb9bfb24ac18e3d736f2aaa2fd65d533/CMakeLists.txt#L211, and finding and linking to the `tbb` target: https://github.com/NVIDIA/stdexec/blob/41ec3de5bb9bfb24ac18e3d736f2aaa2fd65d533/CMakeLists.txt#L235-L242.

Finally, for the test I'm not sure if @ericniebler would prefer a separate test executable (as for CUDA) or if it's fine to have it in the main executable? If the former you'll want to look at https://github.com/NVIDIA/stdexec/blob/main/test/nvexec/CMakeLists.txt. If the latter you'll need to conditionally include the test file and targets with `$<$<BOOL:STDEXEC_ENABLE_TBB>:tbbexec/test_tbb_thread_pool.cpp>` and `$<$<BOOL:STDEXEC_ENABLE_TBB>:tbbexec>`, respectively.

Let me know if any of this doesn't make sense.

ericniebler: > Finally, for the test I'm not sure if @ericniebler would prefer a separate test executable (as for CUDA) or if it's fine to have it in the main executable?

Probably a separate executable would be cleanest. I would actually love if _all_ the tests were in their own executable, but I haven't had time to make that change.

ericniebler:
Apologies for not being more on top of this. I've been OOTO on emergency family business.

CMakeLists.txt
```diff
@@ -247,16 +248,47 @@ if(STDEXEC_ENABLE_CUDA)
...
-endif()
+    endif ()
+endif ()
+
+
+#////////BFD
+option(STDEXEC_ENABLE_TBB "Enable TBB targets" OFF)
+
+if (STDEXEC_ENABLE_TBB)
+    find_package(TBB REQUIRED)
```
msimberg, line 259: Does replacing this with
```
include(rapids-find)
rapids_find_package(
    TBB REQUIRED
    BUILD_EXPORT_SET stdexec-exports
    INSTALL_EXPORT_SET stdexec-exports
)
```
not work?

test/CMakeLists.txt
```diff
@@ -70,12 +70,19 @@ set(stdexec_test_sources
...
 
 add_executable(test.stdexec ${stdexec_test_sources})
 
 target_include_directories(test.stdexec PRIVATE ${CMAKE_CURRENT_LIST_DIR})
-target_link_libraries(test.stdexec PUBLIC STDEXEC::stdexec stdexec_executable_flags Catch2::Catch2)
+target_link_libraries(test.stdexec
+    PUBLIC
+    STDEXEC::stdexec
+    $<$<BOOL:STDEXEC_ENABLE_TBB>:tbbexec>
+    $<$<BOOL:STDEXEC_ENABLE_TBB>:TBB::tbb>
```
msimberg, line 83: This should be enough (`STDEXEC::` only for consistency with the `STDEXEC::stdexec` dependency above):
```suggestion
    $<$<BOOL:STDEXEC_ENABLE_TBB>:STDEXEC::tbbexec>
```

ericniebler:
Hi @BenFrantzDale. I fixed up the cmake errors. Are you finished with this PR? I would like at least one test running that sets up TBB and runs the TBB tests. Is that something you can do?

BenFrantzDale:
> Hi @BenFrantzDale. I fixed up the cmake errors. Are you finished with this PR? I would like at least one test runner that sets up TBB and runs the TBB tests. Is that something you can do?

I can do that. Give me a few days.

test/tbbexec/test_tbb_thread_pool.cpp
```diff
@@ -0,0 +1,246 @@
...
+  stdexec::sync_wait(std::move(snd));
+  // the work should be executed
+  REQUIRE(called);
+}
+
+namespace {
+
+int compute(int x) { return x + 1; }
+
+std::size_t legible_thread_id() {
```
mhoemmen, line 59: I generally just print the std::hash of the thread ID.  That should save you from needing a mutex here.

include/tbbexec/tbb_thread_pool.hpp
```diff
@@ -0,0 +1,413 @@
...
+    }
+    void bulk_enqueue(task_base* task, std::uint32_t n_threads) noexcept {
+        for (std::size_t i = 0; i < n_threads; ++i) {
+            // FIXME: Is this right? i is the tid?
+            m_arena.enqueue(
+                [task, i] { tbb::this_task_arena::isolate([task, i] { task->__execute(task, /*tid=*/i); }); });
```
mhoemmen, line 383: My hope as a user who cares about performance, is that the implementation wouldn't need to worry about work stealing (the original Cilk term, which I prefer to the less descriptive "moonlighting").  Work stealing is essential to performance of not-necessarily-load-balanced parallel loops.

Does C++17 `parallel_for` need to worry about users doing funny things with `thread_local`?  If not, then I would hope that the P2300 wording would let `bulk` forbid that kind of user behavior.  Otherwise, then perhaps that's a change to consider for P2300 itself.

include/tbbexec/tbb_thread_pool.hpp
mhoemmen: Please see comment on `bulk_enqueue` below.  Does P2300 have things to say about ungood things like messing with `thread_local`, for work not in `bulk`?

include/tbbexec/tbb_thread_pool.hpp
```diff
@@ -0,0 +1,413 @@
...
+                                                      const scheduler& sch,
+                                                      S&& sndr,
+                                                      Shape shape,
+                                                      Fn fun) noexcept {
+            return bulk_sender_t<S, Shape, Fn>{*sch.pool_, (S &&) sndr, shape, (Fn &&) fun};
+        }
+
+        friend stdexec::forward_progress_guarantee tag_invoke(stdexec::get_forward_progress_guarantee_t,
+                                                              const tbb_thread_pool&) noexcept {
+            return stdexec::forward_progress_guarantee::parallel;
```
mhoemmen, line 359: This looks right.

https://spec.oneapi.io/versions/1.0-rev-3/elements/oneTBB/source/task_scheduler/task_arena/task_arena_cls.html

> The task is scheduled for eventual execution by a worker thread even if no thread ever explicitly waits for the task to complete.

include/tbbexec/tbb_thread_pool.hpp
mhoemmen: @ericniebler says that we shouldn't need `isolate` here.

include/tbbexec/tbb_thread_pool.hpp
mhoemmen: Would you also consider removing `isolate` here?

mhoemmen (commented):
@ericniebler asked me to review this based on my (somewhat outdated) experience with TBB.  It would be excellent if we could consider removing those `isolate` calls, especially in `bulk`, as work stealing is essential for good performance of parallel loops.

mhoemmen:
@ericniebler doesn't think that we need `isolate` in either case.  I would recommend against it.

include/tbbexec/tbb_thread_pool.hpp
BenFrantzDale: Sure thing. The [one library feature I wrote for TBB](https://github.com/oneapi-src/oneTBB/issues/267) got me paranoid, but that was `tbb::collabrative_call_once` which has the potential to deadlock without isolation.

What should go in the `tid` field of `__execute`? It's just
```c++
  struct task_base {
    task_base* next;
    void (*__execute)(task_base*, std::uint32_t tid) noexcept;
  };
```
without comment.

BenFrantzDale (commented):
OK, I cleaned it up and it seems to be working.

I really don't like how big `tbbexec/tbb_thread_pool.hpp` is for what boils down to a few c'tor and one `m_arena.enqueue` call. I left a few other questions.

include/tbbexec/tbb_thread_pool.hpp
mhoemmen: @BenFrantzDale Thanks for asking!  The last time I used TBB seriously was in 2010, so I'll have to look carefully at the documentation to answer this.  (Eric pulled me in to review because I was the only one available on the spot who knew _anything_ about TBB ; - ) .)  I did have to subclass `task`, but that was already considered expert-only back then.  I can only guess that "tid" suggests the native thread id.

include/tbbexec/tbb_thread_pool.hpp
```diff
@@ -0,0 +1,435 @@
...
+//! This is a P2300-style thread pool wrapping tbb::task_arena, which its docs describe as "A class that represents an
+//! explicit, user-managed task scheduler arena."
+//! Once set up, a tbb::task_arena has
+//! * template<F> void enqueue(F &&f)
+//! and
+//! * template<F> auto execute(F &&f) -> decltype(f())
+//!
+//! See https://spec.oneapi.io/versions/1.0-rev-3/elements/oneTBB/source/task_scheduler/task_arena/task_arena_cls.html
+namespace detail {
+template <typename DerivedPoolType> // CRTP
```
BenFrantzDale, line 24: I went ahead and pulled out a base class so I could separate concerns...

include/tbbexec/tbb_thread_pool.hpp
```diff
@@ -0,0 +1,435 @@
...
+
+    template <typename PoolType, typename ReceiverId>
+    friend class operation;
+
+    void enqueue(task_base* task, std::uint32_t tid = 0) noexcept {
+        arena_.enqueue([task, tid] { task->__execute(task, /*tid=*/tid); });
+    }
+
+    tbb::task_arena arena_{tbb::task_arena::attach{}};
+};
```
BenFrantzDale, line 408: ... Now with `detail::thread_pool_base` pulled out it's clear how little actually depends on `tbb::task_arena`: There's a constructor, `available_parallelism()`, `void enqueue(task_base* task, std::uint32_t tid = 0)`, and the member.

BenFrantzDale (commented):
Comments about pulling out a base class.

include/tbbexec/tbb_thread_pool.hpp
BenFrantzDale: If it's conservative, which would the less-conservative value be?

include/tbbexec/tbb_thread_pool.hpp
```diff
@@ -0,0 +1,416 @@
+#pragma once
+
```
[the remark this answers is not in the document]
ericniebler: This file needs a copy of the license, as other files in the repo. Since you copied the `static_thread_pool` code and hacked it, you should have two copyright lines: one for NVIDIA and one for yourself.

ericniebler:
Need a hand setting up a GitHub Action to test this in CI?

BenFrantzDale:
> Need a hand setting up a GitHub Action to test this in CI?

I don’t know how to do that, so yes please. I know you are a busy guy; if it’s much work, if you can point me in the right direction I’m sure I could do it too.

BenFrantzDale:
> > Need a hand setting up a GitHub Action to test this in CI?
> 
> I don’t know how to do that, so yes please. I know you are a busy guy; if it’s much work, if you can point me in the right direction I’m sure I could do it too.

@ericniebler Point me at what you are looking for and I'll see if I can make it happen.

include/tbbexec/tbb_thread_pool.hpp
```diff
@@ -0,0 +1,455 @@
...
+template <typename PoolType, typename ReceiverId>
+class operation : task_base {
+    using Receiver = stdexec::__t<ReceiverId>;
+    friend class detail::thread_pool_base<PoolType>;
+
+    PoolType& pool_;
+    Receiver receiver_;
+
+    explicit operation(PoolType& pool, Receiver&& r) : pool_(pool), receiver_(std::move(r)) {
+        this->__execute = [](task_base* t, std::uint32_t /* tid What is this needed for? */) noexcept {
```
BenFrantzDale, line 440: @ericniebler Can you comment on the `tid` arg that I'm ignoring here?

ericniebler: `tid` is the thread id. It is used in the implementation of `bulk` to assign a chunk of the input to each thread in the pool. It's safe to ignore here.

ericniebler:
> > > Need a hand setting up a GitHub Action to test this in CI?
>
> @ericniebler Point me at what you are looking for and I'll see if I can make it happen.

It's a little tricky right now because of how we have our CPU CI set up right now. To test this in CI, I need to update the cmake configure command to add `-DSTDEXEC_ENABLE_TBB:BOOL=TRUE`, but the configure command is burned into the container image. :-/

@trxcllnt is going to update the images so we have a bit more control over the CI configuration.

ericniebler:
@BenFrantzDale I got the TBB tests running in CI. It looks like one of the tests fails:

```
[ctest] -------------------------------------------------------------------------------
[ctest] exec::on works when changing threads with tbbexec::tbb_thread_pool
[ctest] -------------------------------------------------------------------------------
[ctest] /workspaces/stdexec/test/tbbexec/test_tbb_thread_pool.cpp:80
[ctest] ...............................................................................
[ctest] 
[ctest] /workspaces/stdexec/test/tbbexec/test_tbb_thread_pool.cpp:83: FAILED:
[ctest]   CHECK( stdexec::get_forward_progress_guarantee(pool) == stdexec::forward_progress_guarantee::parallel )
[ctest] with expansion:
[ctest]   2 == 1
[ctest] 
```

Thoughts?

BenFrantzDale:
Exciting! Thanks @ericniebler. This will let me use Senders within my current everything-assumes-tbb code! In particular, I find myself writing parallel algorithms that depend on TBB; now I should be able to write them in terms of Senders and so make them far more reusable. For example, I have an algorithm like this (not the most elegant one, but easiest to explain):
```c++
//! Just like std::ranges::to<std::vector>(FWD(inRange))
//! for when if the range's size is known in O(1) and
//! the value_type is default-constructible.
template <typename InputRange>
[[nodiscard]] auto parallelRangeToVector(InputRange&& inRange) {
    auto result = decltype(std::ranges::to<std::vector>(FWD(inRange)))(std::size(inRange));
    // Assign into result in parallel
    return result;
}
```
which is great for `std::views::transform` pipelines where the function is expensive to evaluate. As in
```c++
auto vec = parallelRangeToVector(items | std::views::transform(expensiveFn));
```

PR #750, opened 2023-01-27
Initial R7 env changes

ccotter:
This does not enable the trait on receivers or senders (ones returned from algos, or the ones defined in the tests or examples). The exception is stdexec::just, which now exposes `using is_sender = void` to prove out a new test.

One significant-ish change I'll need feedback on is the refactor of `get_env`, which provides forwards and backwards compatibility for R5 and R7 sender types. I also introduced a facility to emit deprecation diagnostics for sender types that are not R7 compatible, along with hints about what to change ("add get_env" or "satisfy enable_sender"). I think we'll want a compile time enablement switch/macro for this, but I'm mostly looking for feedback on whether we should introduce a deprecation notice via this strategy of guessing whether a type is R5 vs R7 with "close enough" concept checks.

The [sed -i -r get-attrs -> get_env](https://github.com/NVIDIA/stdexec/commit/af41a488731cac45ac4dc95cb082ad7b1fa4237e) commit is find/replace, and most interesting changes are in execution.hpp anyway.

### DIFF
--- a/include/exec/inline_scheduler.hpp
+++ b/include/exec/inline_scheduler.hpp
@@ -47,14 +47,14 @@ namespace exec {
           return {(R&&) rec};
         }
 
-      struct __attrs {
+      struct __env {
         friend inline_scheduler
-        tag_invoke(stdexec::get_completion_scheduler_t<stdexec::set_value_t>, const __attrs&) noexcept {
+        tag_invoke(stdexec::get_completion_scheduler_t<stdexec::set_value_t>, const __env&) noexcept {
           return {};
         }
       };
 
-      friend __attrs tag_invoke(stdexec::get_env_t, const __sender&) noexcept {
+      friend __env tag_invoke(stdexec::get_env_t, const __sender&) noexcept {
         return {};
       }
     };
--- a/include/exec/inline_scheduler.hpp
+++ b/include/exec/inline_scheduler.hpp
@@ -54,7 +54,7 @@ namespace exec {
         }
       };
 
-      friend __attrs tag_invoke(stdexec::get_attrs_t, const __sender&) noexcept {
+      friend __attrs tag_invoke(stdexec::get_env_t, const __sender&) noexcept {
         return {};
       }
     };
--- a/include/stdexec/execution.hpp
+++ b/include/stdexec/execution.hpp
@@ -626,6 +626,12 @@ namespace stdexec {
   /////////////////////////////////////////////////////////////////////////////
   // [execution.receivers]
   template <class _Receiver>
+    inline constexpr bool enable_receiver =
+      requires {
+        typename _Receiver::is_receiver;
+      };
+
+  template <class _Receiver>
     concept receiver =
       environment_provider<__cref_t<_Receiver>> &&
       move_constructible<remove_cvref_t<_Receiver>> &&
--- a/include/stdexec/execution.hpp
+++ b/include/stdexec/execution.hpp
@@ -342,7 +342,7 @@ namespace stdexec {
             (!tag_invocable<get_env_t, const _EnvProvider&>)
         constexpr decltype(auto) operator()(const _EnvProvider& __with_env) const
           noexcept {
-          if constexpr (__r5_sender<_EnvProvider>) {
+          if constexpr (!enable_sender<_EnvProvider>) {
             return __with_env;
           } else {
             return __empty_env{};
--- a/include/stdexec/execution.hpp
+++ b/include/stdexec/execution.hpp
@@ -318,8 +318,7 @@ namespace stdexec {
             (!tag_invocable<get_env_t, const _EnvProvider&>) &&
             // NOT TO SPEC: Remove the R5/R7 sender checks when
             // deprecating R5 support.
-            (!__r5_sender<_EnvProvider>) &&
-            __r7_sender<_EnvProvider>
+            (!__r5_sender<_EnvProvider>)
         constexpr auto operator()(const _EnvProvider& __with_env) const
           noexcept -> __empty_env {
           return {};
@@ -2095,6 +2094,8 @@ namespace stdexec {
         struct __t {
           using __id = __basic_sender;
           using completion_signatures = __completion_signatures_<_Tag, _Ts...>;
+          using is_sender = void;
+
           std::tuple<_Ts...> __vals_;
 
           template <receiver_of<completion_signatures> _Receiver>
--- a/include/stdexec/execution.hpp
+++ b/include/stdexec/execution.hpp
@@ -4191,18 +4191,18 @@ namespace stdexec {
             return {&__loop_->__head_, __loop_, (_Receiver &&) __rcvr};
           }
 
-          struct __attrs {
+          struct __env {
             run_loop* __loop_;
 
             template <class _CPO>
             friend __scheduler
-            tag_invoke(get_completion_scheduler_t<_CPO>, const __attrs& __self) noexcept {
+            tag_invoke(get_completion_scheduler_t<_CPO>, const __env& __self) noexcept {
               return __scheduler{__self.__loop_};
             }
           };
 
-          friend __attrs tag_invoke(get_env_t, const __schedule_task& __self) noexcept {
-            return __attrs{__self.__loop_};
+          friend __env tag_invoke(get_env_t, const __schedule_task& __self) noexcept {
+            return __env{__self.__loop_};
           }
 
           explicit __schedule_task(run_loop* __loop) noexcept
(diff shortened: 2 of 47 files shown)


include/stdexec/execution.hpp
```diff
@@ -243,8 +294,35 @@ namespace stdexec {
           noexcept(nothrow_tag_invocable<get_env_t, const _EnvProvider&>)
           -> tag_invoke_result_t<get_env_t, const _EnvProvider&> {
           static_assert(queryable<tag_invoke_result_t<get_env_t, const _EnvProvider&> >);
+          __check_sender_version<remove_cvref_t<_EnvProvider>>();
```
ericniebler, line 297: I'm trying to figure how this doesn't give sender warnings on receivers passed to `get_env`. Any hints?

include/stdexec/execution.hpp
```diff
@@ -243,8 +294,35 @@ namespace stdexec {
...
+      template <class _EnvProvider>
+          requires
+            (!tag_invocable<get_env_t, const _EnvProvider&>) &&
+            // NOT TO SPEC: Remove the R5/R7 sender checks when
+            // deprecating R5 support.
+            (!__r5_sender<_EnvProvider>)
+        constexpr auto operator()(const _EnvProvider& __with_env) const
+          noexcept -> __empty_env {
+          return {};
+        }
```
ericniebler, line 325: I _think_ these two overloads can be combined into one that basically does:

```c++
if constexpr (__r5_sender<_EnvProvider>)
  return __with_env;
else
  return __empty_env{};
```
... provided you make my suggested change to the `__r5_sender` concept above.

include/stdexec/execution.hpp
```diff
@@ -235,6 +245,47 @@ namespace stdexec {
...
+      //
+      // All R5 receivers already provide get_env, so the extra overloads
+      // of get_env would at most conflict if a receiver type also
+      // has completion signatures or the new enable_sender trait, neither
+      // of which is too likely.
+
+      template <class _Sender>
+      concept __r5_sender = requires(_Sender&& __sender) {
+        __declval<get_completion_signatures_t>()((_Sender&&)__sender);
+      };
```
ericniebler, line 263: I think you should add `&& !__r7_sender<_Sender>`.

include/stdexec/execution.hpp
```diff
@@ -637,64 +721,16 @@ namespace stdexec {
...
       // NOT TO SPEC
       // The sender related concepts are temporarily "in flight" being
-      // upgraded from P2300R5 to the get_attrs aware version of P2300.
+      // upgraded from P2300R5 to the get_env / enable_sender aware version
+      // in P2300R7.
       requires (const remove_cvref_t<_Sender>& __sndr) {
-        { get_attrs(__sndr) } -> queryable;
+        { get_env(__sndr) } -> queryable;
       } &&
       __sender<_Sender, no_env> &&
```
ericniebler, line 735: I'd like to drop this requirement checking with `no_env`. Basically, we want to know if something is an R5 sender _or_ an R7 sender. I think we can express that directly.

ericniebler (changes requested):
Some thoughts about how best to do the `get_env` CPO and the `sender` concept.

include/stdexec/execution.hpp
ccotter: Ah, you're right, receiver types do trigger this. I'd need to add another half baked concept to sense for a completion signature operation on the type.

ccotter: I moved the deprecation logic inside of the connect CPO, invoking separate checks on each of the deduced Sender and Receiver types.

include/stdexec/execution.hpp
ccotter: This was interesting - if I use trailing return type of `__if_c<__r5_sender<_EnvProvider>, const _EnvProvider&, __empty_env>`, clang gave

```
error: satisfaction of constraint 'stdexec::__valid<exec::__task::awaiter_context_t, __promise>' depends on itself
```

GCC 12 permitted the conditional return type. I've left the return type as `decltype(auto)` for now.

include/stdexec/execution.hpp
ccotter: I haven't addressed this yet

test/stdexec/concepts/test_concepts_sender.cpp
```diff
@@ -144,6 +144,18 @@ TEST_CASE("check completion signatures for sender that also supports error codes
...
+    ex::completion_signatures<             //
+      ex::set_value_t(),                   //
+      ex::set_error_t(std::exception_ptr), //
+      ex::set_stopped_t()>;
+
+  friend oper tag_invoke(ex::connect_t, my_r5_sender0, empty_recv::recv0&& r) { return {}; }
+};
+TEST_CASE("r5 sender emits deprecated diagnostics", "[concepts][sender]") {
+  ex::get_env(my_r5_sender0{});
+}
```
ericniebler, line 158: `my_r5_sender0` isn't an R5 sender because it doesn't support `get_completion_signatures(sndr, no_env{})`.

ccotter: Hm, why not? It satisfies the `sender` concept (which as of my branch is only changed from R5 by the additional requirement of `get_env(sndr)` being valid). The `completion_signatures` member type should handle any env, including `no_env.`

include/stdexec/execution.hpp
```diff
@@ -547,8 +650,24 @@ namespace stdexec {
 
   /////////////////////////////////////////////////////////////////////////////
   // [execution.receivers]
+
   template <class _Receiver>
     concept receiver =
+      // NOT TO SPEC:
+      // As we upgrade the receiver related entities from R5 to R7,
```
ccotter, line 657: Without additionally constraining the `receiver` concept to keep requiring receiver types provide a `get_env`, types like `no_env` became receivers. With CWG 2369, GCC-11 ended up instantiating more non-immediate context templates, and, for example, would try instantiating `when_all`'s `tag_invoke(connect_t, ...)`  with `_Receiver=no_env`. Clang and GCC-10 aren't bothered by this because (I'm guessing) CWG 2369 is not implemented there.

I don't have an explanation for why line 670 was needed for the code to compile with GCC-11 (it's also not needed for Clang) - it doesn't seem right, but I need to look further at that one.

ccotter: I think line 670 would go away once https://github.com/NVIDIA/stdexec/pull/752 is fixed

include/stdexec/execution.hpp
ericniebler: These sorts of errors can be tricky to track down, but it's usually worth it because it generally indicates that something is fishy somewhere. I'll try to have a look.

test/stdexec/concepts/test_concepts_sender.cpp
ericniebler: Yeah, don't listen to my. My brain is addled.

include/exec/inline_scheduler.hpp
```diff
@@ -54,7 +54,7 @@ namespace exec {
         }
       };
 
-      friend __attrs tag_invoke(stdexec::get_attrs_t, const __sender&) noexcept {
+      friend __attrs tag_invoke(stdexec::get_env_t, const __sender&) noexcept {
```
ericniebler, line 57: This is picky and unimportant, but the type being returned here should no longer be called `__attrs`. I don't care if you change it though.

include/stdexec/execution.hpp
ccotter: I can have another go at this one and let you know if I'm hopeless stuck. I had a good lesson over the past couple days on the meaning of immediate vs non-immediate context regarding SFINAE.

ericniebler: Maybe rebase on `main` and try again? I've fixed some issues that likely were causing more to be instantiated than necessary.

ccotter: Same result.

ccotter: I think what I have is problematic. When https://github.com/NVIDIA/stdexec/blob/main/include/exec/task.hpp#L313 evaluates the constraint, it ends up checking `env_of_t`, which (with my changes) checks if `get_env`'s _EnvProvider is a sender, which checks if the type is awaitable, and we're back at line 313 evaluating the same constraint.

`get_env` is about as low level as can be, so checking if `_EnvProvider` is a sender risks cycles like this. Suggestions on how to proceed? My original approach of having this be two separate overloads "worked," but I'm not sure if that's even guaranteed given the apparent cycle.

include/exec/inline_scheduler.hpp
ccotter: Thanks for the reminder. There are a number of outstanding types that retain the 'attrs' name that I haven't renamed yet. I'll fix this once I get the more "challenging" open items.

ericniebler (approved):

PR #752, opened 2023-01-29
Add failing test for __env_promise compile failure

ccotter:
This compiles in Clang, but not GCC 11.

### DIFF
--- a/test/stdexec/concepts/test_awaitables.cpp
+++ b/test/stdexec/concepts/test_awaitables.cpp
@@ -17,6 +17,7 @@
 #include <catch2/catch.hpp>
 #include <stdexec/execution.hpp>
 
+#include <exec/static_thread_pool.hpp>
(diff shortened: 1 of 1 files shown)


ccotter:
I think this would be fixed if `ex::__env_promise<ex::__empty_env>` did not satisfy `receiver`, which is true in R7 world. I'm not sure if there's another constraint/short circuit missing somewhere else though.

ericniebler:
> I think this would be fixed if `ex::__env_promise<ex::__empty_env>` did not satisfy `receiver`, which is true in R7 world.

Yup. I think the solution is to start moving in the direction of R7, with an explicit opt-in to the `receiver` concept. Then `__env_promise` can unequivocally state that is is _not_ a receiver.

ericniebler:
#753 fixes a bunch of problems this PR exposed. Thanks! I've cherry-picked your commit into that branch and will close this PR.

PR #755, opened 2023-01-31
Fix ADL isolation for cvref forwarding

ccotter:
These were instances of where `__id` could be passed a reference

### DIFF
--- a/include/nvexec/stream/common.cuh
+++ b/include/nvexec/stream/common.cuh
@@ -578,7 +578,7 @@ namespace nvexec {
           operation_state_base_t<stdexec::__id<OuterReceiver>>&>;
 
     template <class Sender, class InnerReceiver, class OuterReceiver>
-      using stream_op_state_t = operation_state_t<stdexec::__cvref_id<Sender, Sender>,
+      using stream_op_state_t = operation_state_t<stdexec::__cvref_id<Sender, std::remove_cvref_t<Sender>>,
                                                   stdexec::__id<InnerReceiver>,
                                                   stdexec::__id<OuterReceiver>>;
 
--- a/include/nvexec/stream/common.cuh
+++ b/include/nvexec/stream/common.cuh
@@ -538,7 +538,7 @@ namespace nvexec {
         requires stream_receiver<OuterReceiver>
       using exit_operation_state_t 
         = operation_state_t<
-            stdexec::__cvref_id<Sender, Sender>, 
+            stdexec::__cvref_id<Sender, std::remove_cvref_t<Sender>>,
             stdexec::__id<stdexec::__t<propagate_receiver_t<stdexec::__id<OuterReceiver>>>>, 
             stdexec::__id<OuterReceiver>>;
 
--- a/include/nvexec/stream/common.cuh
+++ b/include/nvexec/stream/common.cuh
@@ -446,11 +446,11 @@ namespace nvexec {
         };
       };
 
-    template <class SenderId, class InnerReceiverId, class OuterReceiverId>
+    template <class CvrefSenderId, class InnerReceiverId, class OuterReceiverId>
       struct operation_state_ {
         struct __t : operation_state_base_t<OuterReceiverId> {
           using __id = operation_state_;
-          using sender_t = stdexec::__t<SenderId>;
+          using sender_t = stdexec::__cvref_t<CvrefSenderId>;
           using inner_receiver_t = stdexec::__t<InnerReceiverId>;
           using outer_receiver_t = stdexec::__t<OuterReceiverId>;
           using typename operation_state_base_t<OuterReceiverId>::env_t;
@@ -531,14 +531,14 @@ namespace nvexec {
         };
       };
 
-    template <class SenderId, class InnerReceiverId, class OuterReceiverId>
-      using operation_state_t = stdexec::__t<operation_state_<SenderId, InnerReceiverId, OuterReceiverId>>;
+    template <class CvrefSenderId, class InnerReceiverId, class OuterReceiverId>
+      using operation_state_t = stdexec::__t<operation_state_<CvrefSenderId, InnerReceiverId, OuterReceiverId>>;
 
     template <class Sender, class OuterReceiver>
         requires stream_receiver<OuterReceiver>
       using exit_operation_state_t 
         = operation_state_t<
-            stdexec::__id<Sender>, 
+            stdexec::__cvref_id<Sender, Sender>, 
             stdexec::__id<stdexec::__t<propagate_receiver_t<stdexec::__id<OuterReceiver>>>>, 
             stdexec::__id<OuterReceiver>>;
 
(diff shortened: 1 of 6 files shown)


include/stdexec/execution.hpp
```diff
@@ -3811,14 +3811,14 @@ namespace stdexec {
...
-                stdexec::__id<__copy_cvref_t<_Self, _Sender>>,
+                stdexec::__cvref_id<_Self, _Sender>,
                 stdexec::__id<_Receiver>,
                 _Fun,
                 _Set>>;
           template <class _Self, class _Receiver>
             using __receiver_t =
               __receiver<
-                stdexec::__id<__copy_cvref_t<_Self, _Sender>>,
+                stdexec::__cvref_id<_Self, _Sender>,
```
ccotter, line 3821: this isn't quite the same thing, since it won't preserve the ref-ness of `_Sender`, whereas before at least if `_Sender` were a ref and `_Self` were not, `__copy_cv_ref` would keep the ref (as far as I understand)

ccotter: This might not matter as the parent `__sender` type ends up `decay_t`-ing the `_Sender` already.

ericniebler: Right. At this point, `_Sender` is not cvref-qualified.

include/stdexec/__detail/__meta.hpp
```diff
@@ -490,6 +490,12 @@ namespace stdexec {
   template <class _Ty>
     using __id = __minvoke<__id_<__has_id<_Ty>>, _Ty>;
 
+  template <class _Ty>
+    using __cvref_t = __copy_cvref_t<_Ty, __t<std::remove_cvref_t<_Ty>>>;
+
+  template <class _From, class _To>
+    using __cvref_id = __copy_cvref_t<_From, __id<std::remove_cvref_t<_To>>>;
```
ericniebler, line 497: I don't think the `remove_cvref_t` is necessary. I think it's a bug if `_To` is ever cvref-qualified here. Maybe see if that's true and if so, drop the `remove_cvref_t`.

include/stdexec/execution.hpp
```diff
@@ -3917,12 +3917,12 @@ namespace stdexec {
...
   namespace __stopped_as_xxx {
-    template <class _SenderId, class _ReceiverId>
+    template <class _CvrefSenderId, class _ReceiverId>
       struct __operation;
 
-    template <class _SenderId, class _ReceiverId>
+    template <class _CvrefSenderId, class _ReceiverId>
       struct __receiver {
-        using _Sender = stdexec::__t<_SenderId>;
+        using _Sender = stdexec::__t<_CvrefSenderId>;
```
ericniebler, line 3925: I have an issue to remove the dependency on the sender type here. Just sayin', no action needed for this PR.

include/stdexec/execution.hpp
```diff
@@ -3984,10 +3984,10 @@ namespace stdexec {
 
           template <class _Self, class _Receiver>
             using __operation_t =
-              stdexec::__t<__operation<stdexec::__id<__copy_cvref_t<_Self, _Sender>>, stdexec::__id<_Receiver>>>;
+              stdexec::__t<__operation<stdexec::__cvref_id<_Self, _Sender>, stdexec::__id<_Receiver>>>;
           template <class _Self, class _Receiver>
             using __receiver_t =
-              stdexec::__t<__receiver<stdexec::__id<__copy_cvref_t<_Self, _Sender>>, stdexec::__id<_Receiver>>>;
+              stdexec::__t<__receiver<stdexec::__cvref_id<_Self, _Sender>, stdexec::__id<_Receiver>>>;
```
ericniebler, line 3990: Wow, the code was totally doing the wrong thing before. :-P Nice catch.

ericniebler:
This is terrific. I've been meaning to go through all such places and fix them. Thank you!

include/nvexec/stream/common.cuh
```diff
@@ -578,7 +578,7 @@ namespace nvexec {
           operation_state_base_t<stdexec::__id<OuterReceiver>>&>;
 
     template <class Sender, class InnerReceiver, class OuterReceiver>
-      using stream_op_state_t = operation_state_t<stdexec::__id<Sender>,
+      using stream_op_state_t = operation_state_t<stdexec::__cvref_id<Sender, Sender>,
```
ccotter, line 581: The nvexec `when_all` instantiates this `stream_op_state_t` with `Sender&&`, which lead to my `static_assert` detecting `__id` being passed a reference type.

It's not clear just by looking at this code that `stream_op_state_t` is meant to be passed a possibly cvref Sender.

include/stdexec/__detail/__meta.hpp
ccotter: The static_assert I added should detect if `_To` is cvref-qualified. Indeed, the `remove_cvref_t` is not needed.

include/nvexec/stream/common.cuh
ericniebler: I don't see a use of `stream_op_state_t` in when_all.cuh. Could you point me at the code in question?

ccotter: I got mixed up. `stream_op_state_t` is referenced by then.cuh's `connect` which passes a possible cvref qualified `Sender` to `stream_op_state_t `.


It was `exit_operation_state_t` that when_all.cuh references that also passes a cvref qualified `Sender.`

ericniebler: @senior-zero, could you have a look at the changes Chris made in nvexec?

senior-zero, line 581: Should it be the following instead?
```suggestion
      using stream_op_state_t = operation_state_t<stdexec::__cvref_id<Sender, std::remove_cvref_t<Sender>>,
```

include/nvexec/stream/common.cuh
```diff
@@ -531,14 +531,14 @@ namespace nvexec {
...
-      using operation_state_t = stdexec::__t<operation_state_<SenderId, InnerReceiverId, OuterReceiverId>>;
+    template <class CvrefSenderId, class InnerReceiverId, class OuterReceiverId>
+      using operation_state_t = stdexec::__t<operation_state_<CvrefSenderId, InnerReceiverId, OuterReceiverId>>;
 
     template <class Sender, class OuterReceiver>
         requires stream_receiver<OuterReceiver>
       using exit_operation_state_t 
         = operation_state_t<
-            stdexec::__id<Sender>, 
+            stdexec::__cvref_id<Sender, std::remove_cvref_t<Sender>>,
```
senior-zero, line 541: We might call it `CvrefSender`, analogous to other changes.

senior-zero (approved):
Thank you for the fixes! A few minor suggestions below.

include/nvexec/stream/common.cuh
senior-zero: @ericniebler, @ccotter changes look good. Stream operation should be able to accept possibly cvref sender. It might be worth adding a test case at some point.

PR #757, opened 2023-02-02
Implement at_coroutine_exit

maikel:
This is a draft for at_coroutine_exit as it is found in libunifex.

Im addressing currently the following TODOs 

- [x] unhandled_stop should invoke the callback
- [x] fix completion_signatures for __die_on_stop

### DIFF
--- a/include/exec/at_coroutine_exit.hpp
+++ b/include/exec/at_coroutine_exit.hpp
@@ -1,4 +1,5 @@
 /*
+ * Copyright (c) Facebook, Inc. and its affiliates.
  * Copyright (c) 2021-2022 NVIDIA Corporation
  *
  * Licensed under the Apache License Version 2.0 with LLVM Exceptions
@@ -17,22 +18,6 @@
 
 // The original idea is taken from libunifex and adapted to stdexec.
 
-/*
- * Copyright (c) Facebook, Inc. and its affiliates.
- *
- * Licensed under the Apache License Version 2.0 with LLVM Exceptions
- * (the "License"); you may not use this file except in compliance with
- * the License. You may obtain a copy of the License at
- *
- *   https://llvm.org/LICENSE.txt
- *
- * Unless required by applicable law or agreed to in writing, software
- * distributed under the License is distributed on an "AS IS" BASIS,
- * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
- * See the License for the specific language governing permissions and
- * limitations under the License.
- */
-
 #include <exception>
 #include <type_traits>
 
@@ -48,22 +33,18 @@ struct __die_on_stop_t {
         using __id = __receiver_id;
         _Receiver __receiver_;
 
-        template <class... _Args>
-          friend void tag_invoke(stdexec::set_value_t, __t&& __self,
-                                _Args&&... __args) noexcept {
-            if constexpr (stdexec::__nothrow_callable<stdexec::set_value_t, _Receiver&&, _Args&&...>) {
-              stdexec::set_value((_Receiver &&) __self.__receiver_, (_Args &&) __args...);
-            } else {
-              try {
-                stdexec::set_value((_Receiver &&) __self.__receiver_, (_Args &&) __args...);
-              } catch (...) {
-                stdexec::set_error((_Receiver &&) __self.__receiver_, std::current_exception());
-              }
-            }
+        template <stdexec::__decays_to<__t> _Self, class... _Args>
+            requires stdexec::__callable<stdexec::set_value_t, _Receiver, std::decay_t<_Args>...>
+          friend void tag_invoke(stdexec::set_value_t, _Self&& __self,
+                                _Args&&... __args) noexcept try {
+            stdexec::set_value((_Receiver &&) __self.__receiver_, (_Args &&) __args...);
+          } catch (...) {
+            stdexec::set_error((_Receiver &&) __self.__receiver_, std::current_exception());
           }
 
-        template <class _Error>
-          friend void tag_invoke(stdexec::set_error_t, __t&& __self,
+        template <stdexec::__decays_to<__t> _Self, class _Error>
+            requires stdexec::__callable<stdexec::set_error_t, _Receiver, std::decay_t<_Error>>
+          friend void tag_invoke(stdexec::set_error_t, _Self&& __self,
                                 _Error&& __error) noexcept {
             stdexec::set_error((_Receiver &&) __self.__receiver_, (_Error &&) __error);
           }
@@ -72,7 +53,7 @@ struct __die_on_stop_t {
           std::terminate();
         }
 
-        friend auto tag_invoke(stdexec::get_env_t, const __t& __self) noexcept {
+        friend decltype(auto) tag_invoke(stdexec::get_env_t, const __t& __self) noexcept {
           return stdexec::get_env(__self.__receiver_);
         }
       };
@@ -80,19 +61,19 @@ struct __die_on_stop_t {
   template <class _R> using __receiver = stdexec::__t<__receiver_id<_R>>;
 
   template <class _Sig> struct __return_type {
-    using type = _Sig;
+    using __t = _Sig;
   };
 
   template <class _R, class... _Args> struct __return_type<_R(_Args...)> {
-    using type = _R;
+    using __t = _R;
   };
 
   template <class _Sig>
-  using __return_type_t = typename __return_type<_Sig>::type;
+  using __return_type_t = typename __return_type<_Sig>::__t;
 
   template <class _Tag, class _With = stdexec::__> struct __replace_tag {
     template <class _Arg>
-    using __f = stdexec::__if_c<std::is_same_v<__return_type_t<_Arg>, _Tag>, _With, _Arg>;
+    using __f = stdexec::__if<std::is_same<__return_type_t<_Arg>, _Tag>, _With, _Arg>;
   };
 
   template <class _Tag>
@@ -238,16 +219,16 @@ template <class... _Ts>
 struct __at_coroutine_exit_t {
 private:
   template <typename _Action, typename... _Ts>
-    static __task<_Ts...> at_coroutine_exit(_Action&& __action, _Ts&&... __ts) {
+    static __task<_Ts...> at_coroutine_exit(_Action __action, _Ts... __ts) {
       co_await ((_Action&&) __action)((_Ts&&) __ts...);
     }
 
 public:
   template <typename _Action, typename... _Ts>
-      requires stdexec::__callable<_Action, _Ts...>
+      requires stdexec::__callable<std::decay_t<_Action>, std::decay_t<_Ts>...>
     __task<_Ts...> operator()(_Action&& __action, _Ts&&... __ts) const {
       return __at_coroutine_exit_t::at_coroutine_exit((_Action &&) __action,
-                                                    (_Ts &&) __ts...);
+                                                      (_Ts &&) __ts...);
     }
 };
 inline constexpr __at_coroutine_exit_t at_coroutine_exit{};
--- a/include/exec/at_coroutine_exit.hpp
+++ b/include/exec/at_coroutine_exit.hpp
@@ -43,65 +43,99 @@ namespace exec {
 namespace __at_coroutine_exit {
 struct __die_on_stop_t {
   template <class _Receiver>
-    struct __receiver {
-      _Receiver __receiver_;
+    struct __receiver_id {
+      struct __t {
+        using __id = __receiver_id;
+        _Receiver __receiver_;
 
-      template <class... _Args>
-        friend void tag_invoke(stdexec::set_value_t, __receiver&& __self,
-                              _Args&&... __args) noexcept {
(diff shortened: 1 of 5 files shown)


include/exec/at_coroutine_exit.hpp
```diff
@@ -0,0 +1,256 @@
...
+namespace exec {
+namespace __at_coroutine_exit {
+struct __die_on_stop_t {
+  template <class _Receiver>
+    struct __receiver_id {
+      struct __t {
+        using __id = __receiver_id;
+        _Receiver __receiver_;
+
+        template <class... _Args>
```
ericniebler, line 51:
```suggestion
        template <class... _Args>
            requires __callable<stdexec::set_value_t, _Receiver, _Args...>
```

include/exec/at_coroutine_exit.hpp
```diff
@@ -0,0 +1,256 @@
...
+            } else {
+              try {
+                stdexec::set_value((_Receiver &&) __self.__receiver_, (_Args &&) __args...);
+              } catch (...) {
+                stdexec::set_error((_Receiver &&) __self.__receiver_, std::current_exception());
+              }
+            }
+          }
+
+        template <class _Error>
```
ericniebler, line 65:
```suggestion
        template <class _Error>
            requires __callable<stdexec::set_error_t, _Receiver, _Error>
```

include/exec/at_coroutine_exit.hpp
```diff
@@ -0,0 +1,256 @@
...
+          friend void tag_invoke(stdexec::set_error_t, __t&& __self,
+                                _Error&& __error) noexcept {
+            stdexec::set_error((_Receiver &&) __self.__receiver_, (_Error &&) __error);
+          }
+
+        [[noreturn]] friend void tag_invoke(stdexec::set_stopped_t, __t&&) noexcept {
+          std::terminate();
+        }
+
+        friend auto tag_invoke(stdexec::get_env_t, const __t& __self) noexcept {
```
ericniebler, line 75:
```suggestion
        friend decltype(auto) tag_invoke(stdexec::get_env_t, const __t& __self) noexcept {
```

include/exec/at_coroutine_exit.hpp
```diff
@@ -0,0 +1,256 @@
...
+  template <class _Sig> struct __return_type {
+    using type = _Sig;
+  };
+
+  template <class _R, class... _Args> struct __return_type<_R(_Args...)> {
+    using type = _R;
+  };
+
+  template <class _Sig>
+  using __return_type_t = typename __return_type<_Sig>::type;
```
ericniebler, line 91:
```suggestion
  template <class _Sig> struct __return_type {
    using __t = _Sig;
  };

  template <class _R, class... _Args> struct __return_type<_R(_Args...)> {
    using __t = _R;
  };

  template <class _Sig>
  using __return_type_t = stdexec::__t<__return_type<_Sig>>;
```

include/exec/at_coroutine_exit.hpp
```diff
@@ -0,0 +1,256 @@
...
+
+  template <class _Tag, class _With = stdexec::__> struct __replace_tag {
+    template <class _Arg>
+    using __f = stdexec::__if_c<std::is_same_v<__return_type_t<_Arg>, _Tag>, _With, _Arg>;
```
ericniebler, line 95:
```suggestion
    using __f = stdexec::__if<std::is_same<__return_type_t<_Arg>, _Tag>, _With, _Arg>;
```

The metaprogramming facilities will avoid instantiating the `std::is_same` class template if you write it this way.

include/exec/at_coroutine_exit.hpp
```diff
@@ -0,0 +1,256 @@
...
+  template <class _Sender>
+    struct __sender_id {
+      using __old_sigs = stdexec::__completion_signatures_of_t<_Sender, stdexec::env_of_t<_Sender>>;
+      using __completion_signatures_t =
+          stdexec::__mapply<__remove_signatures_with_tag<stdexec::set_stopped_t>, __old_sigs>;
+      class __t {
+      public:
+        using __id = __sender_id;
+
+        using completion_signatures = __completion_signatures_t;
```
ericniebler, line 116: This part isn't right, I don't think. You should be computing the completion signatures using the receiver's environment by defining a `get_completion_signatures_t` tag_invoke overload.

include/exec/at_coroutine_exit.hpp
```diff
@@ -0,0 +1,256 @@
...
+struct __at_coroutine_exit_t {
+private:
+  template <typename _Action, typename... _Ts>
+    static __task<_Ts...> at_coroutine_exit(_Action&& __action, _Ts&&... __ts) {
+      co_await ((_Action&&) __action)((_Ts&&) __ts...);
+    }
+
+public:
+  template <typename _Action, typename... _Ts>
+      requires stdexec::__callable<_Action, _Ts...>
```
ericniebler, line 247:
```suggestion
      requires stdexec::__callable<decay_t<_Action>, decay_t<_Ts>...>
```

include/exec/at_coroutine_exit.hpp
```diff
@@ -0,0 +1,256 @@
...
+      std::tuple<_Ts&...> __args_{};
+    };
+
+    __coro::coroutine_handle<__promise> __coro_;
+  };
+
+struct __at_coroutine_exit_t {
+private:
+  template <typename _Action, typename... _Ts>
+    static __task<_Ts...> at_coroutine_exit(_Action&& __action, _Ts&&... __ts) {
```
ericniebler, line 241:
```suggestion
    static __task<_Ts...> at_coroutine_exit(_Action __action, _Ts... __ts) {
```

test/exec/test_at_coroutine_exit.cpp
```diff
@@ -0,0 +1,239 @@
...
+  stdexec::sync_wait(invoke_actions_after_stop3(result, stdexec::schedule(scheduler)));
+  REQUIRE(result == 5);
+
+  result = 0;
+  stdexec::sync_wait(invoke_actions_after_stop4(result, stdexec::schedule(scheduler)));
+  REQUIRE(result == 1);
+}
+
+
+#ifdef __linux__
```
ericniebler, line 212: I'm curious why this is restricted to `__linux__`.

test/exec/test_at_coroutine_exit.cpp
```diff
@@ -0,0 +1,239 @@
...
+}
+}
+
+TEST_CASE("at_coroutine_exit terminates after stop within action", "[task][at_coroutine_exit]")
+{
+  stopped_scheduler scheduler;
+  auto stop = stdexec::schedule(scheduler);
+  REQUIRE_TERMINATE([](auto stop) { stdexec::sync_wait(stop_in_action(stop)); }, std::move(stop));
+}
+#endif
```
ericniebler, line 239: Are these all the test cases from libunifex?

ericniebler (commented):
I have yet to convince myself that all of the logic and test cases from the libunifex implementation has been ported over, but here are my initial observations...

test/exec/test_at_coroutine_exit.cpp
maikel: No, not all. I will add all of them today!

test/exec/test_at_coroutine_exit.cpp
maikel: `__linux__` is not accurate. I think one would need to check in CMake if calling those POSIX functions does compile and define some Macro at configure time. As a workaround I could just check the existence of `<unistd.h>` and `<sys/wait.h>`  with  `__has_include`

include/exec/at_coroutine_exit.hpp
maikel: This fixes some nasty compile errors when looking for a promise constructor overload. But I don't quite understand why yet...

test/exec/test_at_coroutine_exit.cpp
maikel: Indeed, the following test fails, because the continuation gets called and I know where I am wrong

```cpp
  int result = 0;
  stdexec::sync_wait(with_continuation(result, test_one_cleanup_action_with_stop(result)));
  REQUIRE(result == 2);
```

maikel:
I have fixed unhandled_stopped, but It needs an intrusive change to `__with_awaitable_senders_base`. Do you have any better ideas on how to get the type erased stop handler?

include/exec/at_coroutine_exit.hpp
```diff
@@ -0,0 +1,256 @@
...
+  template <class _Receiver>
+    struct __receiver_id {
+      struct __t {
+        using __id = __receiver_id;
+        _Receiver __receiver_;
+
+        template <class... _Args>
+          friend void tag_invoke(stdexec::set_value_t, __t&& __self,
+                                _Args&&... __args) noexcept {
+            if constexpr (stdexec::__nothrow_callable<stdexec::set_value_t, _Receiver&&, _Args&&...>) {
```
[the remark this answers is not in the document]
maikel: Okay, I've just kept the try / catch clauses

ericniebler:
> Do you have any better ideas on how to get the type erased stop handler?

Maybe we take a page from libunifex and bundle together the parent promise's `coroutine_handle` and stopped callback into a `continuation_handle<>` struct. Then `with_awaitable_senders::continuation()` would return a `continuation_handle<>` instead of a `coroutine_handle<>`. Thoughts?

test/exec/test_at_coroutine_exit.cpp
```diff
@@ -0,0 +1,307 @@
...
+}
+
+void test_cancel_in_cleanup_action_causes_death(int& result) {
+  task<void> t = []() -> task<void> {
+    co_await at_coroutine_exit([]() -> task<void> {
+      co_await stop();
+    });
+  }();
+  sync_wait(std::move(t)); // causes termination
+  // ADD_FAILURE() << "He didn't fall? Inconceivable!";
```
ericniebler, line 133: LOL nice. I'm curious why this is commented out though.

ericniebler (commented):
A few more minor niggles... Ok the whole, this is looking excellent.

include/exec/at_coroutine_exit.hpp
```diff
@@ -0,0 +1,250 @@
...
+        _Receiver __receiver_;
+
+        template <stdexec::__decays_to<__t> _Self, class... _Args>
+            requires stdexec::__callable<stdexec::set_value_t, _Receiver, std::decay_t<_Args>...>
+          friend void tag_invoke(stdexec::set_value_t, _Self&& __self,
+                                _Args&&... __args) noexcept try {
+            stdexec::set_value((_Receiver &&) __self.__receiver_, (_Args &&) __args...);
+          } catch (...) {
+            stdexec::set_error((_Receiver &&) __self.__receiver_, std::current_exception());
+          }
```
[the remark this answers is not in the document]
maikel: Is set_value not allowed to take arguments by value?

test/exec/test_at_coroutine_exit.cpp
maikel: Those tests are literally copied from unifex, which uses gtest, I think. I defined REQUIRE_TERMINATE to test for termination, although adding here a REQUIRE(false) is more accurate. I will add something equivalent.

include/exec/at_coroutine_exit.hpp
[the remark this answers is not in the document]
maikel: I will obey, but I am not fully convinced. I believe any evil move constructors could throw and I do not see yet where those are forbidden.

include/exec/at_coroutine_exit.hpp
```diff
@@ -0,0 +1,250 @@
...
+    __task(__task&& __that) noexcept
+    : __coro_(std::exchange(__that.__coro_, {})) {}
+
+    bool await_ready() const noexcept { return false; }
+
+    template <typename _Promise>
+        requires requires (_Promise& __promise, __coro::coroutine_handle<promise_type> __h) {
+          { __promise.continuation() } -> std::convertible_to<__coro::coroutine_handle<>>;
+          { __promise.set_continuation(__h) };
+        }
```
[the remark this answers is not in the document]
maikel: I have defined `template <class _Promise> concept __has_continuation;` but I am open for suggestions

include/exec/at_coroutine_exit.hpp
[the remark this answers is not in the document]
ericniebler: `set_value_t::operator()` takes by forwarding reference, is unconditionally `noexcept`, and mandates that the `tag_income` overload to which it dispatches is `noexcept`.

ericniebler:
Eventually we'll have to make `continuation_handle` normative and add it to P2300, but I'd like to iterate on its design a bit. But not in this PR. I think this one is ready to go, once tests pass.

Thank you!

ericniebler (approved):

bustercopley:
Hi @maikel,

In `test\exec\test_at_coroutine_exit.cpp`, the definition of the `REQUIRE_TERMINATE` macro is protected by `#ifdef STDEXEC_HAS_FORK`, but there are test definitions that use `REQUIRE_TERMINATE` and aren't protected by an `#ifdef`. (The invocations are protected, but not the definitions.)

This leads to compilation errors when `STDEXEC_HAS_FORK` is not defined, as for example on MinGW.

PR #758, opened 2023-02-05
when_any implementation

maikel:
`when_any` is to `std::variant` what `when_all` is to `std::tuple`.

- `when_any(s1, ..., sN)` completes if all senders of s1 to sN complete.
- whenever the first sender `S` completes with any result `R` (via either `set_value_t`, `set_error_t`, `set_stopped_t`), a stop is requested for all remaining senders.
- once all senders have completed `when_any(s1, ..., sN)` completes with `R`
- If a stop of `when_any(s1, ..., sN)` is requested before all senders have completed it will itself complete with `set_stopped`.


include/stdexec/execution.hpp
```diff
@@ -5485,6 +5485,203 @@ namespace stdexec {
   using __when_all::transfer_when_all_with_variant_t;
   inline constexpr transfer_when_all_with_variant_t transfer_when_all_with_variant{};
 
+  namespace __when_any_ {
+  template <class _BaseEnv>
+  using __env_t =
+      __make_env_t<_BaseEnv, __with<get_stop_token_t, in_place_stop_token>>;
+
+  template <class _Receiver, class _Op> struct __receiver_id {
```
ericniebler, line 5493: Template heads on a separate line, pls. Here and elsewhere.

```suggestion
  template <class _Receiver, class _Op>
    struct __receiver_id {
```

And although I like the naming scheme of the `_id` suffix for class templates that are ids, the point of the feature is to make the type names read well. A type made up of things like `__foo::__sender` and `__foo::__receiver` will read better than if they were `__foo::__sender_id` and `__foo::__receiver_id`.

Finally, I have all sorts of trouble when trying to parameterize the receiver on the type of the operation state. It's too easy to get into a situation where contraints on the receiver's methods require the type of the operation state type to be complete, or ending up with unfathomable recursion in constraints.

What I generally do instead is to factor out into a base class the part of the op state that the receiver needs to know about. This is usually everything except the inner operation state(s). Then the receiver stores a pointer to the base, keeping cycles out of the type system.

include/stdexec/execution.hpp
```diff
@@ -5485,6 +5485,203 @@ namespace stdexec {
...
+      }
+
+      __env_t<env_of_t<_Receiver>> get_env() const noexcept {
+        using with_token = __with<get_stop_token_t, in_place_stop_token>;
+        auto __token = with_token{__op_->__stop_source_.get_token()};
+        return __make_env(stdexec::get_env(__op_->__receiver_), (with_token &&) __token);
+      }
+
+      template <__one_of<set_value_t, set_error_t, set_stopped_t> _CPO,
+                class... _Args>
```
ericniebler, line 5515: This needs to be constrained somehow.

include/stdexec/execution.hpp
```diff
@@ -5485,6 +5485,203 @@ namespace stdexec {
...
+      template <class _CPO, class... _Args>
+      void notify(_CPO, _Args&&... __args) noexcept {
+        __op_->notify(_CPO{}, ((_Args &&) __args)...);
+      }
+
+      __env_t<env_of_t<_Receiver>> get_env() const noexcept {
+        using with_token = __with<get_stop_token_t, in_place_stop_token>;
+        auto __token = with_token{__op_->__stop_source_.get_token()};
+        return __make_env(stdexec::get_env(__op_->__receiver_), (with_token &&) __token);
+      }
```
ericniebler, line 5512: I'm curious why you made this a member function instead of just doing this in the `get_env_t` `tag_invoke` overload below.

include/stdexec/execution.hpp
```diff
@@ -5485,6 +5485,203 @@ namespace stdexec {
...
+  };
+
+  template <class _Env, class... _Senders>
+  using __completion_signatures_t = __t<__completion_signatures<_Env, _Senders...>>;
+
+  template <class _Sig> struct __signature_to_tuple;
+
+  template <class _Ret, class... _Args> struct __signature_to_tuple<_Ret(_Args...)> {
+    using __t = std::tuple<_Ret, _Args...>;
+  };
```
ericniebler, line 5545: I would have to benchmark it to be sure, but I bet something like this would compile quicker:

```c++
template <class _Ret, class... _Args>
  std::tuple<_Ret, _Args...> __signature_to_tuple_(_Ret(*)(_Args...));
template <class _Sig>
  using __signature_to_tuple =
    decltype(stdexec::__signature_to_tuple_((_Sig*) nullptr));
```

Instantiating a class template is expensive.

I bet you could do something similar with the `__completion_signatures` class template above.

include/stdexec/execution.hpp
```diff
@@ -5485,6 +5485,203 @@ namespace stdexec {
...
+
+  template <class _Env, class... _Senders>
+  using __result_type_t =
+      __mapply<__transform<__q<__signature_to_tuple_t>,
+                          __mbind_front_q<std::variant, std::monostate>>,
+              __completion_signatures_t<_Env, _Senders...>>;
+
+  template <class _Receiver, class... _Senders> struct __op_id {
+    struct __t {
+      __t(__t&&) = delete;
```
ericniebler, line 5558:
```suggestion
    struct __t : __immovable {
```

include/stdexec/execution.hpp
```diff
@@ -5485,6 +5485,203 @@ namespace stdexec {
...
+
+      __t(std::tuple<_Senders...>&& __senders, _Receiver&& __rcvr)
+          : __t{(std::tuple<_Senders...> &&) __senders, (_Receiver &&) __rcvr,
+                std::make_index_sequence<sizeof...(_Senders)>{}} {}
+
+      template <std::size_t... _Is>
+      __t(std::tuple<_Senders...>&& __senders, _Receiver&& __rcvr,
+          std::index_sequence<_Is...>)
+          : __receiver_{(_Receiver &&) __rcvr}, __ops_{__conv{[&__senders, this] {
```
ericniebler, line 5567:
```suggestion
          : __receiver_{(_Receiver &&) __rcvr}
          , __ops_{__conv{[&__senders, this] {
```

include/stdexec/execution.hpp
```diff
@@ -5485,6 +5485,203 @@ namespace stdexec {
...
+
+      struct __on_stop_requested {
+        in_place_stop_source& __stop_source_;
+        void operator()() noexcept { __stop_source_.request_stop(); }
+      };
+
+      // If this hits true, we store the reuslt
+      std::atomic<bool> __emplaced_{false};
+      // If this hits zero, we forward any reuslt to the receiver
+      std::atomic<int> __count_{sizeof...(_Senders)};
```
ericniebler, line 5580:
```suggestion
      // If this hits true, we store the result
      std::atomic<bool> __emplaced_{false};
      // If this hits zero, we forward any result to the receiver
      std::atomic<int> __count_{sizeof...(_Senders)};
```

include/stdexec/execution.hpp
```diff
@@ -5485,6 +5485,203 @@ namespace stdexec {
...
+
+      template <class _R>
+      friend __op<std::remove_cvref_t<_R>, _Sender, _Senders...> 
+        tag_invoke(connect_t, __t&& __self, _R&& __rcvr) noexcept {
+          return {(std::tuple<_Sender, _Senders...> &&) __self.__senders_, (_R &&) __rcvr};
+        }
+
+      friend __empty_env tag_invoke(get_env_t, const __t& __self) noexcept {
+        return {};
+      }
```
ericniebler, line 5657: This is the default behavior of `get_env` now, so this friend function isn't needed. I would strike it.

include/stdexec/execution.hpp
```diff
@@ -5485,6 +5485,203 @@ namespace stdexec {
...
+    struct __t {
+      using __id = __sender_id;
+
+      std::tuple<_Sender, _Senders...> __senders_;
+
+      template <class _R>
+      friend __op<std::remove_cvref_t<_R>, _Sender, _Senders...> 
+        tag_invoke(connect_t, __t&& __self, _R&& __rcvr) noexcept {
+          return {(std::tuple<_Sender, _Senders...> &&) __self.__senders_, (_R &&) __rcvr};
+        }
```
ericniebler, line 5653: This needs constraints. Also, I'm curious why you're not supporting `connect`-ing lvalue `when_any` senders.

include/stdexec/execution.hpp
```diff
@@ -5485,6 +5485,203 @@ namespace stdexec {
...
+         -> __completion_signatures_t<_Env, _Sender, _Senders...> requires true;
+    };
+  };
+
+  template <class... _Senders> using __sender = __t<__sender_id<_Senders...>>;
+
+  struct __when_any_t {
+    template <class... _Senders>
+      requires ((stdexec::sender<_Senders> && ...) && sizeof...(_Senders) > 0)
+    auto operator()(_Senders&&... __senders) const
```
ericniebler, line 5674: Prefer explicitly specifying return types. If you do, then you can `decltype` the function call expression without instantiating the function template. That's a compile-time win, and it can also prevent hard errors when the function is under-constrained.

include/stdexec/execution.hpp
```diff
@@ -5485,6 +5485,203 @@ namespace stdexec {
...
+      friend auto tag_invoke(get_completion_signatures_t, _Self&& __self, _Env __env) noexcept
+         -> __completion_signatures_t<_Env, _Sender, _Senders...> requires true;
+    };
+  };
+
+  template <class... _Senders> using __sender = __t<__sender_id<_Senders...>>;
+
+  struct __when_any_t {
+    template <class... _Senders>
+      requires ((stdexec::sender<_Senders> && ...) && sizeof...(_Senders) > 0)
```
ericniebler, line 5673: Simpler:
```suggestion
    template <stdexec::sender... _Senders>
      requires (sizeof...(_Senders) > 0)
```

ericniebler (changes requested):
This is beautiful! A new small suggestions. Also, since this is not in P2300 it can't be in namespace stdexec. Please put it in `include/exec/when_any.hpp`, and rename it from `__when_any` to `when_any`.

Thank you!

maikel:
> This is beautiful! A few small suggestions. Also, since this is not in P2300 it can't be in namespace stdexec. Please put it in `include/exec/when_any.hpp`, and rename it from `__when_any` to `when_any`.
> 
> Thank you!

Thank you for the review and thank you for considering to include `when_any` into the library extensions. I admit that I consider this algorithm as vital as `when_all` and tried to sneak it into `stdexec` on purpose. But I understand the priorities and I will try to address everything after my office time.

Since `when_any` and `when_all` have similar properties for its `completion_scheduler` should I also provide `transform_* ` or `*_into_variant` variations? Do you consider a `when_any` algorithm with a predicate useful?

include/stdexec/execution.hpp
maikel: I have tried to benchmark that with `-ftime-report` both on g++-11 and clang++-16 on a reduced example that uses `exec::when_any`. My benchmark was probably bad but I couldn't find a clear winner. gcc and clang had differing favourites, by small margins. But I will adopt that change and the analouge one for `__completion_signatures` anyway, because the code seems simpler to me.

include/stdexec/execution.hpp
maikel: Good to know!

maikel:
I am not sure, if I handled all raised issues. I will reiterate on the weekend or on monday.

include/exec/when_any.hpp
```diff
@@ -0,0 +1,267 @@
...
+              // stop pending operations
+              __stop_source_.request_stop();
+            }
+            // make __result_ emplacement visible when __count_ goes from one to zero
+            // This relies on the fact that each sender will call notify() at most once
+            if (__count_.fetch_sub(1, std::memory_order_acq_rel) == 1) {
+              __on_stop_.reset();
+              if (!__result_) {
+                // TODO: can this happen? assert?
+                set_stopped((_Receiver &&) __receiver_);
```
ericniebler, line 121: Assert

ericniebler (approved):

maikel:
I think the noexcepts around connect_t and the op state are still wrong. I will continue on Monday. I also think I should come up with more tests around exception handling. But it's taking some shape 💪

maikel:
Ok, I think I have polished the code as far as I can. There is this final question of when to add an `set_error_t(std::exception_ptr)` signature to the `when_any` sender. I've chosen to omit it only in case where all input senders send only `nothrow_copy_constructible` arguments. This might be pessimistic in front of move-only situations but I am not sure how to detect those.

PR #761, opened 2023-02-09
simplify at_coroutine_exit metaprogramming

ericniebler:


include/exec/at_coroutine_exit.hpp
```diff
@@ -26,104 +26,82 @@
 
 namespace exec {
 namespace __at_coro_exit {
+using namespace stdexec;
+
+// TODO __die_on_stop(snd) is just upon_stopped(snd, []{std::terminate();}).
```
ericniebler, line 31:
```suggestion
```

Issue #764, opened 2023-02-13
Lifetime guarantees of a sender's queryable environment


ericniebler:
It's clear that a receiver's environment should have reference semantics. But is that also true for a sender's? Might it not be sometimes desirable to copy a sender's environment into the operation state, letting go of the sender itself?

If so, a sender's environment needs value semantics, or else a way to "deep copy" the environment. The latter could be as simple as a way of enumerating the queries that an environment supports, which could be useful elsewhere as well; _e.g._, it could be useful in a facility that generates type-erased wrappers for queryables.

Runner-2019:
mark

PR #765, opened 2023-02-14
[WIP] Implement type erased wrappers

maikel:
This is a work in progress PR. I've put it here, such that if someone sees something very wrong, you can stop me early.


include/stdexec/execution.hpp
```diff
@@ -5736,13 +5736,428 @@ namespace stdexec {
...
+          constexpr void (*operator()(_Tag(*)(void(*)(_As...) noexcept)) const noexcept)(void*, _As...) noexcept {
+          return +[](void* __storage, _As... __as) noexcept -> void {
+            static_assert(__nothrow_callable<_Tag, __type<_T>, _Storage&, _As...>);
+            return _Tag{}(__type<_T>{}, *(_Storage*) __storage, (_As&&) __as...);
+          };
+        }
+      };
+
+    struct __get_vtable_t {
+      template <class Storage>
```
ericniebler, line 5833:
```suggestion
      template <class _Storage>
```

... and elsewhere.

include/stdexec/execution.hpp
```diff
@@ -5736,13 +5736,428 @@ namespace stdexec {
   inline constexpr sync_wait_with_variant_t sync_wait_with_variant{};
 
   namespace __any {
+    template <class T>
+    struct __type{};
```
ericniebler, line 5740: You can use `__mtype` from the `__meta.hpp` header.

include/stdexec/execution.hpp
```diff
@@ -5736,13 +5736,428 @@ namespace stdexec {
...
+          requires __tag_invocable_r<void*, __get_object_pointer_t, const Storage&>
+        void* operator()(const Storage& __storage) const noexcept {
+          return tag_invoke(__get_object_pointer_t{}, __storage);
+        }
+    };
+    inline constexpr __get_object_pointer_t __get_object_pointer{};
+
+    struct __delete_t {
+      template <class _Storage, class _T>
+          // requires tag_invocable<__delete_t, __type<_T>, _Storage&>
```
ericniebler, line 5852: I'm curious why this constraint is commented out.

include/stdexec/execution.hpp
```diff
@@ -5736,13 +5736,428 @@ namespace stdexec {
...
+          return __self.__object_pointer_;
+        }
+
+        template <class _T>
+          friend void tag_invoke(__delete_t, __type<_T>, __t& __self) noexcept {
+            using _Alloc = typename 
+                std::allocator_traits<_Allocator>::template rebind_alloc<_T>;
+            if (__self.__object_pointer_) {
+              _Alloc __alloc{__self.__allocator_};
+              _T* __pointer = reinterpret_cast<_T*>(std::exchange(__self.__object_pointer_, nullptr));
```
ericniebler, line 6017: I'm pretty sure this can be a `static_cast`.

include/stdexec/execution.hpp
```diff
@@ -5736,13 +5736,428 @@ namespace stdexec {
...
+            } catch (...) {
+              std::allocator_traits<_Alloc>::deallocate(__alloc, __pointer, 1);
+              throw;
+            }
+          }
+
+        const __vtable_t* __vtable_{nullptr};
+        void* __object_pointer_{nullptr};
+        [[no_unique_address]] _Allocator __allocator_{};
+      };
```
ericniebler, line 6152: It should be possible to share implementation between the different storage types, particularly the movable and copyable ones.

include/stdexec/execution.hpp
```diff
@@ -5848,10 +6237,130 @@ namespace stdexec {
...
+      using __unique_operation_storage = __storage_t<__unique_storage<>, __operation_vtable>;
+
+      template <class _Op>
+        inline constexpr __operation_vtable __op_vtbl{__operation_vtable_fn<_Op>{}()};
+
+      template <class _Op>
+        constexpr const __operation_vtable*
+        tag_invoke(__create_vtable_t<__operation_vtable>, __type<_Op>) noexcept {
+          return &__op_vtbl<_Op>;
+        }
```
ericniebler, line 6270: I try to make all `tag_invoke` overloads hidden friend functions. Hint: the type `_Op` in the signature above is an associated entity of `__type<_Op>` and will be searched for `tag_invoke` overloads.

include/stdexec/execution.hpp
```diff
@@ -5848,10 +6237,130 @@ namespace stdexec {
...
+
+      template <class _Sigs, class _Queries>
+        using __receiver_ref = __mapply<__mbind_front<__q<__rec::__ref>, _Sigs>, _Queries>;
+
+      template <class _Sender, class _Receiver, typename _Queries>
```
ericniebler, line 6275:
```suggestion
      template <class _Sender, class _Receiver, class _Queries>
```

ericniebler (commented):
This looks very good. I would love to see a storage option that used an internal buffer to store small senders/op states _in situ_.

maikel:
> This looks very good. I would love to see a storage option that used an internal buffer to store small senders/op states _in situ_. EDIT: I'm fine with making that a separate PR though, if you prefer.

No, I have planned to implement sbo here, but didn't have time today. I just pushed whatever I had when I finished today. Sorry for the noise

include/stdexec/execution.hpp
maikel: Oh I have missed that!

include/stdexec/execution.hpp
maikel: I will try!

include/stdexec/execution.hpp
maikel: My bad. Forgot to redo that after debugging an issue.

include/stdexec/execution.hpp
maikel: Correct

include/stdexec/execution.hpp
```diff
@@ -5874,112 +5872,100 @@ namespace stdexec {
...
 
-       private:
-        friend const _Vtable* tag_invoke(__get_vtable_t, const __t& __self) noexcept {
-          return __self.__vtable_;
-        }
-
-        friend void* tag_invoke(__get_object_pointer_t, const __t& __self) noexcept {
-          return __self.__object_pointer_;
-        }
+        using __vtable_t = std::conditional_t<_Copyable,
```
ericniebler, line 5907:
```suggestion
        using __vtable_t = __if_c<_Copyable,
```

include/stdexec/execution.hpp
```diff
@@ -6000,114 +5986,30 @@ namespace stdexec {
...
-          : __vtable_(&__storage_vtbl<__t, decay_t<_T>, _Vtable, 
-                                      __delete_t(void(*)() noexcept),
-                                      __copy_construct_t(void(*)(const __t& other))>)
-          {
-            using _ValueType = decay_t<_T>;
-            using _Alloc = typename 
-                std::allocator_traits<_Allocator>::template rebind_alloc<_ValueType>;
+        template <class _T, class... _As>
+          void __construct_small(_As&&... __args...) {
+            static_assert(sizeof(_T) <= __buffer_size);
```
ericniebler, line 5991: A `static_assert` about alignment would be good also.

include/stdexec/execution.hpp
```diff
@@ -5848,10 +6139,133 @@ namespace stdexec {
...
+
+          template <receiver_of<_Sigs> _Rcvr>
+              requires sender_to<__sender, _Rcvr>
+            friend __operation<__sender, std::decay_t<_Rcvr>, _ReceiverQueries> 
+            tag_invoke(connect_t, __sender&& __self, _Rcvr&& __rcvr) {
+              return {(_Rcvr&&) __rcvr, (__sender&&) __self};
+            }
+        };
+    } // namespace __sender
+  } // namepsace __any
```
ericniebler, line 6268: I just realized I started this work in the wrong place. It can't be in `stdexec::`. Could you move it to `exec/any_sender_of.hpp`, or something?

include/stdexec/functional.hpp
```diff
@@ -77,6 +77,12 @@ namespace stdexec {
           tag_invoke((_Tag&&) __tag, (_Args&&) __args...);
         };
 
+    template <class _Ret, class _Tag, class... _Args>
+      concept __tag_invocable_r =
+        requires (_Tag __tag, _Args&&... __args) {
+          { tag_invoke((_Tag&&) __tag, (_Args&&) __args...) } -> same_as<_Ret>;
```
ericniebler, line 83: Let's make this behave like `std::is_invocable_r`. The constraint should be `convertible_to<_Ret>`, except that if `_Ret` is `void`, any return type will do.

ericniebler (commented):
Looks fantastic. Needs more tests obvs, and a good surface syntax. :thinking:

include/stdexec/execution.hpp
```diff
@@ -5736,13 +5737,329 @@ namespace stdexec {
...
+          if (__vtable_) {
+            (*__vtable_)(__delete, this);
+            __object_pointer_ = nullptr;
+            __vtable_ = nullptr;
+          }
+        }
+
+       private:
+        template <class _T, class... _As>
+          void __construct_small(_As&&... __args...) {
```
ericniebler, line 5990:
```suggestion
          void __construct_small(_As&&... __args) {
```

include/stdexec/functional.hpp
maikel: Yep

include/stdexec/execution.hpp
```diff
@@ -5736,13 +5737,329 @@ namespace stdexec {
...
+            __self.__reset();
+            if (!__other.__object_pointer_) {
+              return;
+            }
+            const _T& __other_object = *static_cast<const _T*>(__other.__object_pointer_); 
+            if constexpr (__is_small<_T>) {
+              __self.__construct_small<_T>(__other_object);
+            } else {
+              __self.__construct_large<_T>(__other_object);
+            }
```
ericniebler, line 6046:
```suggestion
            if constexpr (__is_small<_T>) {
              __self.template __construct_small<_T>(__other_object);
            } else {
              __self.template __construct_large<_T>(__other_object);
            }
```

include/stdexec/execution.hpp
maikel: Yes I will do so

include/stdexec/execution.hpp
```diff
@@ -5736,13 +5737,359 @@ namespace stdexec {
...
+      using __copyable_storage = __basic_storage<_Allocator, true>;
+
+    template <class _Allocator, bool _Copyable, std::size_t _Alignment, std::size_t _InlineSize>
+    template <class _Vtable> 
+      class __basic_storage<_Allocator, _Copyable, _Alignment, _InlineSize>::__storage<_Vtable>::__t {
+        static constexpr std::size_t __buffer_size = std::max(_InlineSize, sizeof(void*));
+        static constexpr std::size_t __alignment = std::max(_Alignment, alignof(void*));
+        using __with_copy = __copy_construct_t(void(*)(const __t&));
+        using __with_move = __move_construct_t(void(*)(__t&&) noexcept);
+        using __with_delete = __delete_t(void(*)() noexcept);
```
ericniebler, line 5921: Incidentally, these can be simply:

```suggestion
        using __with_copy = __copy_construct_t(void(const __t&));
        using __with_move = __move_construct_t(void(__t&&) noexcept);
        using __with_delete = __delete_t(void() noexcept);
```

C++ is weird like that.

include/stdexec/functional.hpp
ericniebler: I suppose you could just check the expression:

```c++
static_cast<_Ret>(tag_invoke((_Tag&&) __tag, (_Args&&) __args...));
```

maikel:
Ok, I have implemented a copyable sender env. Did you imagine in #764 something like this? The difference to `__rec::__ref` is, that the "any_sender" class is a factory for the env. But I am open to any alternative suggestions and designs. I still need to reiterate a couple of times to cleanup the code and fix some pending issues. But I think we habe a good start to think about a good syntax.

ericniebler:
> Ok, I have implemented a copyable sender env. Did you imagine in https://github.com/NVIDIA/stdexec/issues/764 something like this?

Not quite. Firstly, I'm wondering why we need `any_env_of` at all. I don't think `any_sender_of` needs it. But the real trouble with copying environments is that they often have reference semantics, like `std::reference_wrapper`. You can copy the env, but you're not _deep_ copying it, which can lead to lifetime issues.

To really copy an env and get a value type, you would need a way to iterate over the queries that the env supports. Then you would create a new object with the same queries but that stores everything by value.

include/exec/any_sender_of.hpp
```diff
@@ -0,0 +1,683 @@
...
+            friend void tag_invoke(__move_construct_t, __mtype<_T>, __t& __self, __t&& __other) noexcept {
+              if (!__other.__object_pointer_) {
+                return;
+              }
+              _T* __pointer = static_cast<_T*>(std::exchange(__other.__object_pointer_, nullptr));
+              if constexpr (__is_small<_T>) {
+                _T& __other_object = *__pointer; 
+                __self.template __construct_small<_T>((_T&&)__other_object);
+              } else {
+                __self.__object_pointer_ = __pointer;
```
ericniebler, line 385: I suggest null-ing out the source object pointer here.

maikel:
> Not quite. Firstly, I'm wondering why we need `any_env_of` at all. I don't think `any_sender_of` needs it. But the real trouble with copying environments is that they often have reference semantics, like `std::reference_wrapper`. You can copy the env, but you're not _deep_ copying it, which can lead to lifetime issues.

I see. I really misunderstood multiple things here. Thank you for clarifying!

maikel:
I will reopen against the main branch

PR #771, opened 2023-02-17
Implement any_sender_of

maikel:
This PR is WIP and implements `exec::any_sender_of`. Tests are still missing. 

It is not complete yet but I (try to describe and) propose the following syntax:

`any_sender_of<Ts...>` shall be a type-erased sender with the following completion signatures:
- For each type `Ts` that is not a valid completion signature `set_value_t(Ts)` is a completion signature of `any_sender_of<Ts...>`
- Otherwise if `Ts` is a valid completion signature (one of `set_value_t(As...)`, `set_error_t(E)` or `set_stopped()`), `Ts` itself is a completion signature of  `any_sender_of<Ts...>`

We can also type-erase queries for the sender and the receivers that will be connected to.
Let `any-sender-of` denote an instance of the type erased sender. Let `Queries(Ret(As...)...` denote a parameter pack of query signatures. 

`with_sender_queries<any-sender-of, Queries(Ret(As...))...>` is a type erased sender which is queryable for each query signature
`with_receiver_queries<any-sender-of, Queries(Ret(As...)...>` is a type erased sender, that connects with receivers that are queryable for each query signature.

Furthermore, I want to introduce `any_scheduler<>`, which is a scheduler with 
`completion_signatures_of_t<tag_invoke_result_t<schedule, const any_scheduler<>&>> = completion_signatures<set_value_t()>` Each `Ts` in `any_scheduler<Ts...>` shall contribute in a similar fashion to the completion signatures as in the case of `any_sender_of<Ts...>`.

In addition to `with_sender_queries` and `with_receiver_queries` there is a `with_scheduler_queries`.

Is this ok? Do we need `any_stop_token`  or any other type erased types?


include/exec/any_sender_of.hpp
```diff
@@ -0,0 +1,805 @@
...
+  template <class... _Sigs>
+    using __sender_of = __t<__any::__sender<completion_signatures<_Sigs...>>>;
+
+  template <class _T>
+    using __transform_value = __if_c<__is_compl_sig<_T>, _T, set_value_t(_T)>;
+
+  } // namepsace __any
+
+  template <class... _Ts>
+    using any_sender_of = __any::__sender_of<__any::__transform_value<_Ts>...>;
```
ericniebler, line 797: This is clever, but I'm a tad concerned that `any_sender_of<int, short>` might get mistaken for `any_sender_of<set_value_t(int, short)>` instead of `any_sender_of<set_value_t(int), set_value_t(short)>`.

Maybe cap the permitted value types at one. You can additionally require that only the first template parameter is allowed to name a value type; the rest have to be completion signatures.

include/exec/any_sender_of.hpp
```diff
@@ -0,0 +1,805 @@
...
+
+  template <class... _Ts>
+    using any_scheduler = __any::__scheduler<stdexec::completion_signatures<__any::__transform_value<_Ts>...>>;
```
ericniebler, line 800: Making `any_sender` and `any_scheduler` aliases for complex types has a negative impact on diagnostics. I would prefer that they were classes that inherited (or aggregated w/ interface forwarding) their implementations.

include/exec/any_sender_of.hpp
```diff
@@ -0,0 +1,805 @@
...
+      using with_receiver_queries = decltype(__with_receiver_queries_fn((_S*)nullptr, (_Queries*)nullptr...));
+
+    template <class _Sigs, class _SchSQs, class _RQ, class _SQs, class... _NewSchSQs>
+      auto __with_scheduler_queries_fn(__scheduler<_Sigs, _SchSQs, _SQs, _RQ>*, _NewSchSQs*...) 
+        -> __scheduler<_Sigs, __minvoke<__mconcat<>, _SchSQs, __types<_NewSchSQs...>>, _SQs, _RQ>;
+
+    template <class _S, class... _Queries>
+      using with_scheduler_queries = decltype(__with_scheduler_queries_fn((_S*)nullptr, (_Queries*)nullptr...));
+
+    // BUGBUG: __completion_signature<_T> returns always true???
```
ericniebler, line 778: I've added test cases for that, and I don't see a problem.

include/exec/any_sender_of.hpp
```diff
@@ -0,0 +1,805 @@
...
+
+    template <class _S, class... _Queries>
+      using with_receiver_queries = decltype(__with_receiver_queries_fn((_S*)nullptr, (_Queries*)nullptr...));
+
+    template <class _Sigs, class _SchSQs, class _RQ, class _SQs, class... _NewSchSQs>
+      auto __with_scheduler_queries_fn(__scheduler<_Sigs, _SchSQs, _SQs, _RQ>*, _NewSchSQs*...) 
+        -> __scheduler<_Sigs, __minvoke<__mconcat<>, _SchSQs, __types<_NewSchSQs...>>, _SQs, _RQ>;
+
+    template <class _S, class... _Queries>
+      using with_scheduler_queries = decltype(__with_scheduler_queries_fn((_S*)nullptr, (_Queries*)nullptr...));
```
ericniebler, line 776: I see what you're doing here. I like the direction. Needs tests, obvs.

Another potential direction I've noodled on is to have folks name a type-erased receiver type w/ queries, use that to name a type-erased sender w/ queries, and then use that to name a type-erased scheduler w/ queries.

I'm not sure yet which one I would like better.

include/exec/any_sender_of.hpp
```diff
@@ -0,0 +1,805 @@
...
+    using __return_value_t = decltype(__return_value_((_Sig*)nullptr));
+
+    template <class _Sigs = completion_signatures<set_value_t()>,
+              class _SchedulerQueries = __types<>,
+              class _SenderQueries = __types<>,
+              class _ReceiverQueries = __types<>>
+      class __scheduler {
+      public:
+        template <class _Scheduler>
+            requires (!__decays_to<_Scheduler, __scheduler>) // && scheduler<_Scheduler>)
```
ericniebler, line 675: Seems to be a bad interaction with Catch2. This recursion is bothersome. I haven't  yet managed to track it down, but naturally it will need to be fixed.

include/stdexec/execution.hpp
```diff
@@ -239,11 +243,9 @@ namespace stdexec {
...
 
           // Forward the receiver queries:
-          template <
-              __none_of<__tag_of<_WithIds>..., get_completion_signatures_t> _Tag,
-              same_as<__t> _Self,
-              class... _As>
-              requires __callable<_Tag, const __base_env_of<_Self>&, _As...>
+          template <__forwarding_query _Tag, same_as<__t> _Self, class... _As>
+              requires __none_of<_Tag, __tag_of<_WithIds>...> &&
+                __callable<_Tag, const __base_env_of<_Self>&, _As...>
```
ericniebler, line 248: Nice catch, thanks.

include/exec/any_sender_of.hpp
maikel: Yep, I had the very same doubts

include/exec/any_sender_of.hpp
maikel: I have no strong feelings for any variation. Is it really too verbose to specify all possible completion signatures? I'm not sure anymore.

include/stdexec/execution.hpp
maikel: Thanks, but that was you in a prior commit 3weeks ago, I think.

include/exec/any_sender_of.hpp
maikel: I think gcc-11 fails

ericniebler: Yup, just pushed a workaround.

include/exec/any_sender_of.hpp
ericniebler: I think I fixed this in d4bdc56.

ericniebler:
> - `any_receiver<Sigs, Queries...>` is non-owning and is cheap to copy.

True, better call it `any_receiver_ref`.

> - `any_sender<Queries...>` is owning and move-assignable and only rvalue connectable

This is as it should be.

> - `any_scheduler<Queries...>` is "owning" and copy-assignable. Copies are "deep" but we assume that schedulers are handles to execution contexts and thus themself cheap to copy.

Also correct.

> Should there be a `any_sender` alternative which is be copy assignable and "const-lref"-connectable? 

Yeah, probably called `any_multishot_sender`. Doesn't have to be in this PR though.

include/exec/any_sender_of.hpp
```diff
@@ -0,0 +1,847 @@
...
+            : __sender_((_Sender&&) __sender) {}
+
+          template <auto... _SchedulerQueries>
+            class any_scheduler {
+              using __schedule_completions = 
+                  stdexec::__concat_completion_signatures_t<
+                    _Completions,
+                    stdexec::completion_signatures<stdexec::set_value_t()>>;
+              using __schedule_receiver = any_receiver<__schedule_completions, _ReceiverQueries...>;
+              using __schedule_sender = __schedule_receiver::template any_sender<
```
ericniebler, line 810:
```suggestion
              using __schedule_sender = typename __schedule_receiver::template any_sender<
```

PR #777, opened 2023-02-22
Update to `rapidsai/devcontainers`, nvhpc 23.1

trxcllnt:


.github/workflows/ci.yml
```diff
@@ -52,8 +52,8 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          - { name: "nvc++ 22.11", cuda: "11.8", cxx: "mpic++",  tag: "nvhpc22.11",       gpu: "v100", driver: "520", arch: "amd64" }
-          - { name: "clang 16", cuda: "11.8",    cxx: "clang++", tag: "llvmdev-cuda11.8", gpu: "v100", driver: "520", arch: "amd64" }
+          - { name: "nvc++ 23.1", cuda: "12.0",  cxx: "mpic++",  tag: "nvhpc23.1",       gpu: "v100", driver: "520", arch: "amd64" }
```
miscco, line 55: To be sure shouldnt the driver version also change?

trxcllnt: Yeah, but I'll have to check if we have any 525 action runners yet. If not, we could disable the `NVIDIA_REQUIRE_CUDA` checks and `cuda-compat` might save us?

trxcllnt: Nevermind, I forgot `cuda-compat` doesn't apply to compilation. Will ask about our timelines for 525 runners.

trxcllnt:
`clang++-16` doesn't play well with CUDA 12, since 12 dropped support for Kepler. We have to explicitly specify the arch during configure via `-DCMAKE_CUDA_ARCHITECTURES=75-real` to make them work together.

Setting `-DCMAKE_CUDA_ARCHITECTURES=native` makes CMake compile a test file (via `clang++ -x cuda`). Since clang defaults to `ptxas -gpu_name sm_35`, and that's now illegal in CUDA 12, this test compilation fails. Forcing it to use a supported arch seems to work.

.github/workflows/ci.yml
trxcllnt: Yeah, we'll have to wait until we have action runners with driver 525.

PR #778, opened 2023-02-22
Conditionally call `set_error` on receiver connected to `schedule_from`

msimberg:
I think there's more to be done here, and I'm not 100% sure about this, hence still a draft. I'd appreciate confirmation on a few questions.

This intends to fix `schedule_from` advertising itself as not sending `std::exception_ptr` in its completion signatures, but still instantiating `set_error(receiver, exception_ptr)` unconditionally. This is of course problematic if the receiver needs to store the `exception_ptr`.

This PR currently wraps signaling the connected receiver into the `__try_call` helper so that `set_error` isn't unnecessarily instantiated with `exception_ptr`.

Questions:
- As far as I can tell the variant can never be valueless by exception here. If emplacing into the variant fails earlier `set_error` will be called elsewhere and we won't even reach this `__complete` function. Does this sound correct? I've got an assertion there for this which isn't triggered by tests, but I'd appreciate someone else checking if this makes sense.
- I see there's a conditional `set_error_t(std::exception_ptr)` added if any of the _error_ types sent by the scheduler are potentially throwing when copied. Shouldn't the value and error types sent by the input senders be checked for that as well, or is that already handled somewhere where I'm not seeing it? (cf. https://github.com/NVIDIA/stdexec/blob/caaf7e2cab6a5429ebeff7aa3500ec7646f44fc1/include/stdexec/execution.hpp#L4701-L4717).

Still to do: add a test case (this failure was triggered by a simple `transfer(...) | then(...) | transfer(...)` so I'll add at least that as a test).

### DIFF
--- a/include/stdexec/execution.hpp
+++ b/include/stdexec/execution.hpp
@@ -4701,7 +4701,7 @@ namespace stdexec {
         using __all_nothrow_decay_copyable = __bool<(__nothrow_decay_copyable<_Errs> && ...)>;
 
         template <class _Env>
-        using __with_error_t = //
+        using __scheduler_with_error_t = //
           __if_c<
             __v<error_types_of_t<schedule_result_t<_Scheduler>, _Env, __all_nothrow_decay_copyable>>,
             completion_signatures<>,
@@ -4712,15 +4712,25 @@ namespace stdexec {
           __make_completion_signatures<
             schedule_result_t<_Scheduler>,
             _Env,
-            __with_error_t<_Env>,
+            __scheduler_with_error_t<_Env>,
             __mconst<completion_signatures<>>>;
+
+        template <class _Env>
+        using __input_sender_with_error_t = //
+          __if_c<
+            __v<value_types_of_t<_Sender, _Env, __all_nothrow_decay_copyable, __mand>> &&
+            __v<error_types_of_t<_Sender, _Env, __all_nothrow_decay_copyable>>,
+            completion_signatures<>,
+            __with_exception_ptr>;
 
         template <__decays_to<__t> _Self, class _Env>
         friend auto tag_invoke(get_completion_signatures_t, _Self&&, _Env)
           -> __make_completion_signatures<
             __copy_cvref_t<_Self, _Sender>,
             _Env,
-            __scheduler_completions_t<_Env>,
+            __concat_completion_signatures_t<
+              __scheduler_completions_t<_Env>,
+              __input_sender_with_error_t<_Env>>,
             __decay_signature<set_value_t>,
             __decay_signature<set_error_t>>;
       };
--- a/include/stdexec/execution.hpp
+++ b/include/stdexec/execution.hpp
@@ -4634,7 +4634,8 @@ namespace stdexec {
           start(__op_state.__state1_);
         }
 
-        void __complete() noexcept try {
+        void __complete() noexcept {
+          STDEXEC_ASSERT(!__data_.valueless_by_exception());
           std::visit(
             [&]<class _Tup>(_Tup& __tupl) -> void {
               if constexpr (same_as<_Tup, std::monostate>) {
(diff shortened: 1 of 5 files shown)


include/stdexec/execution.hpp
```diff
@@ -4634,23 +4634,22 @@ namespace stdexec {
...
+                    __try_call(
+                      (_Receiver &&) __rcvr_, __tag, (_Receiver &&) __rcvr_, (_Args &&) __args...);
                   },
                   __tupl);
               }
             },
             __data_);
-        } catch (...) {
-
-          set_error((_Receiver&&) __rcvr_, std::current_exception());
```
ericniebler, line 4653: The `__complete` function isn't doing anything that can fail with an exception. In the bad old days, `set_value` could throw, which is probably why this `try`/`catch` is here, but it can simply be removed now. No need for `__try_call`.

include/stdexec/execution.hpp
```diff
@@ -4634,23 +4634,22 @@ namespace stdexec {
           start(__op_state.__state1_);
         }
 
-        void __complete() noexcept try {
+        void __complete() noexcept {
+          STDEXEC_ASSERT(!__data_.valueless_by_exception());
```
ericniebler, line 4638: This is a good addition, thanks.

ericniebler (changes requested):
> I see there's a conditional `set_error_t(std::exception_ptr)` added if any of the _error_ types sent by the scheduler are potentially throwing when copied. Shouldn't the value and error types sent by the input senders be checked for that as well [...] ?

Yes, the scheduler sender doesn't send values, but the input sender may. If decay-copying those values are potentially throwing, then `set_error_t(std::exeption_ptr)` _should_ be added to the list of possible completions. Good catch.

include/stdexec/execution.hpp
msimberg: Yep, I figured the try-catch was a leftover for `set_value` previously being potentially throwing. However, the `__try_call` is still needed for potentially throwing value type copies, right?

> If decay-copying those values are potentially throwing, then set_error_t(std::exeption_ptr) should be added to the list of possible completions.

Or am I misunderstanding this comment:

> No need for __try_call.

?

ericniebler: The concepts ensure this can never happen. The `connect` function requires that a receiver is capable of receiving all the completions of the sender. If a `set_value` expression is potentially throwing, then the concept check fails. The sender cannot connect to such a receiver.

See: https://godbolt.org/z/9Pohne6nr

include/stdexec/execution.hpp
```diff
@@ -4634,23 +4634,22 @@ namespace stdexec {
...
           std::visit(
             [&]<class _Tup>(_Tup& __tupl) -> void {
               if constexpr (same_as<_Tup, std::monostate>) {
                 std::terminate(); // reaching this indicates a bug in schedule_from
               } else {
                 std::apply(
                   [&]<class... _Args>(auto __tag, _Args&... __args) -> void {
-                    __tag((_Receiver&&) __rcvr_, (_Args&&) __args...);
+                    __try_call(
+                      (_Receiver &&) __rcvr_, __tag, (_Receiver &&) __rcvr_, (_Args &&) __args...);
```
ericniebler, line 4647:
```suggestion
                    __tag((_Receiver &&) __rcvr_, (_Args &&) __args...);
```

include/stdexec/execution.hpp
msimberg: Huh, TIL, thanks for explaining! I didn't realize it applied to the whole expression but now that you spell it out it certainly makes sense.

msimberg:
Thanks @ericniebler for the comments.

Heads up, I'll be less available in the next week and as you can tell not all tests compile yet, so this won't be ready for a while. There are a few tests in `test_on2.cpp` that don't like the addition of `__input_sender_with_error_t<_Env>`, but I have not been able to figure out if I'm expressing it wrong or if there's something fishy in `on` or something in that direction. Do the error messagese say something meaningful to you?

ericniebler:
Ya, I think #782 should address these failures.

ericniebler (approved):
Looks good to me. Anything else you plan to do, @msimberg ?

msimberg:
> Ya, I think #782 should address these failures. Do you mind if I rebase this PR once that fix has landed?

Oh nice, that seems to have worked out just fine!

> Looks good to me. Anything else you plan to do, @msimberg ?

Nope, I'm done here.

Issue #770, opened 2023-02-17
Fix completion signatures of `exec::when_any`


maikel:
I put this as a reminder here.

- the completion signatures of `when_any` should transform `set_value_t(Args...)` sigs into `set_value_t(std::decay_t<Args>&&...)`. 
- `set_error_t(std::exception_ptr)` should be added if all input signatures are `__nothrow_decay_copyable` and nothrow `move_constructible`. Currently we just check for nothrow `copy_constructible` which is pessimistic, if not even wrong for degenerate types, which are nothrow copy contsructible but not nothrow move constructible.

PR #779, opened 2023-02-22
Turn when_any into when_any_value and fix issue #770

maikel:
Hi,

we had this discussion in Discord, where `when_any(Senders...)` could be a `dematerialize(when_any_value(materialize(Senders...)))` but I think there is no way in the other direction.

- This PR solves issue #770 
- rewrites the current `when_any` into a `when_any_value`
  - completes if all input senders completed. 
  - If a stop is requested from an upstream Receiver, then `when_any_value` will complete with `set_stopped()`
  - When any sender completes with a call to `set_value` this algorithm requests a stop for all other senders.
  - If no stop is requested, `when_any_value` completes with `set_value(decay_t<Args>&&...)` of the first call in form of `set_value(Args&&...)`
  - If all senders complete with either `set_stopped` or `set_error` the last completion will be forwarded unless a stop was requested.
  - If moving  `__decayed_tuple<Args...>` or its construction from `Args&&... args` throw, an exception will be stored and forwarded upon completion of all senders.

I plan to add better tests, especially around completion signatures and error handling. 
I also plan to add materialize and dematerialize to put `when_any` back to place.

### DIFF
--- a/include/exec/when_any.hpp
+++ b/include/exec/when_any.hpp
@@ -40,7 +40,7 @@ namespace exec {
     using __signature_to_tuple_t = decltype(__signature_to_tuple_((_Sig*) nullptr));
 
     template <class _Tag>
-    struct __ret_equals_to {
+    struct __tag_equals_to {
       template <class _Sig>
       using __f = std::is_same<_Tag, __tag_of_sig_t<_Sig>>;
     };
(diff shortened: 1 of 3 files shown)


ericniebler:
I do not like this change. I do want `materialize` and `dematerialize` algorithms, but I don't want `when_any` implemented in terms of them. `when_any` is expensive enough at compile-time. I don't want any changes that will cause it to instantiate more templates than are necessary.

I do want #770 fixed though.

maikel:
OK let me reorganize my sources and I will come back with a new PR

Issue #767, opened 2023-02-15
Implement `on_coroutine_succeeded()` and `on_coroutine_failed()`


ericniebler:
In addition to `at_coroutine_exit()`, the `task<>` type can support `on_coroutine_succeeded()` and `on_coroutine_failed()`. Those will register a (possibly asynchronous) cleanup action that knows whether `unhandled_exception()` was called on the promise or not.

maikel:
I wonder if there is a particular reason to ignore the third channel, i.e. `on_coroutine_stopped()`. AFAIK we also have an indicator for with `unhandled_stopped`

ericniebler:
No reason. I just didn't think of it!

maikel:
This is done, right?

ericniebler:
Yup

PR #784, opened 2023-02-24
Implement on_coroutine_[completion]

maikel:
Hi,

This PR addresses the feature idea #767 and shows a sample implementation. 

Opposed to what, I thought, was suggested, I did not see how to use `unhandled_{exception,stopped}` directly. Instead I type-erase the index of the `__data_` variant in the parent promise and use `any_sender` to type erase either `just()` or the `__die_on_stop((_Awaitable&&) awaitable)` sender.

What I need is some direction for the implementation:
 - Is the index approach okay?
 - `any_sender` is ok until something like variant_sender is there?
 - emulate an `if`-like sender with something like
 ```cpp
 when_any(
     completes_if(is_completion) | let_value([] { return die_on_stop }), 
     completes_if(!is_completion) | let_value([] { return just(); }))
```

### DIFF
--- a/include/exec/at_coroutine_exit.hpp
+++ b/include/exec/at_coroutine_exit.hpp
@@ -23,7 +23,6 @@
 
 #include "stdexec/execution.hpp"
 #include "exec/task.hpp"
-#include "exec/any_sender_of.hpp"
 
 namespace exec {
   namespace __at_coro_exit {
--- a/include/exec/at_coroutine_exit.hpp
+++ b/include/exec/at_coroutine_exit.hpp
@@ -148,8 +148,8 @@ namespace exec {
 
      private:
       struct __final_awaitable {
-        static std::false_type await_ready() noexcept {
-          return {};
+        static constexpr bool await_ready() noexcept {
+          return false;
         }
 
         static __coro::coroutine_handle<>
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: include/exec/at_coroutine_exit.hpp [hunk 1]
```diff
@@ -226,4 +227,188 @@ namespace exec {
...
+  namespace __on_coroutine_succeeded {
+    using namespace stdexec;
+
+    enum class __on_coroutine_completion {
+      __stopped,
+      __value,
+      __error
+    };
+
+    __on_coroutine_completion __from_int(int __i) noexcept {
```
ericniebler, line 240:
```suggestion
    inline __on_coroutine_completion __from_int(int __i) noexcept {
```

include/exec/at_coroutine_exit.hpp
```diff
@@ -226,4 +227,188 @@ namespace exec {
   } // namespace __at_coro_exit
 
   inline constexpr __at_coro_exit::__at_coro_exit_t at_coroutine_exit{};
+
+  namespace __on_coroutine_succeeded {
+    using namespace stdexec;
+
+    enum class __on_coroutine_completion {
```
ericniebler, line 234:
```suggestion
    enum class __on_coroutine_completion : int {
```

ericniebler (approved):
Neat! Looks pretty great.

include/exec/at_coroutine_exit.hpp
```diff
@@ -226,4 +227,188 @@ namespace exec {
...
+        template <class _Awaitable>
+        decltype(auto) await_transform(_Awaitable&& __awaitable) noexcept {
+          STDEXEC_ASSERT(__parent_ && __get_index_callback_);
+          const int __index = __get_index_callback_(__parent_.address());
+          using __any_sender_t = typename any_receiver_ref<
+            completion_signatures<set_value_t(), set_error_t(std::exception_ptr)>>::any_sender<>;
+          __any_sender_t __sender = just();
+          if (__from_int(__index) == _OnCompletion) {
+            __sender = __at_coro_exit::__die_on_stop((_Awaitable&&) __awaitable);
+          }
```
ericniebler, line 345: This is a good place to use a variant sender once we have one. Instead, I would just give the `__die_on_stopped` sender adaptor a runtime `bool` to decide whether a stopped signal should either be forwarded or else `terminate()`.

include/exec/at_coroutine_exit.hpp
maikel: Good catch!

include/exec/at_coroutine_exit.hpp
maikel: briliant, gimme a second

maikel:
fantastic

PR #788, opened 2023-02-25
Migrate customization points to portability macros that support dispatching to named member functions

villevoutilainen:
Don't merge, for review only for now.

EDIT: The customization macros expand to one of three possible customization protocols:
1. `tag_invoke`
2. Named static member functions that take the CPO as a tag
3. Named instance member function that take the CPO as a tag (with deducing `this`)


ericniebler:
I'm of half a mind to nuke `execution_access`. If you want to make your `start()` member function private, you can friend `::execution::start_t`. Ditto for the other CPOs.

ericniebler:
Next steps on this PR:

Needs a helper for defining CPOs. At first, I'll want all CPO's to look first for `obj.member(tag)`, then `Obj::member(obj, tag)`, and finally `tag_invoke(tag, obj)` and deprecate the `tag_invoke` option. Then I'll want to drop tag_invoke but keep the other two, and (once everybody has deducing `this`) look only for the `obj.member(tag)` form. Ideally, I would only need to change this in one place.

I'm concerned about diagnostics during the transition. They're bad enough already, I don't want to make them any worse. I'll need to play around with ways to define these CPOs before we can forge ahead porting the other CPOs to members.

But it somebody beats me to it, all the better.

include/nvexec/stream/transfer.cuh
```diff
@@ -41,15 +41,15 @@ namespace nvexec::STDEXEC_STREAM_DETAIL_NS {
...
+          STDEXEC_DEFINE_CUSTOM(Env get_env)(this const receiver_t& self, get_env_t) noexcept {
             return self.op_state_.make_env();
           }
         };
 
         using task_t = continuation_task_t<receiver_t, variant_t>;
 
         cudaError_t status_{cudaSuccess};
-        context_state_t context_state_;
+        s context_state_t context_state_;
```
ericniebler, line 52:
```suggestion
        context_state_t context_state_;
```

include/exec/linux/io_uring_context.hpp
```diff
@@ -344,6 +344,12 @@ namespace exec {
         __wakeup_operation_.start();
       }
 
+      ~__context() noexcept {
+        if (!__pending_.empty()) {
+          run_until_empty();
+        }
+      }
```
ericniebler, line 351: I know @kirkshoop has strong feelings about blocking in destructors, and good reasons for those feelings. This change makes me sus.

maikel: Yes I know sorry. I shouldn't have pushed that. I will open a pr with a better solution. Currently an application terminates when the context is default constructed and run is not called at all. This is because the default constructor puts the wakeup op into the pending queue which has an assertion of  being not empty in its destructor.

ericniebler:
This effort will start again in a clean branch, since this PR is too far out of date.

PR #790, opened 2023-02-26
Only use std::exception_ptr in when_all when needed

maikel:


maikel:
The calls to variants emplace for the error types is not noexcept. Std variant is really bothersome here. I think wrapping the error variant into an optional might be better. Especially since it's not true anymore that the error variant is default constructible. Thoughts?

ericniebler:
> The calls to variants emplace for the error types is not noexcept. Std variant is really bothersome here.

Can you say more? On which line of code is `variant::emplace` causing trouble?

maikel:
It bothers me that emplace is unconditionally `noexcept(false)`. But looking at it again, it is at least specified when it throws an exception and it should be ok here. But since this commit conditionally takes out `std::execption_ptr` from the ErrorVariant its still needs to use something like `__nullable_variant` or `std::optional` or `__manual_lifetime`, correct?

include/stdexec/execution.hpp
```diff
@@ -5411,11 +5432,19 @@ namespace stdexec {
...
+          error_types_of_t<_Senders, __env_t<__id<_Env>>, __types>... >;
+
+      using __errors_variant = //
+        __if<
+          __all_value_and_error_args_nothrow_decay_copyable<_Env, __id<_Senders>...>,
+          __if<
+            std::is_same<__error_types, __types<>>,
+            std::variant<std::monostate>,
+            __mapply<__q<std::variant>, __error_types>>,
+          __minvoke<__push_back_unique<__q<std::variant>>, __error_types, std::exception_ptr>>;
```
ericniebler, line 5447: Probably best to just use `__nullable_variant` here and always have the `monostate` alternative first. Then you don't have to wrap the variant in an optional above.

maikel: Ok I did as you asked, but that had its own problems. I wasnt able to detect whether `std::monostate` is a valid error type and I had to manually ignore it as an error type. Constraining on `__callable<_Tag, _Receiver&&, std::monostate>` didn't work as expected, probably due to a bug in constrains of `schedule_from`s receiver. I couldnt solve that one.

ericniebler: I suppose you could avoid that problem by using an unspecified, non-public type instead of std::monostate.

maikel: I think the cleanest solution is manual lifetime but I will go with the private monostate type until I sneaked in the former.

PR #792, opened 2023-02-27
Scheduler affine coroutine task

maikel:
Hi here is a very rough and basic skeleton for a scheduler affine task type.

I would be glad about feedback and general direction for the implementation.


include/exec/task.hpp
```diff
@@ -21,16 +21,24 @@
 #include <utility>
 #include <variant>
 
-#include "../stdexec/coroutine.hpp"
-#include "../stdexec/execution.hpp"
-#include "../stdexec/__detail/__meta.hpp"
-#include "scope.hpp"
+#include "stdexec/coroutine.hpp"
+#include "stdexec/execution.hpp"
+#include "stdexec/__detail/__meta.hpp"
```
ericniebler, line 26: Please continue using relative paths for the includes.

include/exec/task.hpp
```diff
@@ -82,6 +103,72 @@ namespace exec {
     template <class _Ty>
     using default_task_context = __default_task_context_impl;
 
+    // This is another context that might be associated with a basic_task's promise type.
+    // It is used to implement the scheduler affinity feature.
+    class __scheduler_affine_task_context_impl {
```
ericniebler, line 108: Scheduler affinity should be part of the default behavior of `task`. So integrate these changed directly into the `__default_task[...]` stuff.

include/exec/task.hpp
```diff
@@ -257,6 +358,29 @@ namespace exec {
...
+        template <class _Awaitable>
+        decltype(auto) await_transform(_Awaitable&& __awaitable) noexcept
+          requires(
+            __completion_scheduler_provider<_Context>
+            && !stdexec::__decays_to<_Awaitable, __stick_on_scheduler_t>)
+        {
+          return stdexec::as_awaitable(
+            stdexec::transfer(
+              (_Awaitable&&) __awaitable,
+              stdexec::get_completion_scheduler<stdexec::set_value_t>(__context_)),
```
ericniebler, line 370: Rather than relying on the sender's completion scheduler, the `co_await` should squirrel the current scheduler away in a custom awaitable , and then post back to that scheduler on resume.

ericniebler (commented):
A few hints, I haven't given this a deep look-over.

include/exec/task.hpp
maikel: Ah, okay.

ericniebler: BTW, I particularly appreciate the greenfield effort you're making here. I implemented this feature in libunifex and feel I did a poor job of it. Don't emulate it, at least not blindly.

maikel: I have taken a look into your PR in libunifex. After I have understood all the whys I will happly apply all the changes

ericniebler: I don't think registering a cleanup action is necessary, FWIW. That's something I'd like to try to move away from.

maikel: I still try to find the goals/requirements of my "task" (not `exec::task`). Sorry I should have asked that way before.

I've pondered over your implementation and the difference to mine. Is the goal, that each `task` object has an associated `completion_scheduler<set_value_t>` which it is guaranteed to complete on? Then my implementation fails to satisfy that requirement, because it only relies on the _current_ scheduler within the context, which might lead to a different scheduler in the end. But how am I supposed to start a task with a specified `scheduler`? I can imagine a promise constructor, which checks for a `scheduler_arg_t` argument analogue to `allocator_arg_t` elsewhere in the standard. But how was that in your original design?

ericniebler: The goal is for the task to complete on whichever was the current scheduler when the task was started. So in general, a task will not know its completion scheduler.

To start a task on a particular scheduler, the idea is to just use the `on` algorithm, which sets the current scheduler. In other words, that part is already implemented.

maikel: Is it correct to reschedule on the initial scheduler within task awaitable type? What if my task is used with other sender algorithms? I will need to add tests. I've played around with rescheduling in the final suspend. That yields good results on clang but ICE on gcc ...

maikel: > Is it correct to reschedule on the initial scheduler within task awaitable type? What if task is used with other sender algorithms? I will need to add tests.

include/exec/task.hpp
```diff
@@ -141,22 +231,37 @@ namespace exec {
...
     template <>
     struct __default_awaiter_context<void> {
-      template <class _Ty>
-      explicit __default_awaiter_context(__default_task_context_impl&, _Ty&) noexcept {
+      template <__scheduler_affiinty _A, class _Ty>
+      explicit __default_awaiter_context(__default_task_context_impl<_A>&, _Ty&) noexcept {
       }
 
-      template <__indirect_stop_token_provider _ParentPromise>
+      template <__scheduler_affiinty _A, __indirect_stop_token_provider _ParentPromise>
```
mfbalin, line 255: typo here, affinity instead of affiinty.

maikel: thank you <3

include/exec/task.hpp
maikel: I've found the problem and improved the implementation. Next, I will look into clang-12 and why its not working there.

include/exec/task.hpp
maikel: Fixed

include/exec/task.hpp
```diff
@@ -267,6 +374,30 @@ namespace exec {
...
+              (_Awaitable&&) __awaitable,
+              stdexec::get_completion_scheduler<stdexec::set_value_t>(__context_)),
+            *this);
+        }
+
+        template <stdexec::__decays_to<complete_inline_t> _Awaitable>
+        decltype(auto) await_transform(_Awaitable&& __awaitable) noexcept {
+          __context_.set_scheduler(__awaitable.__scheduler_);
+          return stdexec::as_awaitable(stdexec::schedule(__awaitable.__scheduler_), *this);
+        }
```
ericniebler, line 393: I need to investigate how this works. I was imagining that complete_inline would be a sender adaptor.

test/exec/test_task.cpp
```diff
@@ -0,0 +1,172 @@
...
+
+  task<void> test_old_behaviour(
+    scheduler auto sched1,
+    scheduler auto sched2,
+    auto id_main_thread,
+    auto id1,
+    auto id2) {
+    CHECK(std::this_thread::get_id() == id_main_thread);
+    co_await schedule(sched1);
+    CHECK(std::this_thread::get_id() == id1); // We changed from main thread to context1
```
ericniebler, line 152: This should be on the main thread. The rule is: for any sender not wrapped in a complete_inline, transfer back to what was the "current" scheduler before the co_await.

I'd also like any use of complete_inline to cause an at_coroutine_exit action to be posted that transfers back to the starting scheduler.

ericniebler (changes requested):
More comments. I'm trying to review this on a phone (not near my computer) so I still haven't really studied with this as closely as I should.

maikel:
> More comments. I'm trying to review this on a phone (not near my computer) so I still haven't really studied with this as closely as I should.

No problem, I think it is still incomplete anyway! I am happy about any comments

include/exec/task.hpp
```diff
@@ -80,13 +128,40 @@ namespace exec {
...
 
     template <class _Ty>
-    using default_task_context = __default_task_context_impl;
+    using default_task_context = __default_task_context_impl<__scheduler_affinity::__sticky>;
+
+    template <class _Ty>
+    using __raw_task_context = __default_task_context_impl<__scheduler_affinity::__none>;
+
+    struct complete_inline_t {
+      __any_scheduler __scheduler_;
```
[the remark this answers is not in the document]
maikel: The promise sets its current scheduler in the await transform overload for this awaitable

include/exec/task.hpp
```diff
@@ -80,13 +128,40 @@ namespace exec {
...
-      explicit __default_awaiter_context(__default_task_context_impl&, _ParentPromise&) noexcept {
+      template <__scheduler_affinity _A>
+      explicit __default_awaiter_context(
+        __default_task_context_impl<_A>& __self,
+        _ParentPromise& __parent) noexcept {
+        if constexpr (
+          __indirect_completion_scheduler_provider<_ParentPromise>
+          && _A == __scheduler_affinity::__sticky) {
+          __self.__scheduler_ = stdexec::get_completion_scheduler<stdexec::set_value_t>(
+            get_env(__parent));
```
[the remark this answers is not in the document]
maikel: Good catch. The access to the scheduler is named wrong. This is the current scheduler

include/exec/task.hpp
```diff
@@ -267,6 +374,30 @@ namespace exec {
...
 
+        template <stdexec::sender _Awaitable>
+          requires(
+            !stdexec::__decays_to<_Awaitable, complete_inline_t>
+            && __completion_scheduler_provider<_Context>)
+        decltype(auto) await_transform(_Awaitable&& __awaitable) noexcept {
+          return stdexec::as_awaitable(
+            stdexec::transfer(
+              (_Awaitable&&) __awaitable,
+              stdexec::get_completion_scheduler<stdexec::set_value_t>(__context_)),
```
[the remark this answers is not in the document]
maikel: Good idea, I will add comments in the next iteration

include/exec/task.hpp
maikel: Currently, it's a special awaitable just to set the scheduler from within a coro. Currently it's the only way to modify the current scheduler. I plan to add a specialized on(sched, task) that sets the initial scheduler, which otherwise defaults to inline scheduler unless it is suspended as an awaitable in a parent context where inherits the current scheduler of its parent

test/exec/test_task.cpp
maikel: I need to check since I'm on phone too. But I think the current scheduler is inline scheduler and that's why, it behaves non sticky

include/exec/task.hpp
maikel: Regarding the clang-12 issue: The problem seems to be, that std::this_thread::get_id() reports a wrong thread id. When I step through the debugger in Release mode with debug symbols, I can see that the coro was scheduled on the correct thread but the ID remains unchanged. This happens only in Release mode and does not happen with clang-16. Can someone verify that?

include/exec/task.hpp
```diff
@@ -80,13 +128,40 @@ namespace exec {
...
+    using default_task_context = __default_task_context_impl<__scheduler_affinity::__sticky>;
+
+    template <class _Ty>
+    using __raw_task_context = __default_task_context_impl<__scheduler_affinity::__none>;
+
+    struct complete_inline_t {
+      __any_scheduler __scheduler_;
+
+      static constexpr bool await_ready() noexcept {
+        return false;
```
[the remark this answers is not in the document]
maikel: Good catch, changed it to return `std::true_type`. But I think its never called anyway, since we `await_transform` that into a  `schedule(__scheduler_)`

include/exec/task.hpp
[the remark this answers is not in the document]
maikel: I have changed that to `stdexec::get_env(__parent).get_scheduler()`. Should I be more explicit and call it `get_current_scheduler()` ?

maikel:
I refactored the tests such that it works around the buggy behaviour of clang-12. The tests should all pass now, hopefully

test/exec/test_task.cpp
maikel: Ok, I changed the implementation to use `stdexec::get_scheduler` and `stdexec::__scheduler_provider`. Now The test performs as *you* expect it will be rescheduled on the main thread. This is because the task takes scheduler from the `sync_wait`'s run_loop. This is so exciting!

maikel: > I'd also like any use of complete_inline to cause an at_coroutine_exit action to be posted that transfers back to the starting scheduler.

You are right. The task is currently always rescheduled. It should be better to do that only if we used complete inline. I was slow at that one and will change that asap 😍

ericniebler:
I'm working on the last little bit of this PR. Just need to change `complete_inline` into a sender adaptor and then this will be done.

ericniebler (commented):
I'm working on the last little bit of this PR. Just need to change `complete_inline` into a sender adaptor and then this will be done.

ericniebler:
> Just need to change `complete_inline` into a sender adaptor and then this will be done.

It occurs to me that this can only be made to work for senders that know their completion scheduler. Being able to use this with only _some_ senders would be a very frustrating experience.

I think I like your idea better. But maybe `reschedule_coroutine_on(scheduler)` might be more self-documenting than `complete_inline(scheduler)`.

ericniebler (approved):

maikel:
Do we need an extra overload of `await_transform` to transfer for Awaitables that are not `basic_task` ?

ericniebler:
> Do we need an extra overload of `await_transform` to transfer for Awaitables that are not `basic_task` ?

Isn't that what [this overload](https://github.com/NVIDIA/stdexec/pull/792/files#diff-843f9c6b4dbda84cb3e226899ec39be9195e8235821edd173183b692280f541bR371-R378) is for?

PR #793, opened 2023-02-28
less convoluted implementation of `__connect_awaitable`

ericniebler:


include/stdexec/execution.hpp
```diff
@@ -1243,6 +1244,30 @@ namespace stdexec {
...
+          static constexpr bool await_ready() noexcept {
+            return false;
+          }
+
+          void await_suspend(__coro::coroutine_handle<>) noexcept {
+            __fn_();
+          }
+
+          void await_resume() noexcept {
+          }
```
lewissbaker, line 1264: Annotate with `[[noreturn]]` and put `std::terminate()` in here (or `std::unreachable()`).

include/stdexec/execution.hpp
```diff
@@ -1243,6 +1244,30 @@ namespace stdexec {
   /////////////////////////////////////////////////////////////////////////////
   // __connect_awaitable_
   namespace __connect_awaitable_ {
+    inline constexpr auto __co_call =
+      []<class _Fun, class... _Ts>(_Fun __fun, _Ts&&... __as) noexcept {
```
lewissbaker, line 1247: Why a lambda instead of just a normal function?

Normal functions should give better info in stack-traces, diagnostics, etc. - naming `__co_call` instead of "::$lambda 25::operator()".

include/stdexec/execution.hpp
```diff
@@ -1377,29 +1377,15 @@ namespace stdexec {
...
-            co_yield __fun(co_await (_Awaitable&&) __await);
+            co_await __co_call(set_value, (_Receiver&&) __rcvr, co_await (_Awaitable&&) __await);
+
         } catch (...) {
           __eptr = std::current_exception();
         }
-        co_yield [&]() noexcept -> void {
-          set_error((_Receiver&&) __rcvr, (std::exception_ptr&&) __eptr);
-        };
+        co_await __co_call(set_error, (_Receiver&&) __rcvr, (std::exception_ptr&&) __eptr);
```
lewissbaker, line 1387: I wonder if this call (and the whole try/catch + `__eptr` variable could be conditionally avoided if the `co_await` expression was noexcept.
It could avoid having to add `exception_ptr` as an error completion signature in that case.

lewissbaker (commented):
A couple of minor suggestions - otherwise looks ok to me.

include/stdexec/execution.hpp
ericniebler: How can I portably detect that?

lewissbaker: Good question. Ideally there would be a `__get_awaiter()` helper that called `operator co_await()` for you and that you could detect the noexceptness of that expression and then just check the await_ready/await_resume/await_suspend methods for noexceptness.

Wrap it all up into a `is_nothrow_awaitable_v<A, P>` query.

Technically, it's not implementable portably in library at the moment, though, as library can't exactly mimic the overload resolution of `operator co_await()`. But you could get pretty close and then just handwave away something that requires compilers to provide the necessary intrinsics to make stdlib implementations do the right thing.

ericniebler: Filed at #795

Issue #472, opened 2022-02-24
`let_xxx` algorithms always specify `set_error_t(exception_ptr)` in the completion signatures


anthonywilliams:
The `let_xxx` algorithms always specify `set_error_t(exception_ptr)` in the completion signatures, even if no exceptions can be thrown.

e.g. if I provide a `sender`, such that `execution::connect(my_sender,r)` is `noexcept`, and the arguments supplied to `set_value` by my sender are nothrow-move-constructible (e.g. `int`), and my sender never completes with `set_error`, then `let_value(my_sender,[](auto ...) noexcept{ return another_noexcept_sender{}; })` should never complete with `set_error` either, since none of the constituent parts throw.

Thus, we need another `potentially-throwing` determination here, that detects if any of:

* invoking `execution::connect(s,r)` can throw
* storing `args'...` in `op_state` can throw (i.e. any of the types in `Args` are *not* `nothrow_move_constructible`) for any of the possible sets of `Args` for the completion of the source sender `s`.
* invoking `connect(f(args'...),out_r)` can throw

and only includes the `set_error_t(exception_ptr)` in the completion signature if things are potentially throwing.

Chlorie:
This might be impossible to implement under the current spec since we only get the environment type but not the receiver type in `get_completion_signatures`, which means that we cannot decide whether the connection would fail.

anthonywilliams:
In order to implement this, then you connect the original sender with a proxy receiver. The proxy receiver then forwards to the final receiver. This way the type of the proxy receiver is known when determining whether or not the internal `connect` throws. I have implemented this.

Chlorie:
> In order to implement this, then you connect the original sender with a proxy receiver. The proxy receiver then forwards to the final receiver. This way the type of the proxy receiver is known when determining whether or not the internal `connect` throws. I have implemented this.

That is really a neat way that I haven't thought of to tackle the problem. Thanks for the inspiration!

ericniebler:
@anthonywilliams Your proxy receiver suggestion requires something like virtual dispatch if I understand correctly. You want to test the noexcept-ness of `connect` before the actual receiver type is known. The indirect function call will lead to worse runtime perf.

Do I have that right? In your opinion, this trade-off is worth it?

anthonywilliams:
[...]

PR #794, opened 2023-02-28
Implement the finally algorithm

maikel:
Hey,

here I implement the `finally` algorithm. I have done this such that the algorithm lazily connects the final sender and detects whether the second call to `connect` will throw or not and omits `set_error_t(std::exception_ptr)` in case it doesnt.

This flavor of the algorithm starts the final operation only if the first op completed through a value channel, which I think is different to `libunifex`, so beware. I thought the original behaviour can be achieved through the marvelous `materialize`.

Unfortunately, this PR contains two little more things: I found a bug in `materialize` / `dematerialize`. I also added `noexcept` specifier to the `then_t` operation.

I did this to experiment, whether the suggestion in issue #472 is viable and I would like to change the `let_xxx` algorithms similarily to how it is done in this implementation. But that would mean to change the spec, too. 

Any feedback welcome

### DIFF
--- a/include/exec/finally.hpp
+++ b/include/exec/finally.hpp
@@ -19,7 +19,7 @@
 #include "../stdexec/execution.hpp"
 
 #include "__detail/__manual_lifetime.hpp"
-#include "any_sender_of.hpp"
+#include "materialize.hpp"
 
 namespace exec {
   namespace __finally {
@@ -29,12 +29,32 @@ namespace exec {
     using __nonempty_decayed_tuple = __decayed_tuple<_Arg, _Args...>;
 
     template <class _Sigs>
-    using __result_type_ =
-      __gather_signal<set_value_t, _Sigs, __q<__nonempty_decayed_tuple>, __q<__variant>>;
+    using __value_types_ = __gather_signal<
+      set_value_t,
+      _Sigs,
+      __mbind_front_q<__decayed_tuple, set_value_t>,
+      __q<__types>>;
 
     template <class _Sigs>
-    using __result_type =
-      __if<std::is_same<__result_type_<_Sigs>, __not_a_variant>, __ignore, __result_type_<_Sigs>>;
+    using __error_types_ = __gather_signal<
+      set_error_t,
+      _Sigs,
+      __mbind_front_q<__decayed_tuple, set_error_t>,
+      __q<__types>>;
+
+    template <class _Sigs>
+    using __stopped_types_ = __gather_signal<
+      set_stopped_t,
+      _Sigs,
+      __mbind_front_q<__decayed_tuple, set_stopped_t>,
+      __q<__types>>;
+
+    template <class _Sigs>
+    using __result_variant = __minvoke<
+      __mconcat<__q<__variant>>,
+      __value_types_<_Sigs>,
+      __error_types_<_Sigs>,
+      __stopped_types_<_Sigs>>;
 
     template <class _ResultType, class _ReceiverId>
     struct __final_operation_base {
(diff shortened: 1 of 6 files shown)


include/exec/finally.hpp
```diff
@@ -0,0 +1,371 @@
...
+         public:
+          template <__none_of<__t, const __t> _Rcvr>
+            requires receiver_of<_Rcvr, completion_signatures<_Sigs...>>
+          __t(_Rcvr& __rcvr) noexcept
+            : __receiver_{std::addressof(__rcvr)}
+            , __env_{get_env(__rcvr)}
+            , __vtable_{__any::__create_vtable(__mtype<__vtable_t>{}, __mtype<_Rcvr>{})} {
+          }
+        };
+      };
```
ericniebler, line 117: I take it this type erasure is necessary because you want to test the nothrow-ness of `connect` before the receiver type is known. The type-erasure is going to lead to worse runtime performance. I don't know whether that's the right trade-off or not.

maikel: Exact. I did it to start a discussion about exactly this. It wouldn't be needed if completion signatures would depend on the receiver and not its environment. Also, an operation is only fully described if a receiver is known. Who knows what else limits the true transparency of signatures

ericniebler:
The nvc++ failure looks like a compiler bug. I could look into it next week. I'm travelling right now.

maikel:
> The nvc++ failure looks like a compiler bug. I could look into it next week. I'm travelling right now.

OK, if I find the time I will try to look, too. Do we keep the type erasure or should I change that?

test/exec/test_finally.cpp
```diff
@@ -0,0 +1,86 @@
...
+    just(21) | then([](int x) {
+      throw 42;
+      return x;
+    }),
+    just() | then([&called]() noexcept { called = true; }));
+  static_assert(__v<__equivalent<
+                  completion_signatures_of_t<decltype(s), empty_env>,
+                  completion_signatures<set_error_t(std::exception_ptr), set_value_t(int&&)>>>);
+  CHECK_THROWS_AS(sync_wait(s), int);
+  CHECK(!called);
```
ericniebler, line 69: This surprises me. I would expect an algorithm called `finally` to _always_ execute the finally action regardless of how the input sender completes.

maikel: Yes, I could implement finally in terms of this algorithm. But I think similar to the when any algorithm you prefer a direct implementation, right?

maikel:
> > The nvc++ failure looks like a compiler bug. I could look into it next week. I'm travelling right now.
> 
> OK, if I find the time I will try to look, too. Do we keep the type erasure or should I change that? 

This algorithm actually does not need virtual dispatch to drop the exception. We could eagerly connect both operation  states before starting the first op. That is a decision which cannot be done in the let algorithms. What do you think?

test/exec/test_finally.cpp
ericniebler: I don't understand what this algorithm is for. If you want to execute an action on a value completion, but otherwise forward the values through, you use an algorithm called `tap`. `finally` means something different. `finally` means always execute the final action.

ericniebler:
> This algorithm actually does not need virtual dispatch to drop the exception. We could eagerly connect both operation states before starting the first op. That is a decision which cannot be done in the let algorithms. What do you think?

If you connect the final action lazily, the two operation states can share space, right? My default is to make things efficient, and use layering to achieve stronger requirements at the expense of size/speed.

So, I would favor lazily starting the finally action so the operation states can share space, and avoiding the virtual dispatch. Then we can find adaptors that let us layer on the other behaviors.

maikel:
> > This algorithm actually does not need virtual dispatch to drop the exception. We could eagerly connect both operation states before starting the first op. That is a decision which cannot be done in the let algorithms. What do you think?
> 
> If you connect the final action lazily, the two operation states can share space, right? My default is to make things efficient, and use layering to achieve stronger requirements at the expense of size/speed.
> 
> So, I would favor lazily starting the finally action so the operation states can share space, and avoiding the virtual dispatch. Then we can find adaptors that let us layer on the other behaviors.

Yes I think they could share space, buts that's currently not happening. I will revisit that part.

test/exec/test_finally.cpp
maikel: Yes you are right. This is very confusing of me doing this. What happened was, that I realized while implementing finalize that I was basically reimplementing materialize within the finally algorithm. I factored out materialize from finally and it gives the current algorithm. The real goal is still finally, tho... Sorry about that

maikel:
I think I addressed all raised concerns. Semantics, no virtual dispatch and reusing storage for the operations.

PR #804, opened 2023-03-07
Implement an io_uring context

maikel:
Hey, this PR just tracks the development of the io context. If the timeout operation is finished I will request a review.

Edit:
The present PR has the goal to implement an `io_uring` based execution context with the following properties:

- its scheduler satisfies a yet-to-be-defined `timed_scheduler` concept, which subsumes the `scheduler` concept. This is done by additionally providing `schedule_after(sched, duration)` and `scheduler_at(sched, time_point)`
- the design is open to extensions for file and network I/O. The base class  `__io_uring::__io_task_base<_Derived>`  requires the following three functions to be defined:
1. `bool ready() const noexcept`: Returns true for eager tasks and false for those who need to be submitted to the io uring submission queue.
2. `void submit(io_uring_sqe& sqe) noexcept`: Fills the submission queue entry, which describes the I/O operation. Will be skipped for tasks that returned true for `ready()`.
3. `void complete(const io_uring_cqe& cqe) noexcept` This will be called after an I/O operation completed. For stopped operations `cqe.res` will contain the value `cqe.res = -ECANCLED`. Its implementation usually notifies a `Receiver`
- This PR contains a `__stoppable_task_base<_Derived, _Receiver>`  base class which uses the `opcode`  `IORING_OP_ASYNC_CANCEL` to automatically implement the stopping of an I/O task which was previously submitted to the `io_uring` submission queue. If `IORING_OP_ASYNC_CANCEL` is not available, then each class that derives from `__stoppable_task_base` is required to provide a `void submit_stop(io_uring_sqe& sqe)` member function, which defines operation-specific steps to stop an already started task. This base class contains a member object of type `__stop_operation` which is a nested async operation. It is started when the parent task needs to be stopped.
- Stopping the whole context is possible by calling `io_uring_context::request_stop()`. The following behaviour will happen:
1. Each new submissions through `io_uring_context::submit(__task*)` will be stopped before reaching the submission queue unless it sets the opcode `IORING_OP_ASYNC_CANCEL`, i.e. unless the task is a cancelation op.
2. When the io uring submission and completion queues are empty and all in-flight operations have been completed, new submissions will complete inline with `set_stopped`. Note: an initially constructed `io_uring_context` behaves differently.


maikel:
Hey, I think the PR is quite big already. The branch is ready to be played with. If anyone wants to review, feel free. Currently supported are the normal `schedule()` and a `schedule_after(scheduler, duration)`. Both operations are stoppable.

include/exec/linux/io_uring_context.hpp
```diff
@@ -0,0 +1,231 @@
...
+  struct schedule_after_t {
+    template <class _Scheduler, class _Duration>
+      requires stdexec::tag_invocable<schedule_after_t, const _Scheduler&, _Duration>
+    auto operator()(const _Scheduler& __scheduler, _Duration __duration) const
+      -> stdexec::tag_invoke_result_t<schedule_after_t, const _Scheduler&, _Duration> {
+      return tag_invoke(*this, __scheduler, __duration);
+    }
+  };
+
+  inline constexpr schedule_after_t schedule_after{};
```
ericniebler, line 61: `schedule_after` is part of a `timed_scheduler` concept, which subsumes `scheduler`. We should factor this out into a `exec/timed_scheduler.hpp` header and define the concept there, since we'll have many more models of it and algorithms that will need to be constrained to accept only `timed_scheduler`s.

maikel: Indeed, I wasn't sure where to put it and needed any CPO for now. I will do as you propose

examples/io_uring.cpp
```diff
@@ -0,0 +1,73 @@
+#include "exec/linux/io_uring_context.hpp"
```
ericniebler, line 1: Needs a license & copyright header.

maikel: Nod

maikel:
> Anything else you'd like to do to this PR?

Indeed. I would like to add a unit test file that conditionally runs if the Linux headers are found. I would like to add a github action based on container with newer Linux header files.

Also, the scheduler misses an equality comparison.

Is that okay?

maikel:
> > Anything else you'd like to do to this PR?
> 
> Indeed. I would like to add a unit test file that conditionally runs if the Linux headers are found. I would like to add a github action based on a container image with newer Linux header files.
> 
> Also, the scheduler misses an equality comparison.
> 
> Is that okay? 

I'm also not sure how the timerfd implementation behaves for zero or negative duration values. So I simply need to add some useful tests

maikel:
OK, I have added a GCC-11 action on Ubuntu-22.04 which supports more `io_uring` opcodes, yay! I've also enabled thread santizier and disabled TBB to not trigger false positives (it is not compiled with TSAN enabled).

If we want a similar test for Clang-12, I couldnt find a Ubuntu-22.04 image. Maybe @trxcllnt could help

ericniebler:
This is an exciting piece of work. Any reason why I shouldn't merge it?

maikel:
> This is an exciting piece of work. Any reason why I shouldn't merge it?

I am aware of at least one more deficit: AFAIK, the completion queue can "overflow" becaues I am not limiting the *total* number of submissions. The kernel consumes submission entries faster than it completes them by taking an internal copy of them. The completion queue is by default twice as big as the submission queue but it in extreme siutations its still the userlands responsibilty to prevent such an "overflow". Its simple to implement *something*. I wanted to check whats done in practise. I haven't done so, yet.

maikel:
I also think the context is clumsy to use, if not wrongly named. Maybe we call it `io_uring_run_loop` and make an RAII wrapper for it `io_uring_context` which owns a thread and requests a stop in its destructor.

maikel:
We can merge this if you want to. But we should keep the interface in flow with better alternatives. I got mulltiple suggestions from @lewissbaker and @kirkshoop .

include/exec/timed_scheduler.hpp
```diff
@@ -0,0 +1,84 @@
...
+  template <class _Scheduler, class _TimePoint>
+  concept __has_schedule_at = requires(_Scheduler&& __sched, const _TimePoint& __time_point) {
+    { schedule_at((_Scheduler&&) __sched, __time_point) } -> stdexec::sender;
+  };
+
+  // TODO: Add more requirements such as __has_schedule_at or __has_now
+  template <class _Scheduler, class _Clock = std::chrono::system_clock>
+  concept timed_scheduler = /* std::chrono::is_clock_v<_Clock> && */
+    stdexec::scheduler<_Scheduler> && __has_schedule_after<_Scheduler, typename _Clock::duration>
+    && __has_schedule_at<_Scheduler, typename _Clock::time_point>;
```
kirkshoop, line 80: `timed_scheduler` needs a `now() -> time_point` cpo.

`now()` on the `timed_scheduler` allows a `virtual_time_scheduler` to be used in tests to do tests of seconds of elapsed time in ms.

virtual-time is also used in log processing and audio, video, and animations

kirkshoop (approved):
This is great!

include/exec/timed_scheduler.hpp
maikel: OK, I will add that cpo asap

ericniebler:
>  I got mulltiple suggestions from @lewissbaker and @kirkshoop .

Oh, where has that discussion been happening?

include/exec/timed_scheduler.hpp
ericniebler: It's done now.

PR #815, opened 2023-03-13
Improvements for io_context

maikel:
Hi,

ideally I wish for an `driveable_context` concept which we could use a drop-in replacement for `run_loop` in the `sync_wait` algorithm. Until we decide to change `stdexec::sync_wait` I propose the following extension to `io_uring_context`:

`io_uring_context::sync_wait(sender)` runs the context until the single value input sender completes and returns its result.

The sync wait algorithm requires of `stdexec::run_loop` a way to start the loop with `run` and break from the loop, currently via the method `stdexec::run_loop::finish()`.

Analog to `stdexec::run_loop` this PR includes the member function `io_uring_context::finish()` which breaks out its run loop without stopping the context. ~~As a consequence this also means that the destructor of `io_uring_context` will be potentially blocking if the context has pending operations.~~

A discussion with @kirkshoop on Discord also revealed that *composable* contexts would need some kind `run_some()` function that would do progress in a non-blocking fahsion. I did so, but if thats too much in this PR, let me know.


include/exec/linux/__detail/io_uring_context.hpp
```diff
@@ -234,6 +223,17 @@ namespace exec { namespace __io_uring {
...
+  }
+
+  inline __context::~__context() {
+    // It is a fatal error if the context is destroyed while it is still in a running state.
+    STDEXEC_ASSERT(!__is_running_.load(std::memory_order_relaxed));
+    if (__n_submissions_in_flight_.load(std::memory_order_relaxed) != __no_new_submissions) {
+      request_stop();
+      __break_loop_.store(false, std::memory_order_release);
+      run();
+    }
```
ericniebler, line 236: I know @kirkshoop has strong feelings about blocking in destructors. Kirk, do you want to chime in here?

ericniebler (commented):
How is this `sync_wait` member different than doing `stdexec::sync_wait(on(iouring_sched, snd))`? Is the idea that this is lighter weight and will perform better?

Long term, we don't want a context with a `sync_wait` member function. We want the `this_thread::sync_wait` algorithm to be customized to do the right thing. I don't think we're currently at a place where we can ensure this always happens when it should, but we will eventually.

Maybe we can add it temporarily, but have it issue a deprecation warning by default that can be silenced with a flag.

maikel:
> How is this `sync_wait` member different than doing `stdexec::sync_wait(on(iouring_sched, snd))`? Is the idea that this is lighter weight and will perform better?
> 
> Long term, we don't want a context with a `sync_wait` member function. We want the `this_thread::sync_wait` algorithm to be customized to do the right thing. I don't think we're currently at a place where we can ensure this always happens when it should, but we will eventually.
> 
> Maybe we can add it temporarily, but have it issue a deprecation warning by default that can be silenced with a flag.

For `stdexec::sync_wait(on(iouring_sched, snd))` to work, I would need to specialize the `sync_wait_t` CPO, correct? I don't want this for every sender with this `completion_scheduler<set_value_t>`. Even if we would go that way, this  PR contains logic to break out from the run loop without shutting down all ongoing I/O operations.

I want this a way to tell `sync_wait` which "DriveableExecutionLoop" it should use instead of `run_loop` independent of the senders completion scheduler. AFAIK thats not possible with the current `sync_wait`. This PR is meant as drop-in replacement for this.

I prefer

```cpp
  io_uring_context context;
  sync_wait(context, some-sender);
```

over

```cpp
  io_uring_context context;
  context.sync_wait(some-sender);
```

I don't want

```cpp
  io_uring_context context;
  std::jthread t{[&] { context.run(); });
  { 
    scope_guard stop{[&] { context.request_stop(); }};
    sync_wait(some-sender);
  }
```

include/exec/linux/__detail/io_uring_context.hpp
maikel: > I know @kirkshoop has strong feelings about blocking in destructors. Kirk, do you want to chime in here?

Yes. You are right, blocking within the destructor is bad and this is a consequence of unconditionally breaking from the run loop. I could change the semantics to "stop executing the run loop if no I/O operations are ongoing instead of waiting for more ops to arrive".

maikel: Done

maikel:
I have remove the `exec/linux/__detail/io_uring_context.hpp` and implemented `io_uring` related things only in `exec/linux/io_uring_context.hpp`. Hidden friends are defined in-situ now.

Pending TODO is the `sync_wait` adaption. I will ping back if I've come up with a solution using queries and an environment.

include/stdexec/execution.hpp
```diff
@@ -5767,29 +5767,40 @@ namespace stdexec {
...
+
     template <class _Sender>
     using __into_variant_result_t = decltype(stdexec::into_variant(__declval<_Sender>()));
 
-    struct __env {
-      using __t = __env;
-      using __id = __env;
-      stdexec::run_loop::__scheduler __sched_;
+    template <class _Context>
+    struct __base_env {
```
ericniebler, line 5781: It's not a base class, so I would prefer the name `__basic_env`.

include/stdexec/execution.hpp
```diff
@@ -5879,18 +5891,21 @@ namespace stdexec {
...
-      template <__single_value_variant_sender<__env> _Sender>
+      template <
+        __sync_wait::__driveable_context _Context,
+        __single_value_variant_sender<__env> _Sender>
         requires(!__tag_invocable_with_completion_scheduler< sync_wait_t, set_value_t, _Sender>)
              && (!tag_invocable<sync_wait_t, _Sender>) && sender_in<_Sender, __env>
-             && sender_to<_Sender, __receiver_t<_Sender>>
-      auto operator()(_Sender&& __sndr) const -> std::optional<__sync_wait_result_t<_Sender>> {
+             && sender_to<_Sender, __receiver_t<_Context, _Sender>>
+      auto operator()(_Context& __loop, _Sender&& __sndr) const
```
ericniebler, line 5900: I was imagining something more along the lines of:

```c++
template <sender _Sender, class _Env = empty_env>
   requires ...
auto operator()(_Sender&& __sndr, _Env&& __env = _Env()) 
```

... and then pull the driveable context out of `__env`, or use `run_loop` if it's not there. Eventually, we'll want to support other queries on the environment like a stop_token or allocator for the operation.

Also, I think `__single_value_variant_sender<__env>` is the wrong constraint. The `__env` type assumes a `run_loop` drivable context, no?

include/stdexec/execution.hpp
```diff
@@ -5767,29 +5767,40 @@ namespace stdexec {
   // [execution.senders.consumers.sync_wait]
   // [execution.senders.consumers.sync_wait_with_variant]
   namespace __sync_wait {
+    template <class _Context>
+    concept __driveable_context = requires(_Context& __ctx) {
+      { __ctx.get_scheduler() } noexcept -> scheduler;
+      { __ctx.run() };
+      { __ctx.finish() };
+    };
```
ericniebler, line 5775: My gut tells me that we don't want to be in the business of specifying the shape for execution contexts. Maybe what we want is a drivable scheduler, which is a refinement of the scheduler concept with `run` and `finish`.

ericniebler (changes requested):

include/stdexec/execution.hpp
maikel: I admit, today I've only changed the implementation of `io_context` to fix the hidden friend issue.

I imagine that the way you do. This was my previous attempt.

Regarding the `__single_value_variant_sender<__env>`, I believe it was wrong before my commit, too. Its a pending TODO. But I can fix that today or tomorrow.

include/stdexec/execution.hpp
maikel: I agree

maikel:
I removed any changes to `sync_wait` from this commit. Now its purely internal improvements over the current `io_uring_implementation`.  This PR contains

- `tag_invoke` overlaods are defined in-situ
- I've refactored `__stoppable_task_base` and `__io_task_base` to `*_facade`s, which simplifies defining stoppable ops. An example for possible `open, close, read` and `write`  ops can be seen https://github.com/maikel/stdexec/blob/ad9f4017c589041207435b2f89113788fb5da17b/include/exec/linux/io_uring_operations.hpp
- I've implemented `run_until_empty` which runs the loop until the work queue is empty and then returns without stopping the context. If the context is a non-stopped state we can still schedule new work on it such that its not being stopped immediately.

ericniebler (approved):

ericniebler:
@maikel can you have a look at the merge conflict?

PR #823, opened 2023-03-15
Use `__builtin_bit_cast` when available for constexpr-ness

ericniebler:


include/exec/__detail/__bit_cast.hpp
```diff
@@ -16,6 +16,8 @@
  */
 #pragma once
 
+#include <cstring>
```
lewissbaker, line 19: Not related to this PR, but I noticed that the fallback implementation of `bit_cast()` is marked constexpr (which matches the std implementation) but is using std::memcpy, which isn't marked constexpr and so won't compile if actually invoked in a constexpr context.

Maybe it should be implemented in terms of the respective compiler intrinsics instead?
e.g. `__builtin_bit_cast()` on clang
Or not marked constexpr?

ericniebler: I've left the `constexpr` but changed the implementation to use `__builtin_bit_cast` when available.

include/exec/__detail/__bit_cast.hpp
```diff
@@ -37,9 +44,13 @@ namespace exec {
       std::is_trivially_copyable_v<_From>, "bit_cast requires _From to be trivially copyable");
     static_assert(
```
lewissbaker, line 45: Minor nit: the standard version has constraints for these requirements instead of mandates - perhaps so it's SFINAE-friendly?

include/exec/__detail/__bit_cast.hpp
```diff
@@ -24,6 +24,13 @@
 #endif
 
 #include <cstring>
+#include <type_traits>
+
+#ifdef __has_builtin
+#define STDEXEC_HAS_BUILTIN __has_builtin
+#else
+#define STDEXEC_HAS_BUILTIN(...) 0
+#endif
```
lewissbaker, line 33: Does this belong in `include/stdexec/__detail/__config.hpp`?

lewissbaker (approved):

PR #827, opened 2023-03-16
Build io_uring example on Linux only

bustercopley:
Only attempt to build the io_uring example when building on Linux.


ericniebler:
Thanks!

PR #829, opened 2023-03-16
Add CI job for GCC on MSYS2 on Windows

bustercopley:


ericniebler:
Hi @bustercopley. I wouldn't mind adding msys2 to the test matrix, but since I can't commit to maintaining it, I don't want a CI failure in msys2 to cause the entire CI job to get a big red X. Travis lets you tag a job with `allow-failure` for this, but it looks like GH Actions has been slow to implement this feature. See https://github.com/orgs/community/discussions/15452. (GH Actions `continue-on-error` tag is not what's needed here.)

Without that feature, my preference is to keep msys out of our CI matrix. I'm open to other suggestions.

bustercopley:
Hi @ericniebler,
That's understandable. If there's progress on the GH Actions feature I might bring this up again. Thanks!

Issue #825, opened 2023-03-15
Improve compiler diagnostics for when a sender & receiver won't connect


ericniebler:
I think stdexec can absolutely have good diagnostics, but we're a long way from that. I've done a bit of that with the `receiver_of` concept, which gives good diagnostics when it fails. (It tells you which completion signature the receiver lacks a handler for. See [here](https://github.com/NVIDIA/stdexec/blob/bbe72866bcbd24bc9ce572591d882a567768826b/include/stdexec/execution.hpp#L577-L638).)

I'm thinking it should be possible to retrofit this sort of type-based error propagation into, say, `get_completion_signatures` and `make_completion_signatures`, so that users get a sensible diagnostic when a sender and receiver don't connect. And then change the routines like `sync_wait` to surface the error information when the requirements aren't satisfied.

lewissbaker:
Duplicate of #569 ?

PR #835, opened 2023-03-19
[WIP] Improve diagnostics if sync_wait fails due to errors in get_completion_signatures

maikel:
Addresses #825

### DIFF
--- a/include/stdexec/execution.hpp
+++ b/include/stdexec/execution.hpp
@@ -65,8 +65,6 @@ STDEXEC_PRAGMA_IGNORE("-Wundefined-inline")
 STDEXEC_PRAGMA_IGNORE("-Wundefined-internal")
 
 namespace stdexec {
-  using std::remove_cvref_t;
-
   // [exec.queries.queryable]
   template <class T>
   concept queryable = destructible<T>;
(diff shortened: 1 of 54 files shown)


trxcllnt:
ok to test

maikel:
I've had some success with passing substitution errors from `gather_signal` up to `make_completion_signatures` or `get_completion_signatures`. This is still highly experimental and I need to further refine the error reporting. Im not happy yet.

maikel:
The current state catches errors with `then` and `bulk` but not with `let_value`. The following snippet won't have useful diagnostics with `sync_wait`.

```cpp
just(42) | let_value([](int*) {  return just(0); });
```

This is, because the substitution error in `let_value` happens when computing the resulting sender before passing that result into the `make_completion_signatures` template.

include/stdexec/execution.hpp
```diff
@@ -413,6 +413,13 @@ namespace stdexec {
   using __as_awaitable::as_awaitable_t;
   extern const as_awaitable_t as_awaitable;
 
+  template <class _Fn, class... _Args>
+  struct _ERRONEOUS_SUBSTITUTION_ : std::false_type {
+    friend auto operator,(const _ERRONEOUS_SUBSTITUTION_&, auto) -> _ERRONEOUS_SUBSTITUTION_ {
+      return {};
+    }
```
ericniebler, line 420: It occurs to me that we can save some template instantiations here because this friend function doesn't need to be a template. See https://godbolt.org/z/M9599dqM4

ericniebler:
> The current state catches errors with `then` and `bulk` but not with `let_value`. [...] This is, because the substitution error in `let_value` happens when computing the resulting sender before passing that result into the `make_completion_signatures` template.

This is one of the reasons why I think the correct approach is to bake error propagation into the __meta utilties, like #869 does. The approach in this PR is going to get unwieldy pretty quick.

I'm going to close out this PR. We can build on #869.

PR #846, opened 2023-03-21
Start refactoring CUDA error handling:

miscco:
This is a breakout from @brycelelbach async branch, to keep it small

* Add `nvexec::cuda_exception`.
* Change the `cudaError_t` checking macro/function to throw if `STDEXEC_THROW_ON_CUDA_ERROR` (or `_DEBUG`/`DEBUG`) is defined or print if `STDEXEC_PRINT_ON_CUDA_ERROR` is defined.
* Ensure that all CUDA runtime API calls are checked.


include/nvexec/multi_gpu_context.cuh
```diff
@@ -44,11 +44,11 @@ namespace nvexec {
         template <stdexec::__decays_to<R> Receiver>
         operation_state_t(Receiver&& rec)
           : rec_((Receiver&&) rec) {
-          status_ = STDEXEC_DBG_ERR(cudaStreamCreate(&stream_));
+          status_ = STDEXEC_CHECK_CUDA_ERROR(cudaStreamCreate(&stream_));
```
senior-zero, line 47: Code around used to handle the error. It would seem that you are throwing an exception immediately instead. Most of these functions are used inside noexcept context. I'd prefer checking the error and completing with a `cudaError` instead of throwing an exception when possible.

miscco: I dont think so, the main difference now is that we have the option to select the error handling
```cpp
  // If the `cudaError_t` parameter represents an error:
  // * If `STDEXEC_THROW_ON_CUDA_ERROR` is defined, an exception is thrown.
  // * If `STDEXEC_PRINT_ON_CUDA_ERROR` is defined, a diagnostic is printed to stdout.
  // * Otherwise, nothing happens.
```

By default we are not doing anything

ericniebler: I agree with Georgii. Throwing here is never correct. Ditto for most of stdexec.

brycelelbach:
For debugging purposes, it's important to have a mode that reports errors synchronously where they occur, so that when you run it in a debugger, you get the actual location where the error occurred. It's just a mode for debugging, don't use it if you don't want to.

Anytime any CUDA runtime API is called, we *MUST* check the error code returned by it and report the error in some fashion.

miscco:
Yeah, I discussed it with georgii and we agreed that throwing is not the right thing to do. That said I can also sympathize with the need to having a proper error reporting in place if one needs to debug. I will close that PR but keep the branch around

PR #848, opened 2023-03-21
[NFC] Do not use single letter names

miscco:
There are platforms, where single letter macros are abundant *cough MSVC*

While it is generally not supported yet, there is no need to use those names when we can add a letter and be fine


ericniebler:
Thank you!

PR #850, opened 2023-03-23
add `exec::trampoline_scheduler` to bound stack usage

ericniebler:
Used like:

```c++
  exec::trampoline_scheduler sched;
  auto recurse_deeply = retry(exec::on(sched, fails_alot{}));
  sync_wait(std::move(recurse_deeply));
```

### DIFF
--- a/include/exec/trampoline_scheduler.hpp
+++ b/include/exec/trampoline_scheduler.hpp
@@ -1,0 +1,186 @@
+/*
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ * Copyright (c) 2023 NVIDIA Corporation
+ *
+ * Licensed under the Apache License Version 2.0 with LLVM Exceptions
+ * (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ *   https://llvm.org/LICENSE.txt
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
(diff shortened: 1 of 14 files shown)


include/exec/trampoline_scheduler.hpp
```diff
@@ -0,0 +1,186 @@
...
+            __self.__execute();
+            __state.__drain();
+          } else if (__current_state->__recursion_depth_ < __self.__max_recursion_depth_) {
+            ++__current_state->__recursion_depth_;
+            __self.__execute();
+          } else {
+            // Exceeded recursion limit.
+            __self.__next_ = std::exchange(
+              __current_state->__head_, static_cast<__operation_base*>(&__self));
+          }
```
ericniebler, line 85: I admit I have shamelessly copied this implementation from libunifex (and left the Facebook copyright!). Now that I'm looking at this `start` implementation, I have Questions.

If we call `start` on one of these operation states and we've hit the recursion limit, we stuff the operation into the head of a thread-local list. The next time a trampoline operation is started on this thread, we execute all the operations in the list. Fine.

But what if we _don't_ start another trampoline operation on this thread? What becomes of the started operations sitting in the list? Program correctness could demand that this operations complete. If they don't, they can hang the program.

Am I confused?

ericniebler: attn @kirkshoop @lewissbaker

ericniebler: Maybe when the recursion limit is hit, we should post the work to the delegatee scheduler? And hard-error if there isn't one?

lewissbaker: When the recursion limit is reached it posts work back to the first call to start (the first if-branch where it calls drain()).

Each time execution returns to drain() loop it resets the recursion count back to 1 before executing the next item in the queue.

PR #861, opened 2023-03-31
Sequence senders

maikel:
Hi,

this contains a possible design for sequence senders. I hope this is what you have envisioned @kirkshoop @lewissbaker 

Basic ideas, how I understood them

- sequence receivers `Receiver` specialize an additional `set_next(Receiver& self, sender auto&& item) -> sender` CPO
- senders can specialize `connect` or `sequence_connect`
- sender types that specialize `sequence_connect` are sequence senders
- completion signatures of sequence senders describe the completions of the items passed to `set_next`

Included in this PR are 

- concepts: `sequence_receiver_of, sequence_receiver_from, sequence_sender_to`
- few basic generators:
    - `empty_sender() -> sequence-sender`: Returns an empty sequence
    - `once(sender) -> sequence-sender`: Transforms a sender into a sequence-sender. The returned sequence sender completes after the sender have been emitted once.
    - `repeat(sender) -> sequence-sender`: Repeats a given sender endlessly until stopped.
    - `generate_each(fun) -> sequence-sender`: Repeatedly invokes a given function `fun` and emits a sender with its result
- few basic sinks:
    - `ignore_all(sequence-sender) -> sender_of<set_value_t()>`: Ignores all sequence elements and completes whenever the sequence completes
    - `first_value(sequence-sender) -> sender`: Stops the sequence after the first value arrived and completes with the first value upon completion.
    - `last_value(sequence-sender) -> sender` Returns a sender with the last value of a sequence. If any error or stop signal occurs, the sequence will complete with that instead.
    - `reduce(sequence-sender, init, fun) -> sender`: equivalent to `last_value ∘ scan`
- few basic adaptors
    - `scan(sequence-sender, init, fun) -> sequence-sender`: completes with `fun(Tx, Ts...) -> Tx` for each element-wise completion with values `Ts...`
    - `take_while(sequence-sender, predicate) -> sequence-sender`: Forwards all sequence elements until the predicate returns false for the emitted values of a sequence element
    - `filter_each(sequence-sender, predicate) -> sequence-sender`: Drops elements from the sequence for which the predicate returns false.
    - `enumerate_each(sequence-sender) -> sequence-sender`: Transforms input `(args...)` to `(counter, args....)` where counter is an atomic integer that gets increased for each element
    - `merge_each_sequence(sequence-sender-of-sequence-sender) -> sequence-sender`: Returns a sequence-sender that forwards any element completions of all element-wise sequence senders
    - `transform_each(sequence-sender, (sender -> sender)) -> sequence-sender`: Applies a sender adaptor `sender -> sender` to each item that is emitted from a sequence sender.
        - `let_[value,stopped,error]_each(seqeuence-sender) -> sequence-sender`: special case of `transform_each`
        - `then_each(sequence-sender) -> sequence-sender`: special case of `transform_each`
    - `zip(sequence-senders, ...) -> sequence-sender`: Similar to `when_all` for each generation of sequence elements


include/exec/sequence_senders.hpp
```diff
@@ -0,0 +1,777 @@
...
+      }
+
+      constexpr auto operator()() const noexcept -> __binder_back<join_all_t> {
+        return {};
+      }
+    };
+  } // namespace __join_all
+
+  using __join_all::join_all_t;
+  inline constexpr join_all_t join_all;
```
kirkshoop, line 776: FYI I called this ignore_all. it does not join in the async sense..

kirkshoop (approved):
I love it!

include/exec/sequence_senders.hpp
maikel: Yup, that's better, thank you! I didn't know how to name that

include/exec/sequence_senders.hpp
```diff
@@ -0,0 +1,777 @@
...
+#include "../stdexec/execution.hpp"
+
+#include "./trampoline_scheduler.hpp"
+
+namespace exec {
+  namespace __sequence_sender {
+    using namespace stdexec;
+
+    struct set_next_t {
+      template <class _Receiver, class _Item>
```
ericniebler, line 27:
```suggestion
      template <receiver _Receiver, sender _Item>
```

?

include/exec/sequence_senders.hpp
```diff
@@ -0,0 +1,777 @@
...
+    template <class _Receiver, class Error>
+    auto __has_sequence_signature(set_error_t (*)(Error))
+      -> __receiver_concepts::__missing_completion_signal_t<_Receiver, set_error_t, Error>;
+
+    template <class _Receiver, class... _Sigs>
+    auto __has_sequence_signatures(completion_signatures<_Sigs...>*)
+      -> decltype((__has_sequence_signature<_Receiver>(static_cast<_Sigs*>(nullptr)), ...));
+
+    template <class _Signatures, class _Receiver>
+    concept is_valid_next_completions = _Signatures::template _WITH_RECEIVER_<_Receiver>::value;
```
ericniebler, line 91:
```suggestion
    concept __is_valid_next_completions = _Signatures::template _WITH_RECEIVER_<_Receiver>::value;
```

include/exec/sequence_senders.hpp
```diff
@@ -0,0 +1,777 @@
...
+
+      template <class _Rcvr>
+      using __next_sender = __call_result_t<set_next_t, __decay_t<_Rcvr>&, _Source>;
+
+      template <class _Rcvr>
+      using __next_on_scheduler_sender =
+        __call_result_t<on_t, exec::trampoline_scheduler, __next_sender<_Rcvr>&&>;
+
+      template <class _Rcvr>
+      using __recveiver = typename __operation<__id<_Source>, __id<__decay_t<_Rcvr>>>::__receiver;
```
ericniebler, line 250: I've found this to be a problematic pattern. You can't compute the receiver type without instantiating the operation state type. But you're using the receiver type below in the constraints of `connect`. This means that, rather than SFINAE away, an invalid call to `connect` will cause a hard error somewhere in the guts of the operation state.

It's usually best to define the receiver type independent of the operation state. And in most cases it doesn't need to depend on the sender type at all.

include/exec/sequence_senders.hpp
```diff
@@ -0,0 +1,777 @@
...
+
+      struct __t;
+
+      struct __receiver {
+        __t* __op_;
+
+        template <class... _Args>
+        friend void tag_invoke(set_value_t, __receiver&& __self, _Args&&...) noexcept {
+          __self.__op_->repeat();
+        }
```
ericniebler, line 170: It seems a little strange to me that we would just drop these results on the floor. @kirkshoop thoughts?

ericniebler (commented):
Very pretty! Just a few minor comments. Also, I feel like we may want to move each algorithms into its own header. I'm currently on the fence if we want to put all the sequence algorithm headers in a `exec/sequence/` subdirectory.

include/exec/sequence_senders.hpp
maikel: Good catch. I will test and refactor that

maikel: In case of repeat I found it difficult to not depend on the sender type. I will think about it again.

maikel:
> Very pretty! Just a few minor comments. Also, I feel like we may want to move each algorithms into its own header. I'm currently on the fence if we want to put all the sequence algorithm headers in a `exec/sequence/` subdirectory.

Yes, we can split that in multiple files. Most sequence sender adaptors should be cpos, too, right? I wonder if all adaptors that are a trivial transforms  (let, then, upon...) should be a in one file, but I don't really care. Algorithms such as zip are different beasts

include/exec/sequence_senders.hpp
maikel: I've noodled over this question and this turned out to be deeper than I thought.

1. I think you are correct. Its iffy.
2. `completion_signatures` for sequence senders describe the signatures that are passed *to* the `set_next` CPO of a sequence receiver. I believe we should additionally require that the sender *returned by* `set_next` should be a single value typed `sender_of<set_value_t()>`. This would be an additional requirement to `set_next`.

include/exec/sequence/first_value.hpp
```diff
@@ -0,0 +1,361 @@
...
+    struct __operation_base {
+      enum __state {
+        __empty,
+        __emplace_in_progress,
+        __emplaced
+      };
+
+      using _Receiver = __t<_ReceiverId>;
+      using __on_stop =
+        stop_token_of_t<env_of_t<_Receiver&>>::template callback_type<__on_stop_requested>;
```
ericniebler, line 46:
```suggestion
        typename stop_token_of_t<env_of_t<_Receiver&>>::template callback_type<__on_stop_requested>;
```

maikel:
I will split this PR in smaller parts.

PR #870, opened 2023-04-04
Check that bulk callables can take values by reference

msimberg:
This was already done for the default implementation of `bulk`, but not for the `static_thread_pool` and `stream_context` customizations.


ericniebler:
Thanks!

PR #895, opened 2023-04-18
Modify requirement `ensure_started` to accept move only input sender

aurianer:
In the same way as #748 does it for `split`.
Change the environment test to use a non empty_env.

### DIFF
--- a/include/stdexec/execution.hpp
+++ b/include/stdexec/execution.hpp
@@ -4029,8 +4029,7 @@ namespace stdexec {
         using is_sender = void;
 
         explicit __t(_CvrefSender __sndr, _Env __env)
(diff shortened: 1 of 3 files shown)


include/stdexec/execution.hpp
```diff
@@ -4123,12 +4117,11 @@ namespace stdexec {
 
     template <class _Sender>
     concept __ensure_started_sender = //
-      requires(typename _Sender::__id __sndr1) { __test_ensure_started_sender(__sndr1); };
+      requires(typename _Sender::__cvref_id __sndr1) { __test_ensure_started_sender(__sndr1); };
```
ericniebler, line 4120: This change looks wrong to me.

ericniebler (approved):

PR #898, opened 2023-04-21
Refactor `nvexec::reduce` implementation

miscco:
We plan to expand the number of supported algorithms.

A lot of the necessary infrastructure will be reusable between similar algorithms, with the main exception being the `set_value` implementation.

This PR takes a first step by abstracting most of the boilerplate into a crtp base class, so that we can hide most of the complexity there.


include/nvexec/stream/reduce.cuh
```diff
@@ -17,221 +17,103 @@
...
-        }
+          do {
+            if (status = STDEXEC_DBG_ERR(cub::DeviceReduce::Reduce(
+                  d_temp_storage,
+                  temp_storage_size,
+                  first,
+                  d_out,
+                  num_items,
+                  self.fun_,
+                  value_t{},
```
senior-zero, line 65: I guess non-default binary operator doesn't make a lot of sense without an ability to pass initial value. Also, in standard reduce / fold algorithms, there's `reduce(..., init)` and `reduce(..., init, binary_op)` but no `reduce(..., binary_op)`. Do you think you could fix that while we are here?

include/nvexec/stream/algorithm_base.cuh
```diff
@@ -0,0 +1,144 @@
...
+                requires stdexec::
+                receiver_of<Receiver, completion_signatures<Self, stdexec::env_of_t<Receiver>>>
+                friend auto
+                tag_invoke(stdexec::connect_t, Self&& self, Receiver&& rcvr) -> stream_op_state_t<
+                stdexec::__copy_cvref_t<Self, Sender>,
+                receiver_t<Receiver>,
+                Receiver> {
+                return stream_op_state<stdexec::__copy_cvref_t<Self, Sender>>(
+                ((Self&&) self).sndr_,
+                (Receiver&&) rcvr,
```
senior-zero, line 123: Something seems to be wrong with the formatting.

senior-zero (approved):

include/nvexec/stream/algorithm_base.cuh
miscco: Thanks a lot, I had it disabled but reenabled it :+1:

include/nvexec/stream/reduce.cuh
miscco: Done

include/nvexec/stream/algorithm_base.cuh
```diff
@@ -0,0 +1,149 @@
...
+          stdexec::__copy_cvref_t<Self, Sender>,
+          Env,
+          stdexec::completion_signatures<stdexec::set_error_t(cudaError_t)>,
+          set_value_t >;
+
+      template <stdexec::__decays_to<__t> Self, stdexec::receiver Receiver>
+        requires stdexec::
+                   receiver_of<Receiver, completion_signatures<Self, stdexec::env_of_t<Receiver>>>
+      friend auto tag_invoke(stdexec::connect_t, Self&& self, Receiver&& rcvr)
+        -> stream_op_state_t< stdexec::__copy_cvref_t<Self, Sender>, receiver_t<Receiver>, Receiver> {
```
ericniebler, line 123:
```suggestion
      friend auto tag_invoke(stdexec::connect_t, Self&& self, Receiver rcvr)
        -> stream_op_state_t< stdexec::__copy_cvref_t<Self, Sender>, receiver_t<Receiver>, Receiver> {
```

Either that or else you need to decay the type of the receiver.

ericniebler (changes requested):

ericniebler (approved):

ericniebler:
Thanks!

PR #907, opened 2023-04-22
reorganize the forward declarations and the utilities

ericniebler:


include/stdexec/__detail/__concepts.hpp
```diff
@@ -0,0 +1,103 @@
...
+#if STDEXEC_CLANG()
+
+  template <class _Ap, class _Bp>
+  concept __same_as = __is_same(_Ap, _Bp);
+
+#elif STDEXEC_GCC()
+
+  template <class _Ap, class _Bp>
+  concept __same_as = __is_same_as(_Ap, _Bp);
+
```
miscco, line 53: Fyi gcc also supports `__is_same` since gcc-10

PR #910, opened 2023-04-24
Implement views as senders

miscco:
This explores turning std::views into senders by providing appropriate sender / receiver 

It is essentially a rip off from the `then` algorithm but instead of a function we store a closure object that is then piped on the passed in range

### DIFF
--- a/include/stdexec/ranges.hpp
+++ b/include/stdexec/ranges.hpp
@@ -189,6 +189,19 @@ namespace stdexec {
     using __transform::transform_t;
     inline constexpr transform_t transform{};
 
+    namespace __filter {
+      inline constexpr __mstring __filter_context =
+        "In stdexec::views::filter(Sender, Function)..."__csz;
+
+      struct filter_t
+        : public __view_adaptor::
+            __pipe< __callable_error<__filter_context>, decltype(::std::views::filter), filter_t> {
+      };
+    } // __filter
+
+    using __filter::filter_t;
+    inline constexpr filter_t filter{};
+
   } // namespace views
 
 } // namespace stdexec
--- a/include/stdexec/ranges.hpp
+++ b/include/stdexec/ranges.hpp
@@ -1,0 +1,196 @@
+/*
+ * Copyright (c) 2021-2022 NVIDIA Corporation
+ *
+ * Licensed under the Apache License Version 2.0 with LLVM Exceptions
+ * (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ *   https://llvm.org/LICENSE.txt
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#pragma once
+
+#include <cassert>
+#include <concepts>
+
+#ifdef __cpp_lib_ranges
+#include <ranges>
+
+#include "__detail/__intrusive_ptr.hpp"
+#include "__detail/__meta.hpp"
+#include "__detail/__scope.hpp"
+#include "functional.hpp"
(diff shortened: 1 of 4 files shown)


include/stdexec/ranges.hpp
```diff
@@ -0,0 +1,244 @@
...
+#include "__detail/__scope.hpp"
+#include "functional.hpp"
+#include "concepts.hpp"
+#include "coroutine.hpp"
+#include "stop_token.hpp"
+
+namespace stdexec {
+  /////////////////////////////////////////////////////////////////////////////
+  // [execution.senders.adaptors.then]
+  namespace __closure_adaptor {
```
miscco, line 35: I am not happy with that name. Maybe `__views` would be appropriate

include/stdexec/ranges.hpp
```diff
@@ -0,0 +1,244 @@
...
+      }
+
+      template <sender _Sender, __movable_value _Fun>
+        requires(!__tag_invocable_with_completion_scheduler<
+                  DerivedClosure,
+                  set_value_t,
+                  _Sender,
+                  _Fun>)
+             && tag_invocable<DerivedClosure, _Sender, _Fun>
+      sender auto operator()(_Sender&& __sndr, _Fun __fun) const
```
miscco, line 193: I think this can be further generalized. 

We should accept anything that can be passed to the respective `std::view` and returns not a viewable_range, e.g. we would accept `transform(fun)` but not `transform(range, fun)`

ericniebler (changes requested):
I think this can literally be just `stdexec::then` combined with the standard views. E.g.,

```c++
namespace exec::views {
  inline constexpr struct transform_t {

    template <class Fun>
    constexpr auto operator()(Fun fun) const {
      return stdexec::then(std::views::transform((Fun&&) fun));
    }

    template <stdexec::sender Sender, class Fun>
    constexpr auto operator()(Sender&& sndr, Fun fun) const {
      return stdexec::then((Sender&&) sndr, std::views::transform((Fun&&) fun));
    }

  } transform {};
}
```

Considering how simple it is for users to do this themselves, I'm not convinced we need this.

I also wonder about use cases for this. I _know_ we want filters and transforms over _asynchronous_ sequences. When would I push a synchronous range through the value channel and apply a range adaptor to it?

miscco:
> I also wonder about use cases for this. I _know_ we want filters and transforms over _asynchronous_ sequences. When would I push a synchronous range through the value channel and apply a range adaptor to it?

putting my NVIDIA hat on, I would wager that we will rarely get anything that is not a (sub)range.

At the end you will want to perform a reduce or a parallel for_each to leverage the parallelism of any accelerator?

maikel:
We have `exec::iterate` and PR #991 introduces `exec::transform_each` which could be used to apply a sender to each element of a standard view. Its planned to add `exec::then_each`, `exec::let_value_each` etc as aliases to using `stdexec::then` and `stdexec::let_value` etc.

i.e. 

```cpp
auto do_something = exec::iterate(std::views::iota(0, 100)))
                  | exec::transform_each(stdexec::then([](int i) { do_work(i); }))
                  | exec::ignore_all_values();
stdexec::sync_wait(do_something);
```

Domains that customize `exec::iterate` could parallelize the iteration. The above would be a fancy way of defining the `bulk` algorithm.

ericniebler:
@maikel writes:
> We have `exec::iterate` and PR #991 introduces `exec::transform_each` which could be used to apply a sender to each element of a standard view. Its planned to add `exec::then_each`, `exec::let_value_each` etc as aliases to using `stdexec::then` and `stdexec::let_value` etc.

@maikel what do you think the implications are for this PR? Are you saying it will be superfluous?

maikel:
I am not totally sure about the goal of the PR. I don't see much value in the proposed sender. All I'm saying is, that in the mean time we have gotten the iterate algorithm to turn a std views into sequence senders.

I have a lot of usage for this in networking and code with business logic.

I would like to dicuss whether helpful in the use cases of @miscco

ericniebler:
it seems this PR has been abandoned. closing.

PR #914, opened 2023-04-26
too much `stdexec::` namespace qualification in `nvexec/`

ericniebler:


include/nvexec/stream/algorithm_base.cuh
```diff
@@ -102,47 +100,46 @@ namespace nvexec::STDEXEC_STREAM_DETAIL_NS::__algo_range_init_fun {
...
-      friend auto tag_invoke(stdexec::get_completion_signatures_t, Self&&, Env)
+      template <__decays_to<__t> Self, class Env>
+      friend auto tag_invoke(get_completion_signatures_t, Self&&, Env)
         -> completion_signatures<Self, Env>
         requires true;
 
-      friend auto tag_invoke(stdexec::get_env_t, const __t& self) //
-        noexcept(stdexec::__nothrow_callable<stdexec::get_env_t, const Sender&>)
-          -> stdexec::__call_result_t<stdexec::get_env_t, const Sender&> {
-        return stdexec::get_env(self.sndr_);
```
maikel, line 145: Isn't it here a good idea to call `stdexec::get_env(self.sndr_)` instead of `get_env(self.sndr_)` because of ADL ?

ericniebler: `get_env` is an object, not a function. No ADL if it's found in phase one name lookup.

PR #920, opened 2023-04-28
Implement any_stop_token

maikel:
Hi,

the following lines currently do not work because `sync_wait` provides a `never_stop_token`-receiver

```cpp
  using receiver_ref =
    any_receiver_ref<Sigs, get_stop_token.signature<in_place_stop_token() noexcept>>;
  using stoppable_sender = receiver_ref::any_sender<>;
  stoppable_sender sender = when_any(just(21));
  // won't compile
  sync_wait(std::move(sender));
```

Instead this PR defines any_stop_token. Use it like this

```cpp
  using receiver_ref = any_receiver_ref<
    completion_signatures<set_value_t(int), set_stopped_t()>,
    get_stop_token.signature<any_stop_token() noexcept>>;
  using stoppable_sender = receiver_ref::any_sender<>;
  stoppable_sender sender = when_any(just(21));
  // this compiles
  auto maybe_value = sync_wait(std::move(sender));
```


ericniebler:
In many places an arbitrary input stop token type can be adapted into an `in_place_stop_token` by registering a stop callback on the input token that requests stop on an `in_place_stop_source`. I'm not sure what the tradeoffs are here, tho.

@lewissbaker, I know you use the callback-to-inplace-stop-source technique. Is there a reason you have preferred this over direct type-erasure of the token itself?

maikel:
Do you suggest to make the type erased sender always stoppable by using `in_place_stop_*` within the implementation? I like that, because currently it is just too easy to end up with an unstoppable sender by accident

test/exec/test_any_sender.cpp
```diff
@@ -361,19 +366,19 @@ struct stopped_receiver : stopped_receiver_base {
...
 
-TEST_CASE("any_sender does not connect with stop token", "[types][any_sender]") {
-  using unstoppable_sender = any_sender_of<set_value_t(int), set_stopped_t()>;
-  unstoppable_sender sender = when_any(just(21));
+TEST_CASE("any_sender - does connect with stop token", "[types][any_sender]") {
+  using stoppable_sender = any_sender_of<set_value_t(int), set_stopped_t()>;
+  stoppable_sender sender = when_any(just(21));
   in_place_stop_source stop_source{};
-  stopped_receiver receiver{stop_source.get_token(), false};
+  stopped_receiver receiver{stop_source.get_token(), true};
```
ericniebler, line 373: This stoppable_sender should be able to type-erase a sender with any kind of stop token, right? Not just in_place and never.

ericniebler: Thank you. Other than that, this looks great.

include/exec/any_sender_of.hpp
```diff
@@ -544,50 +755,119 @@ namespace exec {
...
+          _SetStopped{}(static_cast<_Receiver&&>(__self.__op_->__rcvr_));
+        }
+
+        template <same_as<get_env_t> _GetEnv, same_as<__t> _Self>
+        friend __env_t<env_of_t<_Receiver>> tag_invoke(_GetEnv, const _Self& __self) noexcept {
+          return __make_env(
+            get_env(__self.__op_->__rcvr_),
+            __with_(get_stop_token, __self.__op_->__stop_source_.get_token()));
+        }
+      };
```
ericniebler, line 819: For future reference, my preference nowadays is to avoid nested classes. It makes diagnostics harder to grok, and it also unnecessarily adds the outer class as an associated entity that is searched for overloads. We can fix this later tho.

PR #925, opened 2023-05-08
switch to a lambda-based tuple for internal implementation details

ericniebler:
The lambda-based tuple is 2x faster to compile than `std::tuple` for gcc, and roughly on par for clang. https://build-bench.com/b/lIckxezdT6TiXczIfPSj0zNHK4I


ericniebler:
I'm not measuring an improvement in compile times. Closing.

PR #934, opened 2023-05-14
Demo of hide<>

kirkshoop:
This is a demo of how to apply the `hide<>` utility


ericniebler:
@kirkshoop Do you have a tl;dr for this pr? (A pr td;dr. Ha.) Is it a compile time win? What metrics does this move?

kirkshoop:
With these changes, This reduces symbol size for large expressions. More reduction would be gained by adding `hide_this` and `hide` to more senders and receivers. 
The big gains happen when large names are nested. users that have long namespace names that get repeated many times in the same symbol will see the most improvement.
This implementation has been structured to minimize symbol length, so the gains are less obvious in small programs.

kirkshoop:
longest symbol in main branch
```sh
➜  stdexec git:(main) ✗ objdump -Ct build/examples/example.scope | grep print_void > sym_scp_main.txt
➜  stdexec git:(main) ✗ awk '{ if (length($0) > max) {max = length($0)} } END { print max }' "sym_scp_main.txt"
63955
```

longest symbol in hiding branch

```sh
➜  stdexec git:(hiding) ✗ objdump -Ct build/examples/example.scope | grep print_void > sym_scp_hide.txt          
➜  stdexec git:(hiding) ✗ awk '{ if (length($0) > max) {max = length($0)} } END { print max }' "sym_scp_hide.txt"
5434
```

With these changes to examples/scope.cpp

```cpp
struct print_void_t {
  struct result {
    template<class... Vn>
    auto operator()(set_value_t, Vn&&...) const noexcept {
      printf("void\n");
      return just();
    }
    template<class Error>
    auto operator()(set_error_t, Error&& err) const noexcept {
      return just_error(err);
    }
    auto operator()(set_stopped_t) const noexcept {
      return just_stopped();
    }
  };
  template<class Tag>
  struct dispatch {
    template<class... Vn>
    auto operator()(Vn&&... vn) const noexcept {
      return result{}(Tag{}, vn...);
    }
  };
  template<class Input>
  auto operator()(Input&& in) const noexcept {
    return let_value(let_error(let_stopped(in, dispatch<set_stopped_t>{}), dispatch<set_error_t>{}), dispatch<set_value_t>{});
  }
};
inline constexpr print_void_t print_void{};
```

```cpp
  sender auto printVoid = print_void(when_all(                              //
    print_void(print_void(print_void(print_void(print_void(begin))))),      //
    print_void(print_void(print_void(print_void(print_void(begin)))))));    // 3
```

ericniebler:
> longest symbol in main branch
> 63955
> 
> longest symbol in hiding branch
> 5434

I don't doubt this makes unmangled names significantly shorter, and that effects diagnostics in good ways and bad ways. Good: less spew to sift through. Bad: potentially important details are omitted.

I don't know how much the demangled name length effects compile and link times, I/O, and memory usage, though. I imagine the toolchain works with mangled names, and (I would guess) the savings there are a lot smaller. The mangling scheme does the de-duping that you want.

I really need to think carefully before I obfuscate the types.

kirkshoop:
Yes, this is not a good situation. I do want to have a solution and a case for why the standard and tools need to change

PR #952, opened 2023-05-31
Fix an access of uninitialized data in the CUDA stream implementation of `schedule_from`

ericniebler:


include/nvexec/stream/schedule_from.cuh
```diff
@@ -40,7 +40,7 @@ namespace nvexec::STDEXEC_STREAM_DETAIL_NS {
 
         template < __completion_tag Tag, class... As>
         friend void tag_invoke(Tag, __t&& self, As&&... as) noexcept {
-          storage_t* storage = static_cast<storage_t*>(self.operation_state_.temp_storage_);
+          local<storage_t> storage(self.operation_state_.temp_storage_);
```
senior-zero, line 43: I believe this change opens possibilities for races. Signal propagation is asynchronous and might return before the value is consumed:

```cpp
struct test_t {
  void use() {
    std::printf("test_t::use(%p)\n", this);
  }

  ~test_t() {
    std::printf("test_t::~test_t(%p)\n", this);
  }
};

int main() {
  using nvexec::is_on_gpu;

  nvexec::stream_context stream_ctx{};
  ex::scheduler auto sch = stream_ctx.get_scheduler();

  auto snd = ex::transfer_just(sch, test_t{}) | ex::then([](test_t&& t) { t.use(); })
           | ex::transfer(exec::inline_scheduler{});

  stdexec::sync_wait(std::move(snd));
}
```
leads to:
```
test_t::~test_t(0x7fd036000460)
test_t::use(0x7fd036000460)
```

Maybe we should allow trivially destructible types until we find a workaround?

include/nvexec/stream/bulk.cuh
```diff
@@ -25,6 +25,7 @@ namespace nvexec::STDEXEC_STREAM_DETAIL_NS {
   namespace _bulk {
     template <int BlockThreads, class... As, std::integral Shape, class Fun>
     __launch_bounds__(BlockThreads) __global__ void kernel(Shape shape, Fun fn, As... as) {
+      static_assert(trivially_copyable<Shape, Fun, As...>);
```
senior-zero, line 28: Should there be a symmetrical assert in the multi gpu bulk kernel?

ericniebler: I'd like to put that assert in every kernel.

include/nvexec/stream/schedule_from.cuh
ericniebler: > Signal propagation is asynchronous and might return before the value is consumed

Thanks, I wondered about that. Restricting to trivially destructible types would be an ok first step if a real fix looks like it will be involved.

ericniebler:
superseded

PR #968, opened 2023-06-09
temporary fix for `nvexec::schedule_from`'s access of uninitialized memory

ericniebler:


include/nvexec/stream/schedule_from.cuh
```diff
@@ -41,6 +41,7 @@ namespace nvexec::STDEXEC_STREAM_DETAIL_NS {
         template < __completion_tag Tag, class... As>
         friend void tag_invoke(Tag, __t&& self, As&&... as) noexcept {
           storage_t* storage = static_cast<storage_t*>(self.operation_state_.temp_storage_);
+          ::new (storage) storage_t();
```
senior-zero, line 44: Please, add a TODO to remove host access later. Also, could you have an optimization for the base case `sizeof...(as) == 0` to just propagate tag without touching temporary storage?

senior-zero (approved):
If there's time, I'd suggest to add a test capturing the fixed behavior.

senior-zero (approved):

ericniebler:
> If there's time, I'd suggest to add a test capturing the fixed behavior.

this is a hard one to test. whether any issue manifests at all depends on the contents of uninitialized managed memory. not even `compute-sanitizer` detected anything amiss. I just got lucky and happened to step through this code when the memory wasn't zeroed.

PR #970, opened 2023-06-11
put `nvexec::sync_wait` op-state in pinned host memory

### LINKED ISSUE #953: track down problem in the CUDA stream implementation of `bulk`
See the `nvexec/reduce` example in the [`nvexec-bulk-bug`](https://github.com/NVIDIA/stdexec/pull/new/nvexec-bulk-bug) branch for a repro, or just run the tests.

Attached is a dump from `compute-sanitizer`.

[error.log](https://github.com/NVIDIA/stdexec/files/11608471/error.log)

ericniebler:
fixes #953

### DIFF
--- a/include/nvexec/detail/memory.cuh
+++ b/include/nvexec/detail/memory.cuh
@@ -1,0 +1,113 @@
+/*
+ * Copyright (c) 2022 NVIDIA Corporation
+ *
+ * Licensed under the Apache License Version 2.0 with LLVM Exceptions
+ * (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ *   https://llvm.org/LICENSE.txt
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#pragma once
+
+#include "../../stdexec/execution.hpp"
+
+#include <memory_resource>
+#include <new>
+#include <type_traits>
+
+#include "config.cuh"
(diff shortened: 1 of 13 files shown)


include/nvexec/detail/memory.cuh
```diff
@@ -0,0 +1,113 @@
...
+
+#include "config.cuh"
+#include "throw_on_cuda_error.cuh"
+
+namespace nvexec::STDEXEC_STREAM_DETAIL_NS {
+
+  struct device_deleter {
+    template <class T>
+    void operator()(T* ptr) {
+      ptr->~T();
```
senior-zero, line 32: This is a new behavior that's shouldn't be necessary for this PR.

ericniebler: I've commented it out. Is your concern that the destructor might get called from a place where the object is inaccessible?

senior-zero: It's the same concern as in the following [comment](https://github.com/NVIDIA/stdexec/pull/952#discussion_r1222674960). Potential managed memory traffic.

include/nvexec/detail/memory.cuh
```diff
@@ -0,0 +1,113 @@
...
+    return make_device<__decay_t<A>>(status, (A&&) t);
+  }
+
+  template <class T>
+  struct host_deleter {
+    std::pmr::memory_resource* resource_{nullptr};
+
+    void operator()(T* ptr) const {
+      if (ptr) {
+        ptr->~T();
```
ericniebler, line 75: @senior-zero would you also like to see me replace this destructor call with a `static_assert(is_trivially_destructible_v<T>)`?

senior-zero: Not really, I think it should be fine to have the destructor for pinned memory as long as it's invoked after all the kernels relying on it are completed.

senior-zero (approved):

PR #972, opened 2023-06-13
Optimize some managed memory movements

senior-zero:
Standard memory resources access allocated memory leading to managed memory traffic. This PR replaces `monotonic_buffer_resource` and `synchronized_pool_resource` with some simplified implementations. It also fixes `upon_error`, which used to dereference error. Main changes are related to `schedule_from`. It's copying input values into kernel arguments if those are trivially copyable and constructs the temporary storage on device in stream order. Otherwise, it constructs the temporary storage on the host and starts asynchronous prefetch on device if possible.

### DIFF
--- a/include/nvexec/detail/memory.cuh
+++ b/include/nvexec/detail/memory.cuh
@@ -17,9 +17,15 @@
 
 #include "../../stdexec/execution.hpp"
 
+#include <cuda/std/bit>
+
 #include <memory_resource>
 #include <new>
 #include <type_traits>
+#include <map>
+#include <mutex>
+#include <optional>
+#include <set>
 
 #include "config.cuh"
 #include "throw_on_cuda_error.cuh"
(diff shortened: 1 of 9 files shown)


include/nvexec/stream/reduce.cuh
```diff
@@ -46,8 +46,16 @@ namespace nvexec {
...
         static void set_value_impl(base::__t&& self, Range&& range) noexcept {
+          cudaError_t status{cudaSuccess};
           cudaStream_t stream = self.op_state_.get_stream();
 
+          // `range` is produced asynchronously, so we need to wait for it to be ready
+          if (status = STDEXEC_DBG_ERR(cudaStreamSynchronize(stream));
+              status != cudaSuccess) {
+            self.op_state_.propagate_completion_signal(stdexec::set_error, std::move(status));
+            return;
+          }
```
senior-zero, line 57: Reduce used to access potentially asynchronous value on the host. Making `schedule_from` asynchronous revealed the race. @miscco I think we need a reduce overload that accepts a range as input rather than in the value channel to provide efficient implementation when the range is known beforehand:

```cpp
bulk(N, update(field)) | reduce(field, field + n)
```

include/nvexec/detail/memory.cuh
```diff
@@ -110,4 +116,243 @@ namespace nvexec::STDEXEC_STREAM_DETAIL_NS {
...
+      block_descriptor_t(size_t bytes)
+        : bin(cuda::std::bit_width(bytes))
+        , bytes(1ull << bin) {
+        if (bin < min_bin) {
+          bin = min_bin;
+          bytes = 1ull << bin;
+        }
+      }
+
+      block_descriptor_t(void* ptr)
```
ericniebler, line 265:
```suggestion
      explicit block_descriptor_t(void* ptr)
```

include/nvexec/detail/memory.cuh
```diff
@@ -110,4 +116,243 @@ namespace nvexec::STDEXEC_STREAM_DETAIL_NS {
...
+      , live_blocks(ptr_comparator_t{}) {
+    }
+
+    void* do_allocate(size_t bytes, size_t /* alignment */) override {
+      std::lock_guard<std::mutex> lock(mutex);
+
+      block_descriptor_t search_key{bytes};
+      cached_blocks_t::iterator block_itr = cached_blocks.lower_bound(search_key);
+
+      while ((block_itr != cached_blocks.end()) && (block_itr->bin == search_key.bin)) {
```
ericniebler, line 303:
```suggestion
      if ((block_itr != cached_blocks.end()) && (block_itr->bin == search_key.bin)) {
```

?

include/nvexec/detail/memory.cuh
```diff
@@ -110,4 +116,243 @@ namespace nvexec::STDEXEC_STREAM_DETAIL_NS {
...
+    };
+
+    struct size_comparator_t {
+      bool operator()(const block_descriptor_t& a, const block_descriptor_t& b) const {
+        return a.bytes < b.bytes;
+      }
+    };
+
+    using cached_blocks_t = std::multiset<block_descriptor_t, size_comparator_t>;
+    using busy_blocks_t = std::multiset<block_descriptor_t, ptr_comparator_t>;
```
ericniebler, line 283: Why `multiset` here? There won't be two block descriptors with the same ptr, will there?

include/nvexec/detail/memory.cuh
```diff
@@ -110,4 +116,243 @@ namespace nvexec::STDEXEC_STREAM_DETAIL_NS {
...
+    bool do_is_equal(const std::pmr::memory_resource& other) const noexcept override {
+      return this == &other;
+    }
+
+    ~synchronized_pool_resource() {
+      while (!cached_blocks.empty()) {
+        cached_blocks_t::iterator begin = cached_blocks.begin();
+        upstream->deallocate(begin->ptr, begin->bytes, alignment);
+        cached_blocks.erase(begin);
+      }
```
ericniebler, line 337: What about the live blocks?

include/nvexec/detail/memory.cuh
```diff
@@ -110,4 +116,243 @@ namespace nvexec::STDEXEC_STREAM_DETAIL_NS {
...
+    }
+  };
```
ericniebler, line 339: I desperately want to optimize this -- slab allocation! free lists! -- but it doesn't make sense unless a profiler says it does, I suppose.

include/nvexec/stream/schedule_from.cuh
```diff
@@ -40,15 +47,61 @@ namespace nvexec::STDEXEC_STREAM_DETAIL_NS {
...
           // As an optimization, if there are no values to persist to temporary
           // storage, skip it and simply propagate the completion signal.
           if constexpr (sizeof...(As) == 0) {
             self.operation_state_.propagate_completion_signal(Tag());
           } else {
             storage_t* storage = static_cast<storage_t*>(self.operation_state_.temp_storage_);
-            // BUGBUG TODO: construct/emplace from the device, not the host!
-            ::new (storage) storage_t();
-            storage->template emplace<decayed_tuple<Tag, As...>>(Tag(), (As&&) as...);
+            constexpr bool construct_on_device = trivially_copyable<__decay_t<As>...>;
```
ericniebler, line 58: This branch could really use an introductory comment explaining what it's doing.

include/nvexec/stream/schedule_from.cuh
```diff
@@ -57,8 +110,7 @@ namespace nvexec::STDEXEC_STREAM_DETAIL_NS {
                     self.operation_state_.propagate_completion_signal(Tag2(), std::move(tas)...);
                   },
                   tpl);
-              },
-              *storage);
+              }, *storage, index);
```
ericniebler, line 113: Why is it necessary to pass the index here? Just curious.

senior-zero: The variant might be constructed on the device asynchronously:

```cpp
if constexpr (construct_on_device) {
  kernel<Tag, storage_t, __decay_t<As>...><<<1, 1, 0, stream>>>(storage, as...);

  if (cudaError_t status = STDEXEC_DBG_ERR(cudaPeekAtLastError());
      status != cudaSuccess) {
    self.operation_state_.propagate_completion_signal(stdexec::set_error, std::move(status));
    return;
  }
} 
```

Without passing index, `visit` reads the index from the storage object, leading to a data transfer, which is not needed. Besides that, following kernel will have to prefetch data back on device. Since we know the index in advance from the `decayed_tuple<Tag, As...>` type, we can avoid this.

include/nvexec/detail/memory.cuh
```diff
@@ -110,4 +116,243 @@ namespace nvexec::STDEXEC_STREAM_DETAIL_NS {
...
+  struct monotonic_buffer_resource : std::pmr::memory_resource {
+    static constexpr size_t block_alignment = 256;
+
+    struct block_descriptor_t {
+      void* ptr{};
+      size_t total{};
+    };
+
+    std::pmr::memory_resource* upstream;
+    std::vector<block_descriptor_t> allocated_blocks;
```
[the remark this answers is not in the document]
senior-zero: Would there still be allocations for each node in the list? The two operations involving this list are pushing a new block and traversing it in the destructor. Pushing a new block into a linked list is O(N) allocations, while it's amortized O(1) for the vector. Traversing the linked list is accessing N random locations, whereas, for the vector, it will be predictable and local. Could you please elaborate on how the locality is improved by linked list usage?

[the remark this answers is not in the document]
ericniebler: What I mean by "intrusive linked list" is for each block allocation to store a pointer to the next block allocation itself, rather than having a separate data structure storing pointers to all the blocks.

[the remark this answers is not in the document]
senior-zero: Just to make sure I understand. You haven't suggested making an intrusive list of the `block_descriptor_t` but rather add the descriptor to the memory allocation itself. I'm afraid this defines the purpose of replacing standard memory resources in the first place. Memory resources provided as part of this PR should guarantee that the memory allocated by upstream resource is not accessed.

include/nvexec/detail/memory.cuh
senior-zero: Good catch, I'll add an assert

include/nvexec/detail/memory.cuh
senior-zero: I wanted to have something simple for now to remove the initial managed transaction as fast as possible. I think we can improve this later.

Issue #960, opened 2023-06-06
Access of uninitialized data in the CUDA stream implementation of `schedule_from`


ericniebler:
The larger issue is that CUDA stream senders put objects into untyped managed temporary storage using placement `new` (or, in the case of `schedule_from`, _not_ calling placement `new` and leaving the memory uninitialized :grimacing:), but never destroys the objects by calling the objects' destructors.

ericniebler:
I have landed #968 as a stop-gap. A larger issue remains that objects placed in temporary managed storage are not always constructed/destructed properly. #952 is an attempt at a proper fix, but it has issues yet to be addressed.

Leaving this open for now.

PR #975, opened 2023-06-15
Fix temp storage lifetime

senior-zero:
This PR is based on https://github.com/NVIDIA/stdexec/pull/972 and addresses remaining issues from https://github.com/NVIDIA/stdexec/issues/960.

### DIFF
--- a/include/nvexec/stream/common.cuh
+++ b/include/nvexec/stream/common.cuh
@@ -15,6 +15,7 @@
  */
 #pragma once
 
+#include <stack>
 #include <atomic>
 #include <memory_resource>
 #include "../../stdexec/execution.hpp"
@@ -77,24 +78,6 @@ namespace nvexec {
       ((STDEXEC_IS_TRIVIALLY_COPYABLE(Ts) || std::is_reference_v<Ts>) &&...);
 #endif
 
-    struct context_state_t {
-      std::pmr::memory_resource* pinned_resource_{nullptr};
-      std::pmr::memory_resource* managed_resource_{nullptr};
-      queue::task_hub_t* hub_{nullptr};
-      stream_priority priority_;
-
-      context_state_t(
-        std::pmr::memory_resource* pinned_resource,
-        std::pmr::memory_resource* managed_resource,
-        queue::task_hub_t* hub,
-        stream_priority priority = stream_priority::normal)
-        : pinned_resource_(pinned_resource)
-        , managed_resource_(managed_resource)
-        , hub_(hub)
-        , priority_(priority) {
-      }
-    };
-
     inline std::pair<int, cudaError_t> get_stream_priority(stream_priority priority) {
       int least{};
       int greatest{};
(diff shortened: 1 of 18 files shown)


include/nvexec/stream/common.cuh
```diff
@@ -386,34 +442,65 @@ namespace nvexec {
...
+              cudaMemRangeGetAttribute(
+                &device_id, 4, cudaMemRangeAttributeLastPrefetchLocation, ptr, sizeof(T));
+
+              if (cudaCpuDeviceId == device_id) {
+                ptr->~T();
+              } else {
+                destructor_kernel<<<1, 1, 0, stream>>>(ptr);
+
+                // TODO Bury all the memory associated with the stream provider and then 
+                //      deallocate the memory
```
ericniebler, line 493: Does this TODO need to be addressed in this PR? If not, let's open an issue for it.

include/nvexec/stream/let_xxx.cuh
```diff
@@ -110,17 +110,17 @@ namespace nvexec::STDEXEC_STREAM_DETAIL_NS {
...
           cudaStream_t stream = __self.__op_state_->get_stream();
-
           result_sender_t* result_sender = static_cast<result_sender_t*>(
             __self.__op_state_->temp_storage_);
           kernel_with_result<_As&&...><<<1, 1, 0, stream>>>(
             std::move(__self.__op_state_->__fun_), result_sender, (_As&&) __as...);
 
           if (cudaError_t status = STDEXEC_DBG_ERR(cudaStreamSynchronize(stream));
               status == cudaSuccess) {
+            __self.__op_state_->defer_temp_storage_destruction(result_sender);
```
ericniebler, line 120: If `status != cudaSuccess`, then the result sender may or may not be constructed, right? How can we know for certain whether the destructor needs to be called?

include/nvexec/stream/split.cuh
```diff
@@ -181,12 +172,11 @@ namespace nvexec::STDEXEC_STREAM_DETAIL_NS {
           if constexpr (stream_sender<Sender>) {
             STDEXEC_DBG_ERR(cudaEventDestroy(event_));
           }
-          STDEXEC_DBG_ERR(cudaStreamDestroy(stream_));
         }
       }
 
       env_t make_env() const noexcept {
-        return _split::__make_env(stop_source_, stream_);
+        return _split::__make_env(stop_source_, &const_cast<stream_provider_t&>(stream_provider_));
```
ericniebler, line 179: I wonder if there's a way to keep this `const`-correct. What mutating operation does the env need of the stream provider?

include/nvexec/stream/common.cuh
```diff
@@ -386,34 +442,65 @@ namespace nvexec {
...
+                cudaStreamSynchronize(stream); 
+              }
+
+              managed_resource->deallocate(ptr, sizeof(T));
+            });
+          }
```
ericniebler, line 499: The purist in me wants to see the registration of the destroy/deallocate action to happen as part of the allocate/construct action, rather than as a separate step on a separate data structure. Is that possible?

test/nvexec/then.cpp
```diff
@@ -201,12 +202,95 @@ TEST_CASE("nvexec then can return values of non-trivial types", "[cuda][stream][
...
+      cuda::std::atomic_ref<int> ref{*(is_on_gpu() ? d_counter_ : h_counter_)};
+      ref.fetch_add(val, cuda::std::memory_order_relaxed);
+    }
+
+  public:
+    void more() {
+      diff(+1);
+    }
+
+    void less() {
```
ericniebler:
```suggestion
    void fewer() {
```

Sorry, my inner grammar nazi got the best of me. ;-)

include/nvexec/stream/split.cuh
senior-zero: Env itself doesn't mutate the stream provider. It only distributes the reference to it. On the usage side, we need to register temporary storage destruction, which mutates the vector inside the stream provider.

include/nvexec/stream/common.cuh
senior-zero: Registration of deallocation has to happen at a particular place (before propagation of the completion signal and after some kernel) in a particular condition (no CUDA errors). I don't think it can happen as part of allocation/construction action without some additional memory traffic.

include/nvexec/stream/let_xxx.cuh
senior-zero: The destructor is registered only when `status == cudaSuccess`, which should be an improvement compared to the current state. If the `status != cudaSuccess`, it's still possible that the constructor was called and we miss the destruction, but I suggest we address this as a follow up PR (once the mechanism is in place).

include/nvexec/stream/common.cuh
senior-zero: https://github.com/NVIDIA/stdexec/issues/979

ericniebler (approved):

PR #977, opened 2023-06-16
Io uring fix stop facade

maikel:
Using the stopping facade writes currently the wrong pointer into the `IORING_OP_ASYNC_CANCEL` operation. This PR fixes this by passing the pointer to `__task` which is used for the original submission.

Its still a draft because I stlil want to think about a cleaner solution

### DIFF
--- a/test/exec/test_io_uring_context.cpp
+++ b/test/exec/test_io_uring_context.cpp
@@ -309,12 +309,12 @@ TEST_CASE("io_uring_context schedule_after -1s", "[types][io_uring][schedulers]"
         CHECK(io_thread.get_id() == std::this_thread::get_id());
         is_called_1 = true;
       }),
-      schedule_after(scheduler, 5ms) | then([&] {
+      schedule_after(scheduler, 100ms) | then([&] {
         is_called_2 = true;
       })));
     auto end = std::chrono::steady_clock::now();
     auto diff = end - start;
-    CHECK(diff < 5ms);
+    CHECK(diff < 100ms);
     CHECK(is_called_1 == true);
     CHECK(is_called_2 == false);
   }
--- a/test/exec/test_io_uring_context.cpp
+++ b/test/exec/test_io_uring_context.cpp
@@ -301,14 +301,22 @@ TEST_CASE("io_uring_context schedule_after -1s", "[types][io_uring][schedulers]"
     scope_guard guard{[&]() noexcept {
       context.request_stop();
     }};
-    bool is_called = false;
+    bool is_called_1 = false;
+    bool is_called_2 = false;
+    auto start = std::chrono::steady_clock::now();
     sync_wait(when_any(
       schedule_after(scheduler, -1s) | then([&] {
         CHECK(io_thread.get_id() == std::this_thread::get_id());
-        is_called = true;
+        is_called_1 = true;
       }),
-      schedule_after(scheduler, 5ms)));
-    CHECK(is_called);
+      schedule_after(scheduler, 5ms) | then([&] {
+        is_called_2 = true;
+      })));
+    auto end = std::chrono::steady_clock::now();
+    auto diff = end - start;
+    CHECK(diff < 5ms);
+    CHECK(is_called_1 == true);
+    CHECK(is_called_2 == false);
   }
 }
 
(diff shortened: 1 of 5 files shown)


test/exec/test_io_uring_context.cpp
```diff
@@ -301,14 +301,22 @@ TEST_CASE("io_uring_context schedule_after -1s", "[types][io_uring][schedulers]"
...
+        is_called_1 = true;
       }),
-      schedule_after(scheduler, 5ms)));
-    CHECK(is_called);
+      schedule_after(scheduler, 5ms) | then([&] {
+        is_called_2 = true;
+      })));
+    auto end = std::chrono::steady_clock::now();
+    auto diff = end - start;
+    CHECK(diff < 5ms);
```
ericniebler, line 317: The failure of this test perhaps indicates that work scheduled to run in the future on the io_uring scheduler doesn't complete promptly when it has been cancelled.

maikel: Hm, it might be as you say. I will look into it asap. The failing test uses kernel 5.4 and the implementation not touched by this pr (yet) because I only changed the  `IORING_OP_ASYNC_CANCEL` code path, which requires  kernel 5.5 and newer.

ericniebler:
It doesn't seem to be a timing issue.

maikel:
I could finally reproduce the issue. It is because the container image is based on Ubuntu-2204 but the GPU host runner is probably on an older system with a Kernel 5.4 (I assume ubuntu-2004). 

I could reproduce the error in a Ubuntu-2004 VM where I compiled and ran the tests in an Ubuntu-2204 container.

We have implemented two strategies for cancelation based on the kernel version. But since this is a header-only solution the detection of the kernel version works via defines in `/usr/include/linux/version.h` which reports a newer kernel version than what is actually being used.

The question is how to solve this now. I know of following solutions

1. Generate a cmake configure file with the actual detected kernel version of the host system and use this detection
2. Make the cancelation strategy a run-time decision instead of a build-time decision by detecting the factual kernel-version in run-time
3. Improve the CI by matching Host kernel and kernel headers within the image.

any thoughts?

@trxcllnt @ericniebler

ericniebler:
> 3. Improve the CI by matching Host kernel and kernel headers within the image.

This would be my preferred solution. I don't think we're under any obligation to support Franken-distros. @trxcllnt, any clue what could have led to this situation?

maikel:
> > 3. Improve the CI by matching Host kernel and kernel headers within the image.
> 
> This would be my preferred solution. I don't think we're under any obligation to support Franken-distros. @trxcllnt, any clue what could have led to this situation?

I want to side note that up until recently the default github action runner were based on ubuntu-20.04 too.

If upgrading the runners host system is too complicated because of other dependencies the test should also be OK with a container image based on 20.04. Or at least faking the Linux header files

ericniebler:
Paul makes the valid point that if code is compiled on one kernel version and run on another, it should still work. That argues for runtime detection.

maikel:
> Paul makes the valid point that if code is compiled on one kernel version and run on another, it should still work. That argues for runtime detection.

We have compatibilty for newer kernel versions. i.e. if we compile an application with kernel version 5.4 it will also run correctly on kernel 5.15. If I build my application against kernel version 5.15 and use features of this newer kernel I don't really expect my application to work correctly on an older kernel.

I will add the runtime check nonetheless. But it will pessimize the size of the timed schedule operations, because the solution that uses timerfd has more state.

io_uring itself provides a query for supported op codes but this is officially available since kernel version 5.6. Since we Ubuntu-20.04 ships with kernel version 5.4 the only way to test for cancelation support is to make dummy cancel request and check the errro code for `-EINVAL`. I will do so in the constructor of `io_uring_context`.

maikel:
How about this: We require a kernel version of 5.5 (which introduces `IORING_OP_ASYNC_CANCEL`) or higher for `io_uring` support and if we detect an older version within CMake we shouldn't build the io_uring tests? I could completely remove the `timerfd` workaround.

maikel:
There was something weird going on for `nvc++` for the run-time detection. I've revert those changes and disabled the tests for GPU CI. We just need to remember re-enable the tests whenever the host machine of the GPU gets a newer kernel version.

maikel:
Thank you!

PR #978, opened 2023-06-17
Fix result type computation in `exec::when_any`

maikel:
There is a missing `__munique` within the variant computation. I also added a test that catches this for me.


ericniebler (approved):

ericniebler:
Thanks!

PR #988, opened 2023-06-26
Add new upon_error test with 0 or many input errors

ccotter:
Add a new test to verify the behavior when there are 0 or many possible input error types.


ericniebler:
Thanks!

PR #989, opened 2023-06-26
Implement ignore_all_values

maikel:
This algorithm is sequence sender consumer. It takes a sequence sender and returns a sender. The returned sender completes with `set_value_t()` whenever the input sequence completes unless one of the input items complete via `set_error_t`  or `set_stopped_t`. 
If one of the sequence items complete with an error or stop signal, then the returned sender will forward this completion. In this case, the input sequence-sender is requested to stop.


ericniebler (approved):
LGTM

PR #990, opened 2023-06-26
Implement the iterate sequence sender

maikel:
This algorithm is a sequence sender factory. It takes an input range as function parameter and returns a lock step sequence sender that iterates through the range. The sequence operation completes if the end of the range is reached or if a next sender completes with via the stop signal.

### DIFF
--- a/include/exec/sequence/iterate.hpp
+++ b/include/exec/sequence/iterate.hpp
@@ -44,6 +44,7 @@ namespace exec {
     template <class _Iterator, class _Sentinel, class _ItemRcvr>
     struct __item_operation {
       struct __t {
+        using __id = __item_operation;
         STDEXEC_NO_UNIQUE_ADDRESS _ItemRcvr __rcvr_;
         __operation_base<_Iterator, _Sentinel>* __parent_;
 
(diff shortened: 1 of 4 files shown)


include/exec/sequence/iterate.hpp
```diff
@@ -0,0 +1,191 @@
...
+      STDEXEC_NO_UNIQUE_ADDRESS _Sentinel __sentinel_;
+    };
+
+    template <class _Range>
+    using __operation_base_t =
+      __operation_base<std::ranges::iterator_t<_Range>, std::ranges::sentinel_t<_Range>>;
+
+    template <class _Iterator, class _Sentinel, class _ItemRcvr>
+    struct __item_operation {
+      struct __t {
```
ericniebler, line 46:
```suggestion
      struct __t {
        using __id = __item_operation;
```

include/exec/sequence/iterate.hpp
```diff
@@ -0,0 +1,191 @@
...
+          return {
+            {std::ranges::begin(__self.__range_), std::ranges::end(__self.__range_)},
+            static_cast<_Receiver&&>(__rcvr)
+          };
+        }
+      };
+    };
+
+    struct iterate_t {
+      template <std::ranges::range _Range>
```
ericniebler, line 179: From the implementation, I think we need a forward-or-better range.

```suggestion
      template <std::ranges::forward_range _Range>
        requires stdexec::__decay_copyable<_Range>
```

test/exec/sequence/test_iterate.cpp
```diff
@@ -0,0 +1,105 @@
...
+ * limitations under the License.
+ */
+
+#include "exec/sequence/iterate.hpp"
+
+#ifdef __cpp_lib_ranges
+
+#include <catch2/catch.hpp>
+
+
```
ericniebler, line 24:
```suggestion
```

ericniebler (approved):
looks great

PR #991, opened 2023-06-27
Implement transform_each

maikel:
This algorithm is a sequence sender adaptor. It takes as input parameters a sequence sender and a sender adaptor. 
The algorithm applies the sender adaptor to each emited item of the input sequence sender and returns the resulting sequence sender.

### DIFF
--- a/include/exec/sequence/transform_each.hpp
+++ b/include/exec/sequence/transform_each.hpp
@@ -1,0 +1,204 @@
+/*
+ * Copyright (c) 2023 Maikel Nadolski
+ * Copyright (c) 2023 NVIDIA Corporation
+ *
+ * Licensed under the Apache License Version 2.0 with LLVM Exceptions
+ * (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ *   https://llvm.org/LICENSE.txt
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#pragma once
+
+#include "../sequence_senders.hpp"
+
+#include "../__detail/__basic_sequence.hpp"
+
+namespace exec {
+  namespace __transform {
+    using namespace stdexec;
+
+    template <class _Receiver, class _Adaptor>
+    struct __operation_base {
+      _Receiver __receiver_;
+      _Adaptor __adaptor_;
+    };
+
+    template <class _ReceiverId, class _Adaptor>
+    struct __receiver {
+      using _Receiver = stdexec::__t<_ReceiverId>;
+
+      struct __t {
+        using is_receiver = void;
+        using __id = __receiver;
+        __operation_base<_Receiver, _Adaptor>* __op_;
+
+        template <same_as<set_next_t> _SetNext, same_as<__t> _Self, class _Item>
+          requires __callable<_Adaptor&, _Item>
+                && __callable<exec::set_next_t, _Receiver&, __call_result_t<_Adaptor&, _Item>>
+        friend auto tag_invoke(_SetNext, _Self& __self, _Item&& __item) noexcept(
+          __nothrow_callable<_SetNext, _Receiver&, __call_result_t<_Adaptor&, _Item>> //
+          && __nothrow_callable<_Adaptor&, _Item>)
+          -> next_sender_of_t<_Receiver, __call_result_t<_Adaptor&, _Item>> {
+          return exec::set_next(
+            __self.__op_->__receiver_, __self.__op_->__adaptor_(static_cast<_Item&&>(__item)));
+        }
+
+        template <same_as<set_value_t> _SetValue, same_as<__t> _Self>
+        friend void tag_invoke(_SetValue, _Self&& __self) noexcept {
+          stdexec::set_value(static_cast<_Receiver&&>(__self.__op_->__receiver_));
+        }
+
+        template <same_as<set_stopped_t> _SetStopped, same_as<__t> _Self>
+          requires __callable<_SetStopped, _Receiver&&>
+        friend void tag_invoke(_SetStopped, _Self&& __self) noexcept {
+          stdexec::set_stopped(static_cast<_Receiver&&>(__self.__op_->__receiver_));
+        }
+
+        template <same_as<set_error_t> _SetError, same_as<__t> _Self, class _Error>
+          requires __callable<_SetError, _Receiver&&, _Error>
+        friend void tag_invoke(_SetError, _Self&& __self, _Error&& __error) noexcept {
+          stdexec::set_error(
+            static_cast<_Receiver&&>(__self.__op_->__receiver_), static_cast<_Error&&>(__error));
+        }
+
+        template <same_as<get_env_t> _GetEnv, __decays_to<__t> _Self>
+        friend env_of_t<_Receiver> tag_invoke(_GetEnv, _Self&& __self) noexcept {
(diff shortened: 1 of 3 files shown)


include/exec/sequence/transform_each.hpp
```diff
@@ -0,0 +1,161 @@
...
+      _Receiver __receiver_;
+      _Adaptor __adaptor_;
+    };
+
+    template <class _ReceiverId, class _Adaptor>
+    struct __receiver {
+      using _Receiver = stdexec::__t<_ReceiverId>;
+
+      struct __t {
+        using is_receiver = void;
```
ericniebler, line 36:
```suggestion
        using is_receiver = void;
        using __id = __receiver;
```

include/exec/sequence/transform_each.hpp
```diff
@@ -0,0 +1,161 @@
...
+          return stdexec::get_env(__self.__op_->__receiver_);
+        }
+      };
+    };
+
+    template <class _Sender, class _ReceiverId, class _Adaptor>
+    struct __operation {
+      using _Receiver = stdexec::__t<_ReceiverId>;
+
+      struct __t : __operation_base<_Receiver, _Adaptor> {
```
ericniebler, line 79:
```suggestion
      struct __t : __operation_base<_Receiver, _Adaptor> {
        using __id = __operation;
```

include/exec/sequence/transform_each.hpp
```diff
@@ -0,0 +1,161 @@
...
+
+        template <class _Self, class _Env>
+        using __some_sender = __copy_cvref_t<
+          _Self,
+          __sequence_sndr::__unspecified_sender_of<
+            completion_signatures_of_t<__copy_cvref_t<_Self, _Sender>, _Env>>>;
+
+        template <class _Self, class _Env>
+        using completion_sigs_t = completion_signatures_of_t<
+          __call_result_t<_Adaptor&, __some_sender<_Self, _Sender>>,
```
ericniebler, line 116: Is this supposed to be:

```suggestion
          __call_result_t<_Adaptor&, __some_sender<_Self, _Env>>,
```

? Also, I don't think I really understand what's going on here. We are saying the completion signatures of `transform_each` are the completions of a sender type that is returned from the adaptor when passed ... some unspecified sender type? Doesn't it matter which sender is passed to the adaptor?

include/exec/sequence/transform_each.hpp
```diff
@@ -0,0 +1,161 @@
...
+          -> completion_sigs_t<_Self, _Env>;
+
+        friend env_of_t<_Sender> tag_invoke(get_env_t, const __t& __self) noexcept {
+          return stdexec::get_env(__self.__sender_);
+        }
+      };
+    };
+
+    struct transform_each_t {
+      template <sender _Sender, class _Adaptor>
```
ericniebler, line 143: I think "`_Adaptor`" here is a sender adaptor closure, right? You can use `stdexec::__sender_adaptor_closure` to constrain the adaptor argument (or `stdexec::__sender_adaptor_closure_for` if you know the type of the sender that will be passed to the adaptor).

test/exec/sequence/test_transform_each.cpp
```diff
@@ -0,0 +1,48 @@
...
+TEST_CASE(
+  "transform_each - transform sender applies adaptor",
+  "[sequence_senders][transform_each][empty_sequence]") {
+  int value = 0;
+  auto transformed = exec::transform_each(
+    stdexec::just(42), stdexec::then([&value](int x) noexcept { value = x; }));
+  auto op = exec::subscribe(transformed, next_rcvr{});
+  stdexec::start(op);
+  CHECK(value == 42);
+}
```
ericniebler, line 48: I'd love to see a test with dependently-typed senders.

ericniebler (changes requested):

include/exec/sequence/transform_each.hpp
maikel: We pass the items of the predecessor sequence to the adaptor. In the current design we chose to not list their concrete item type somewhere. I think it's inaccessible in this context here. One could imagine a design where a sequence does not specify completion signatures but it's item sender types instead.

You will find a similar use of an unspecified sender in the sequence sender concepts for this reason.

The choice to specify only completion signatures of items  has been made for convenience. But from this follows some consistency assumptions, i. e. If I pass an item with signatures S I always get an item with signatures R, indepently of the concrete item type.

A similar consistency assumption is present for set next.

include/exec/sequence/transform_each.hpp
maikel: Yes I have missed this concrete vocabulary. Thank you

include/exec/sequence/transform_each.hpp
ericniebler: > from this follows some consistency assumptions, i. e. If I pass an item with signatures S I always get an item with signatures R, indepently of the concrete item type.

This makes me a little uncomfortable. Algorithms are customizable, and I can imagine some implementations might change signatures slightly: an additional error, or a stop signal.

test/exec/sequence/test_transform_each.cpp
```diff
@@ -0,0 +1,81 @@
...
+  "[sequence_senders][transform_each]") {
+  int value = 0;
+  auto transformed = exec::transform_each(
+    stdexec::just(42), stdexec::then([&value](int x) noexcept { value = x; }));
+  auto op = exec::subscribe(transformed, next_rcvr{});
+  stdexec::start(op);
+  CHECK(value == 42);
+}
+
+#ifdef __cpp_lib_ranges
```
ericniebler, line 65: Could you rebase on `main` and use the new `STDEXEC_HAS_RANGES()` macro here? Thx.

include/exec/sequence/transform_each.hpp
maikel: I have addressed that now with PR #1020 which is included here.

test/exec/sequence/test_transform_each.cpp
maikel: I have included a test where I dependently-typed the sender which is sent by the sequence.

include/exec/sequence/transform_each.hpp
```diff
@@ -0,0 +1,204 @@
...
+
+    template <class _Adaptor>
+    struct _NOT_CALLABLE_ADAPTOR_ { };
+
+    template <class _Item>
+    struct _WITH_ITEM_SENDER_ { };
+
+    template <class _Adaptor, class _Item>
+    auto __try_call(_Item*)
+      -> stdexec::__mexception<_NOT_CALLABLE_ADAPTOR_<_Adaptor&>, _WITH_ITEM_SENDER_<_Item>>;
```
ericniebler, line 125: i think we want to demangle the sender name here.

```suggestion
      -> stdexec::__mexception<_NOT_CALLABLE_ADAPTOR_<_Adaptor&>, _WITH_ITEM_SENDER_<stdexec::__name_of<_Item>>>;
```

ericniebler (approved):

include/exec/sequence/transform_each.hpp
maikel: Good catch

test/exec/sequence/test_transform_each.cpp
```diff
@@ -0,0 +1,121 @@
...
+
+#include "exec/sequence/transform_each.hpp"
+
+#include "exec/sequence/any_sequence_of.hpp"
+#include "exec/sequence/empty_sequence.hpp"
+#include "exec/sequence/iterate.hpp"
+#include "exec/sequence/ignore_all_values.hpp"
+#include <catch2/catch.hpp>
+
+#include <exec/on.hpp>
```
ericniebler, line 26:
```suggestion
#include <exec/env.hpp>
#include <exec/on.hpp>
```

PR #992, opened 2023-07-05
added exec::reduce including a customization for static_thread_pool

muellan:
This adds 
- `exec::reduce`
- a parallelized specialization of `exec::reduce` for running on `exec::static_thread_pool`'s scheduler
- tests covering basic usage with different reduction operations, input value-range edge cases, ranges of non-fundamental types, different schedulers, etc.

This does not (yet) introduce a common customization point with `nvexec::reduce` mainly because this would require a common default reduction operation (`nvexec::reduce` uses `cub::Sum` by default while `exec::reduce` uses `std::plus<>`).
If this is desired then one should probably also provide an entire set of common reduction operations like min, max, sum, product, etc.

### DIFF
--- a/include/exec/reduce.hpp
+++ b/include/exec/reduce.hpp
@@ -1,0 +1,203 @@
+/*
+ * Copyright (c) 2023 NVIDIA Corporation
+ *
+ * Licensed under the Apache License Version 2.0 with LLVM Exceptions
+ * (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ *   https://llvm.org/LICENSE.txt
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#pragma once
+
+#include "../stdexec/execution.hpp"
+
+#include <numeric>
+#include <ranges>
+
+namespace exec {
+
+  namespace __reduce {
+
+    template <class _ReceiverId, class _InitT, class _RedOp>
+    struct __receiver {
+      using _Receiver = stdexec::__t<_ReceiverId>;
+
+      struct __data {
+        _Receiver __rcvr_;
+        STDEXEC_NO_UNIQUE_ADDRESS _InitT __init_;
+        STDEXEC_NO_UNIQUE_ADDRESS _RedOp __redop_;
+      };
+
+      struct __t {
+        using __id = __receiver;
+        __data* __op_;
+
+        template <
+          stdexec::__same_as<stdexec::set_value_t> _Tag,
+          class _Range,
+          class _Value = stdexec::range_value_t<_Range>>
+          requires stdexec::invocable<_RedOp, _InitT, _Value>
+                && stdexec::__receiver_of_invoke_result<_Receiver, _RedOp, _InitT, _Value>
+        friend void tag_invoke(_Tag, __t&& __self, _Range&& __range) noexcept {
+          auto result = std::reduce(
+            std::ranges::begin(__range), std::ranges::end(__range), __self.__op_->__init_, __self.__op_->__redop_);
+
+          stdexec::set_value((_Receiver&&) __self.__op_->__rcvr_, std::move(result));
+        }
+
+        template <stdexec::__one_of<stdexec::set_error_t, stdexec::set_stopped_t> _Tag, class... _As>
+          requires stdexec::__callable<_Tag, _Receiver, _As...>
+        friend void tag_invoke(_Tag __tag, __t&& __self, _As&&... __as) noexcept {
+          __tag((_Receiver&&) __self.__op_->__rcvr_, (_As&&) __as...);
+        }
+
+        friend auto tag_invoke(stdexec::get_env_t, const __t& __self) noexcept
+          -> stdexec::env_of_t<const _Receiver&> {
+          return stdexec::get_env(__self.__op_->__rcvr_);
+        }
+      };
+    };
+
+    template <class _Sender, class _ReceiverId, class _InitT, class _RedOp>
+    struct __operation {
+      using _Receiver = stdexec::__t<_ReceiverId>;
+      using __receiver_id = __receiver<_ReceiverId, _InitT, _RedOp>;
+      using __receiver_t = stdexec::__t<__receiver_id>;
+
+      struct __t : stdexec::__immovable {
+        using __id = __operation;
+        typename __receiver_id::__data __data_;
+        stdexec::connect_result_t<_Sender, __receiver_t> __op_;
+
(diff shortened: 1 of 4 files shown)


include/exec/static_thread_pool.hpp
```diff
@@ -424,6 +422,263 @@ namespace exec {
...
+
+              sh_state.apply(reducer);
+
+              const bool is_last_thread = sh_state.finished_threads_.fetch_add(1)
+                                       == (total_threads - 1);
+              if (is_last_thread) {
+                sh_state.apply(finalize);
+              }
+            };
+          }
```
ericniebler, line 468: We don't appear to be handling exceptions that may be thrown from the reduction function.

include/exec/reduce.hpp
```diff
@@ -0,0 +1,204 @@
...
+
+      template <stdexec::sender _Sender, class _InitT, stdexec::__movable_value _RedOp>
+        requires(!stdexec::__tag_invocable_with_completion_scheduler<
+                  reduce_t,
+                  stdexec::set_value_t,
+                  _Sender,
+                  _InitT,
+                  _RedOp>)
+             && (!stdexec::tag_invocable<reduce_t, _Sender, _InitT, _RedOp>)
+      STDEXEC_DETAIL_CUDACC_HOST_DEVICE __sender<_Sender, _InitT, _RedOp>
```
ericniebler, line 185: This should require that _RedOp is callable with (_InitT, _InitT), right?

include/exec/reduce.hpp
```diff
@@ -0,0 +1,204 @@
...
+            ((_Self&&) __self).__redop_};
+        }
+
+        template <stdexec::__decays_to<__t> _Self, class _Env>
+        friend auto tag_invoke(stdexec::get_completion_signatures_t, _Self&&, _Env&&)
+          -> stdexec::dependent_completion_signatures<_Env>;
+
+        template <stdexec::__decays_to<__t> _Self, class _Env>
+        friend auto tag_invoke(stdexec::get_completion_signatures_t, _Self&&, _Env&&)
+          -> stdexec::completion_signatures<stdexec::set_value_t(_InitT)>
```
ericniebler, line 130: At this point, we have all the type information we need in order to do full type-checking. We can and should check that the sender can only complete one way, with a single value that is a range whose element type is the same as _InitT.

We also know whether the reduction function is potentially throwing, and if it is, to add `set_error_t(exception_ptr)` as a possible completion signature.

If you're feeling ambitious, you can see how an adaptor like `then` detects when the types don't match and tries to return a type that describes the error -- or propagates type errors from inner senders. If that's too much, don't sweat it. We can improve diagnostics in a separate PR.

include/exec/static_thread_pool.hpp
```diff
@@ -424,6 +422,263 @@ namespace exec {
...
+        struct dummy_ {
+          using __t = T;
+        };
+
+        using inrange_t = //
+          stdexec::__decay_t<stdexec::__t<stdexec::__t< stdexec::__value_types_of_t<
+            Sender,
+            stdexec::env_of_t<Receiver>,
+            stdexec::__q<dummy_>,
+            stdexec::__q<dummy_>>>>>;
```
ericniebler, line 481: I think this is equivalent:

```suggestion
        using inrange_t = //
          stdexec::__decay_t<
            stdexec::__value_types_of_t<
              Sender,
              stdexec::env_of_t<Receiver>,
              stdexec::__q<stdexec::__midentity>,
              stdexec::__q<stdexec::__midentity>>>;
```

include/exec/reduce.hpp
```diff
@@ -0,0 +1,204 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#pragma once
+
+#include "../stdexec/execution.hpp"
+
+#include <numeric>
+#include <ranges>
```
ericniebler, line 21: I think it's ok for it to be an error to include exec/reduce.hpp if the stdlib doesn't support `<ranges>`, but the reduce test should simply not run in that case instead of causing a test failure.

include/exec/reduce.hpp
```diff
@@ -0,0 +1,204 @@
...
+
+      struct __t {
+        using is_receiver = void;
+        using __id = __receiver;
+        __data* __op_;
+
+        template <
+          stdexec::__same_as<stdexec::set_value_t> _Tag,
+          class _Range,
+          class _Value = stdexec::range_value_t<_Range>>
```
ericniebler, line 45: `std::ranges::range_value_t` ?

ericniebler (commented):
Overall this looks really great, thanks Andre. A few nits, and needs some more type-checking.

include/exec/static_thread_pool.hpp
```diff
@@ -424,6 +422,263 @@ namespace exec {
...
+            : sh_state_(sh_state) {
+            task_base::__execute = [](task_base* t, const std::uint32_t tid) noexcept {
+              auto& sh_state = *static_cast<reduction_task*>(t)->sh_state_;
+              auto total_threads = sh_state.num_threads_available();
+              auto& partial = sh_state.partials_[tid];
+
+              auto reducer = [&](auto& input) {
+                const auto [first, last] = even_share(std::ranges::size(input), tid, total_threads);
+                if (first != last) {
+                  auto begin = std::ranges::begin(input) + first;
```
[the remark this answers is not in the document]
muellan: I know that I still need to add more type checking.
If it were up to me I would constrain it to random access ranges. On the other hand `std::reduce` requires only input iterators or forward iterators for the execution policy overloads. I don't know if anyone would actually expect `exec::reduce` to work with an input range.

include/exec/static_thread_pool.hpp
muellan: I know. I left exception handling out, just wanted stuff to work first. I'll add it.

include/exec/static_thread_pool.hpp
```diff
@@ -424,6 +422,263 @@ namespace exec {
...
+              };
+
+              auto finalize = [&](auto&) {
+                auto result = std::reduce(
+                  std::ranges::begin(sh_state.partials_),
+                  std::ranges::end(sh_state.partials_),
+                  sh_state.init_,
+                  sh_state.redop_);
+                // deallocate
+                sh_state.partials_ = {};
```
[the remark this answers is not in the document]
muellan: Can't believe I fell for this age-old trap.

include/exec/reduce.hpp
muellan: Right.

include/exec/reduce.hpp
muellan: I've seen the error handling in other sender adaptors. I think I should add it as part of this PR.

include/exec/reduce.hpp
muellan: agreed

include/exec/reduce.hpp
muellan: oops, that's a leftover I forgot to remove

PR #993, opened 2023-07-07
Fix linking the tests with GCC 12 and 13

villevoutilainen:
I'm not entirely sure why this function ends up being discarded by the linker, but this patch cures the problem.

### DIFF
--- a/include/stdexec/execution.hpp
+++ b/include/stdexec/execution.hpp
@@ -1640,7 +1640,7 @@ namespace stdexec {
 
       template <class _Awaitable, class _Receiver>
 #if STDEXEC_GCC() && (__GNUC__ > 11)
-      __attribute__((used))
+      __attribute__((__used__))
 #endif
       static __operation_t<_Receiver> __co_impl(_Awaitable __await, _Receiver __rcvr) {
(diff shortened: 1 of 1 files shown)


include/stdexec/execution.hpp
```diff
@@ -1639,6 +1639,9 @@ namespace stdexec {
       }
 
       template <class _Awaitable, class _Receiver>
+#if STDEXEC_GCC() && (__GNUC__ > 11)
+      __attribute__((used))
```
ericniebler, line 1643:
```suggestion
      __attribute__((__used__))
```

ericniebler:
> I'm not entirely sure why this function ends up being discarded by the linker, but this patch cures the problem.

Are you using LTO? And which linker? Clearly there's a linker bug here to be filed.

villevoutilainen:
> > I'm not entirely sure why this function ends up being discarded by the linker, but this patch cures the problem.
> 
> Are you using LTO? And which linker? Clearly there's a linker bug here to be filed.

The regular GNU ld, and I certainly tried to build without LTO. Gonna need to try and reduce a testcase for a linker bug report...

PR #997, opened 2023-07-14
README: fix target name stdexec::stdexec -> STDEXEC::stdexec

felixguendling:
the alias is named STDEXEC:
add_library(STDEXEC::stdexec ALIAS stdexec)


ericniebler:
Thank you!

PR #1015, opened 2023-08-03
use `get_domain` and `transform_sender` to find `static_thread_pool`'s customization of `bulk`

ericniebler:
Finally. FINALLY. `bulk` execution is run in parallel when the current scheduler is `static_thread_pool`. (This wasn't always the case; e.g., in `stdexec::on(pool, sndr | stdexec::bulk(shape, fun))`.)

### DIFF
--- a/examples/hello_world.cpp
+++ b/examples/hello_world.cpp
@@ -26,6 +26,7 @@ using stdexec::sync_wait;
 int main() {
   exec::static_thread_pool ctx{8};
   scheduler auto sch = ctx.get_scheduler();                              // 1
+                                                                         //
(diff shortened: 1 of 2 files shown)


examples/hello_world.cpp
```diff
@@ -26,7 +26,6 @@ using stdexec::sync_wait;
 int main() {
   exec::static_thread_pool ctx{8};
   scheduler auto sch = ctx.get_scheduler();                              // 1
```
ericniebler, line 28:
```suggestion
  scheduler auto sch = ctx.get_scheduler();                              // 1
                                                                         //
```

PR #1020, opened 2023-08-10
Seuqence item types

maikel:
Hi, this change puts the types of senders produced by a sequence into a separate typedef. I've put it into an extra PR to show where it makes a difference.

Now we have a semantic change of `completion_signatures` of a sequence-sender. It now lists the possible completion signatures of the whole sequence operation. Before this change, it was the union of all possible completions of all item operations.


maikel:
@kirkshoop any thoughts about exposing input sender types of calls to `set_next(rcvr, sender)`? In the current design I fail to compute the correct completion signatures in case of customizations.

ericniebler:
Lovely.

PR #1021, opened 2023-08-10
Fix a couple of bugs in `schedule_from`

msimberg:
- Adds a test that triggers the bug in `schedule_from` using `any_sender`. It's unfortunately not clear to me why `any_sender` triggers this but not others (though I didn't try _all_ the other adaptors), but presumably it instantiates some parts (more) eagerly which hits the buggy part? https://github.com/NVIDIA/stdexec/commit/1b238558991ef64557e286f33dd15d001f218480
- Fixes the buggy part in `schedule_from` which was to to still pass the sender type rather than the variant type as the second template parameter to `receiver1` (introduced in #1018). https://github.com/NVIDIA/stdexec/commit/db0ed0cdc401a2d73f213eeb9cf28b136c6525f8
- Adds ADL isolation for the variant template parameter in the various `schedule_from` types, since it may contain user-defined types. https://github.com/NVIDIA/stdexec/commit/1fe63a93e649bb34e3b607c32f0ac9cb89bd238b

### DIFF
--- a/include/stdexec/execution.hpp
+++ b/include/stdexec/execution.hpp
@@ -4938,7 +4938,7 @@ namespace stdexec {
         set_error_t,
         set_stopped_t>>;
 
-    template <class _SchedulerId, class _Variant, class _ReceiverId>
+    template <class _SchedulerId, class _VariantId, class _ReceiverId>
     struct __operation1_base;
 
     // This receiver is to be completed on the execution context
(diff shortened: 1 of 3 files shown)


include/stdexec/execution.hpp
```diff
@@ -5123,8 +5124,10 @@ namespace stdexec {
 
         template <class _Self, class _Receiver>
         using __receiver_t = //
-          stdexec::__t<
-            __receiver1<_SchedulerId, stdexec::__cvref_id<_Self, _Sender>, stdexec::__id<_Receiver>>>;
+          stdexec::__t< __receiver1<
+            _SchedulerId,
+            stdexec::__id<__variant_for_t<_Self, env_of_t<_Receiver>>>,
+            stdexec::__id<_Receiver>>>;
```
ericniebler, line 5130: nice catch, thanks

PR #1042, opened 2023-08-23
MSVC fixes

vasama:
This is a set of fixes targeting intended behaviours of MSVC. I'm holding off on any compiler bug workarounds for now, as fixes for some of the bugs are pending release or the bugs may be fixed before I can find and work around the rest.

### DIFF
--- a/include/exec/__detail/__sender_facade.hpp
+++ b/include/exec/__detail/__sender_facade.hpp
@@ -109,7 +109,7 @@ namespace exec {
       if constexpr (__lacks_transform_sender<_Kernel>) {
         return __mtype<_Sender>{};
       } else {
-        if constexpr (__valid<__tfx_sender_, _Kernel, _Sender, _Env>) {
+        if constexpr (__mvalid<__tfx_sender_, _Kernel, _Sender, _Env>) {
           return __mtype<__tfx_sender_<_Kernel, _Sender, _Env>>{};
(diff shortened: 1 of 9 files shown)


include/exec/__detail/__sender_facade.hpp
```diff
@@ -109,7 +109,7 @@ namespace exec {
       if constexpr (__lacks_transform_sender<_Kernel>) {
         return __mtype<_Sender>{};
       } else {
-        if constexpr (__valid<__tfx_sender_, _Kernel, _Sender, _Env>) {
+        if constexpr (__valid_<__tfx_sender_, _Kernel, _Sender, _Env>) {
```
ericniebler, line 112: what does `__valid` mean on msvc?

ericniebler (approved):

include/exec/__detail/__sender_facade.hpp
ericniebler: Ah. It's in `<sal.h>` as well: https://gist.github.com/caiorss/0678b5efb2aa2eff120b9e9284fdbb0c#file-sal-h-L2448-L2460

ericniebler: Could you rename `__valid` instead to `__mvalid`?

vasama: Done.

ericniebler:
Thanks!

PR #1050, opened 2023-08-27
MSVC fixes

vasama:
Another set of fixes for issues i found while trying to get stdexec to build with MSVC.


ericniebler:
Thanks!

PR #1055, opened 2023-08-28
Feature/cuda reduce

ZelboK:


include/nvexec/stream/reduce.cuh
```diff
@@ -110,20 +128,117 @@ namespace nvexec {
...
+          return __operation_t<_Child, _Data>{
+            (_Child&&) __child, (_Receiver&&) __rcvr_, (_Data&&) __data};
+        }
       };
 
       template <class SenderId, class InitT, class Fun>
       struct sender_t
         : public __algo_range_init_fun::
             sender_t<SenderId, InitT, Fun, sender_t<SenderId, InitT, Fun>> {
+        using Sender = SenderId;
```
ZelboK, line 211: This was using `stdexec::__t<SenderId>` before iirc. I changed it because I think i ran into compilation issues. Useless alias.

PR #1061, opened 2023-09-02
MSVC workarounds

vasama:
With these changes stdexec now builds and tests pass using the latest Visual Studio preview release as well as using Clang-CL.


vasama:
I'll have to see about adding MSVC CI before this is merged.

ericniebler:
I like all of this with the exception of the coroutine hack. It's an impressive hack, but it comes at a high cost, doubling the number of coroutines created. I'm curious how other coroutine libraries like cppcoro deal with this.

IIUC, this issue only effects `at_coroutine_exit()` and the other scope_guard-like features (`on_coroutine_[succeeded|stopped|failed]`). For now, I think I'd prefer stdexec simply didn't offer that functionality on msvc. Or offered it with an opt-in config macro.

maikel:
This is amazing! Thank you for the wonderful work

maikel:
Can we have a ci job for msvc without blocking PRs if it fails? Or just support msvc from now on?

vasama:
That nvc++ CI failure looks like it uncovered a compiler bug. I couldn't quickly reproduce it using CE.

ericniebler:
> How do you feel about this alternative solution which only creates a single extra coroutine per thread?

I feel awed. I'll take it. Ty!

> Can we have a ci job for msvc without blocking PRs if it fails?

I would also like this if it's possible.

vasama:
Looks like the GitHub Windows images don't provide the Visual Studio preview version out of the box. The next minor release probably won't happen for a few months at least. I'll see if i can work around the relevant bugs using the Visual Studio release version.

vasama:
It looks like making this work in the current VS release version would be a lot more work. The next release will probably be some time around november. I'm okay with waiting for that before merging this. Many of these workarounds should also become unnecessary at that time as the compiler bugs i reported are fixed.

ericniebler:
@vasama do you have any objections to my merging this now instead of waiting until November? I don't want to have to keep fixing merge conflicts.

vasama:
No objections if that's what you prefer. I was going to just fix the conflicts and add any new workarounds that might be needed when the next VS version is released. But i'm also fine with merging this and making another PR when that time comes.

PR #1064, opened 2023-09-04
Feature/cuda reduce

ZelboK:


include/nvexec/stream/algorithm_base.cuh
```diff
@@ -91,22 +91,58 @@ namespace nvexec::STDEXEC_STREAM_DETAIL_NS::__algo_range_init_fun {
...
+        _InitT __initT_;
+        STDEXEC_NO_UNIQUE_ADDRESS _Fun __fun_;
+        static constexpr auto __mbrs_ = __mliterals<&__data::__initT_, &__data::__fun_>();
+      };
+      template <class _InitT, class _Fun>
+      __data(_InitT, _Fun) -> __data<_InitT, _Fun>;
+
+      // this is basically the apply function that sender_apply is looking for. 
+      template <class S, class Fn>
+      auto plscompile(S s, Fn f) {
```
ZelboK, line 137: This should be renamed to `apply`. I was conducting an experiment.

include/nvexec/stream_context.cuh
```diff
@@ -75,6 +89,94 @@ namespace nvexec {
...
+
+      template <>
+      auto reconstitute<reduce_t>() const {
+        return [&](auto data, auto sender) {
+          auto initT = stdexec::__nth_member<0>()(data);
+          auto fun = stdexec::__nth_member<1>()(data);
+          auto transformed = reduce_sender_t<decltype(sender), decltype(initT), decltype(fun)>(
+            {}, sender, initT, fun);
+          return transformed;
+        };
```
ZelboK, line 128: Some thoughts:

1) Does the fact that hte compiler no longer complains about hte errors related to `__on::receiver_ref` not having a call operator thus indicate that the tree has been fully transformed, children and alike? I was going off of this assumption. 

transform_sender will rely on this specialization of reconstitute for every sender alongside their children. Now one caveat is... Am I not transforming the entire tree? If that's the case, wouldn't there be an error similar to the 1st point? The fact htat it's looking for an `apply` function, does that not indicate it has transformed the tree properly?

include/nvexec/stream_context.cuh
```diff
@@ -75,6 +89,94 @@ namespace nvexec {
...
+
+      // I got around the error around __on::receiver_ref not having a call operator by removing `__lazy_sender_for<reduce_t> Sender
+      // instantiated when trying to make a stream_op_state from common.cuh
+      template <class Sender, class Env = empty_env>
+      auto transform_sender(Sender&& sndr, const Env& env = {}) const noexcept {
+        return stdexec::__sender_apply(
+          (Sender&&) sndr,
+          [&]<class Tag, class Data, class... Children>(Tag, Data&& data, Children... children) {
+            if constexpr (sizeof...(Children) == 0) {
+              return sndr;
```
ZelboK, line 139: Maybe I shouldn't return sndr.

PR #1070, opened 2023-09-08
Instantiate the debug sender inside connect()

ccotter:
This doesn't build with clang-16, but does with gcc-12. Is there any reason conceptually we shouldn't be able to check the actual vs expected completions within `connect`?


include/stdexec/execution.hpp
```diff
@@ -1914,6 +1930,8 @@ namespace stdexec {
         if constexpr (!enable_receiver<__decay_t<_Receiver>>)
           _PLEASE_UPDATE_YOUR_RECEIVER_TYPE< __decay_t<_Receiver>>();
 
+        static_assert(__check_signatures<__decay_t<_Sender>, env_of_t<__decay_t<_Receiver>>>());
```
ericniebler, line 1933: i don't think you want to be decaying the sender and receiver types here. cvref qualification sometimes matters.

ericniebler:
> This doesn't build with clang-16, but does with gcc-12. Is there any reason conceptually we shouldn't be able to check the actual vs expected completions within `connect`?

I can't think of a reason why this would be problematic in principle.

include/nvexec/multi_gpu_context.cuh
```diff
@@ -66,7 +66,7 @@ namespace nvexec {
             if (op.status_ == cudaSuccess) {
               continuation_kernel<<<1, 1, 0, op.stream_>>>(std::move(op.rec_), stdexec::set_value);
             } else {
-              continuation_kernel<cudaError_t><<<1, 1, 0, op.stream_>>>(
+              continuation_kernel<<<1, 1, 0, op.stream_>>>(
```
ccotter, line 69: Some of the refactors of `continuation_kernel` which reordered the template parameters seem to have missed this one. The first template parameter is `Receiver`.

ericniebler: I'm pulling this change out into its own commit, since it's unrelated.

PR #1088, opened 2023-09-24
Implement parallelism and cardinality

maikel:
This PR introduces 2 queries for sequence senders, which are useful for optimizations within algorithms.

### DIFF
--- a/include/exec/sequence/empty_sequence.hpp
+++ b/include/exec/sequence/empty_sequence.hpp
@@ -50,6 +50,13 @@ namespace exec {
           return stdexec::__t<__operation<stdexec::__id<__decay_t<_Rcvr>>>>{
             static_cast<_Rcvr&&>(__rcvr)};
         }
+
+        template <__decays_to<__t> _Self>
+        friend auto tag_invoke(get_env_t, _Self&&) noexcept {
+          return make_env(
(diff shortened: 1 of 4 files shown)


maikel:
I think parallelism and cardinality should not be forwarded queries. But `join_env` works only with forwarding. Is that an unnecessary requirement of `join_env` ?

ericniebler:
> I think parallelism and cardinality should not be forwarded queries. But `join_env` works only with forwarding. Is that an unnecessary requirement of `join_env` ?

Possibly. The use I had in mind for `join_env` is when a sender adaptor is augmenting its child sender's environment, in which case the exclusion of non-forwarding queries is correct. 

But if you are just cobbling together a sender's environment from bits and bobs, none of which are from a child sender, then the current behavior is not what you want. In that case, perhaps we want a variant of `join_env` that doesn't exclude non-forwarding queries.

maikel:
Thats a very fair argument. I will provide a helper for what I really need

ericniebler:
> I think parallelism and cardinality should not be forwarded queries. But `join_env` works only with forwarding. Is that an unnecessary requirement of `join_env`?

@maikel stdexec now has the `stdexec::env<>` template for joining environments without any special handling for forwarding queries.

Issue #1111, opened 2023-10-18
example.io_uring isn't getting compiled


lukester1975:
Since c9df036823bd4e5719e4aa491c065e564bdb0f3a (`stdexect_examples` typo). Presumably as a result, it seems to have bitrotted a bit :)

In https://github.com/lukester1975/stdexec/tree/fix-uring-example I've put a cmake fix and a few other hacks to try to get the example working. Happy to separate out the cmake fix in to a PR. I'm sure the rest is largely nonsense...

Presumably `context.run()` should be `context.run_until_stopped()` after 5bb97462362a5ebac31fcf49c9fb69f259d22862?

The example still seems a bit racy. Sometimes it works:
```
Hello, 1!
Hello, 2, stopped.
Hello, 2!
Hello, 1, stopped.
Hello, 0!
Hello, 1!
Hello, 2!
Stop it!
Requesting context stop
Requesting context2 stop
Hello, stopped.
Hello, stopped.
Joining io_thread
Joining io_thread2
Threads finished
The context is stopped!
The context is stopped!
Restarting context
Hello, 2!
Hello, schedule_at!
Finished
```

Other times:

```
Hello, 1!
Hello, 2, stopped.
Hello, 2!
Hello, 1, stopped.
Hello, 0!
Hello, 1!
Hello, 2!
Stop it!
Requesting context2 stop
Hello, stopped.
Hello, stopped.
Joining io_thread
```

I've not looked at that in any depth as it made my brain melt. It did seem `context.reset()` was required to ever get it to work, though.

Maybe I should open separate issue for this, but should `run_until_stopped()` restart `IORING_ENTER_GETEVENTS` if it gets `EINTR`? Otherwise `run_until_stopped()` needs wrapping in a loop and try/catch, presumably? Gets interrupted a lot when debugging.

Kernel is 6.5.6-200.fsync.fc38.x86_64 from nobara. Perhaps I should double check with something more vanilla...

Thanks!

maikel:
Thank you for this! 

I will look into it and fix it. Regarding EINTR you are correct. In fact I have a fix (and some improvements) in the pipeline here 

https://github.com/maikel/senders-io/blob/static_io_pool/source/sio/io_uring/io_uring_context.cpp#L351

I haven't managed yet to merge it yet but I will try do so today.

maikel:
[...]

PR #1114, opened 2023-10-19
Fix io uring example

maikel:
Addresses issue #1111. There still seems to be some unreported data race with cancellation.


maikel:
@lukester1975 Can you please check whether this fixes all of your issues? Thank you!

lukester1975:
Yeah I think so; the example hasn't failed to complete for me after running in a loop for 5 mins.

Thanks

ericniebler:
thank you!

PR #1119, opened 2023-10-22
Implement lock free work stealing for static_thread_pool

maikel:
This PR tracks an improved implementation for `static_thread_pool`

# Design #

The high level design of the implementation is as follows:

- Each worker has one local intrusive list of pending ops, multiple atomic intrusive lists for "remote ops" and a finite block based workstealing deque as described in https://www.usenix.org/system/files/osdi23-wang-jiawei.pdf
- thiefs can only steal from the BWoS deque
- to reduce contention upon submission, each submitter thread gets its own submission queue (the atomic intrusive list)

# Benchmark 1: Remote schedules #

We have N producer threads that schedule 10,000,000 operations on a thread pool of N threads. Each thread also gets a preallocated memory buffer to factor out time spent in dynamic memory allocation. In the end, we are interested in throughput Ops/s. The benchmark is repeated 100x for each tested scheduler.

## Results ##

I made the benchmark only on my laptop which has 8 cores and only one socket.

scheduler_comparison

The maximum throughput for each scheduler in ops/sec

| tbb | old | new | new (with bulk schedules) |
| ----- | ---- | ------- | ------------------------------------- |
| 8.92e+06 | 1.1e+07 | 3.36e+07 | 6.58e+07   |

I suspect that this benchmark is not well suited for the tbb task arena.

# Benchmark 2: Nested schedule #

We have a thread pool of N threads and schedule N work items that each schedule 10,000,000 / N operations on itself. Each thread also gets a preallocated memory buffer to factor out time spent in dynamic memory allocation. In the end, we are interested in throughput Ops/s. The benchmark is repeated 100x for each tested scheduler.

## Result ##

nested_benchmark

The maximum throughput for each scheduler in ops/sec

| tbb | old | new | new (with bulk schedules)|
| ----- | ---- | ------- | ------------------------------------ |
| 4.43e+07| 2.34e+07| 9.49e+07 | 1.27e+08 |

# TODO #

- [ ] comparison with other thread pools, e.g. taskflow
- [x] other micro benchmarks that use more common usage patterns
- [ ] show some improvement in an application
- [x] improve implementation for cases such as scheduling from within a worker thread
- [ ] add numa awareness


ericniebler:
@maikel can you make sure all new files have the right copyright header?

maikel:
> @maikel can you make sure all new files have the right copyright header?

OK, I think I got every file covered

ericniebler:
this looks AWESOME. 
thank-youuuuuuu

PR #1129, opened 2023-11-01
Fix memory ordering in lifo_queue put

maikel:
The error is in a function that is not being used in the `static_thread_pool` implementation. The memory ordering in the member function `bulk_put`, which is being used, is correct.

The `put` function needs a similar release on the `tail_` member variable as it is done in `bulk_put`. There is an analogous acquire operation in the `steal()` member function.


maikel:
CI failure seems like a spurious infrastructure failure to me.

PR #1144, opened 2023-11-16
give generic sender `__sexpr` a generic op state and receiver; inline-able op states

ericniebler:


include/stdexec/__detail/__basic_sender.hpp
```diff
@@ -73,10 +77,239 @@ namespace stdexec {
...
+      using is_receiver = void;
+      using __sexpr = _Sexpr;
+      using __index = _Idx;
+
+      // A pointer to the parent operation state, which contains the one created with
+      // this receiver.
+      __op_state<_Sexpr, _Receiver>* __op_;
+
+      template <class _ChildSexpr, class _ChildReceiver>
+      static __receiver __from_op_state(__op_state<_ChildSexpr, _ChildReceiver>* __child) noexcept {
```
lewissbaker, line 110: I think that if we want to make this generally available in the library then it needs to be an official customisation point / protocol that user-defined operation-states/receivers can implement, and not be hidden behind stdlib ugly names. i.e. define a concept for `inlinable_receiver_in<R, Op>` that requires `receiver<r>` and `static R R::from_op_state(Op*) noexcept;`.

Then have a public `op_state_base<Derived, R>` base class that any user's op-state class can inherit from and get the benefit of op-state inlining.

ericniebler: I don't want this to be a generally available customization point, esp considering that it relies on UB. At some point, compilers might start rejecting this and I have to abandon the effort. That's hard if it's a public interface with users.

lewissbaker: Can we find a way that would allow this to be not-UB?
e.g. by using base-classes to hold child operation-states for which we can validly cast from base-class (child) to derived class (parent)

include/stdexec/__detail/__basic_sender.hpp
```diff
@@ -73,10 +77,239 @@ namespace stdexec {
...
+
+      template <__completion_tag _Tag2, class... _Args>
+      STDEXEC_ATTRIBUTE((always_inline))
+      friend void tag_invoke(_Tag2, __receiver&& __self, _Args&&... __args) noexcept {
+        __self.__op_->__complete(_Tag2(), (_Args&&) __args...);
+      }
+
+      template <same_as<get_env_t> _Tag2>
+      STDEXEC_ATTRIBUTE((always_inline))
+      friend auto tag_invoke(_Tag2, const __receiver& __self) noexcept -> env_of_t<_Receiver> {
```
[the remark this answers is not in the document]
ericniebler: This is another "gift" of `tag_invoke`. All overloads are considered, even ones with mismatching tags, because maybe the compiler finds a conversion from `connect_t` to `get_env_t`, say. So the compiler charges ahead computing the return type and then hard fails for one reason or another.

We avoid that by deducing the tag type and aborting early if it isn't the one we're looking for. Ridiculous, I know.

include/stdexec/__detail/__basic_sender.hpp
```diff
@@ -73,10 +77,239 @@ namespace stdexec {
...
+      _Parent* __parent = (_Parent*) &__buf;
+      const std::ptrdiff_t __offset = (char*) &(__parent->*__mbr_ptr) - __buf;
+      return (_Parent*) ((char*) __child - __offset);
+    }
+
+    template <class _Sexpr, class _Receiver>
+    struct __op_state;
+
+    template <class _Receiver, class _Sexpr, class _Idx>
+    struct __receiver {
```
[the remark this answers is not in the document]
ericniebler: We reference `_Sexpr` via the `__sexpr` alias when deciding whether or not to perform the optimization:

```c++
    template <class _Receiver>
    using __sexpr_connected_with =
      __mapply<
        __mbind_front_q<__m_at, typename _Receiver::__index>,
        typename __call_result_t<__impl_of<typename _Receiver::__sexpr>, __cp, __get_meta>::__child>;
```

This avoids an interesting bug where there is a mismatch between the op state to which the receiver is pointing, and the immediate parent of the op state being initialized.

include/stdexec/__detail/__basic_sender.hpp
ericniebler: I don't know.

lewissbaker: I've been noodling on this a little - maybe something along the lines of this?
https://godbolt.org/z/MM59PTvK4

PR #1145, opened 2023-11-17
gcc: fix get_completion_scheduler for -Og

williamspatrick:


ericniebler:
the summary for this pr is inaccurate, since it's only the one fix.

/ok to test

williamspatrick:
I’ll fix that. I did something funny with the rebase against the content that merged yesterday and then `gh pr` decided to write this.

PR #1151, opened 2023-11-22
Add Windows CI builds

vasama:
* Added CI builds on Windows using MSVC, Clang and Clang-CL.
* Disabled the Clang builds due to a baffling compiler bug causing test failures.


ericniebler:
Cool, thanks!

PR #1160, opened 2023-11-30
Fixing exec::__atomic_intrusive_queue

sithhell:
exec::__atomic_intrusive_queue::pop_all and
exec::__atomic_intrusive_queue::pop_all_reversed used a simple atomic exchange which might lead to a race with a parallel insertion.

This patch fixes that by introducing a CAS loop to account for potential parallel insertions.

### DIFF
--- a/include/exec/__detail/__atomic_intrusive_queue.hpp
+++ b/include/exec/__detail/__atomic_intrusive_queue.hpp
@@ -19,62 +19,68 @@
 #include "../../stdexec/__detail/__intrusive_queue.hpp"
 
 namespace exec {
-  template <auto _NextPtr>
-  class __atomic_intrusive_queue;
+    template<auto _NextPtr>
+    class __atomic_intrusive_queue;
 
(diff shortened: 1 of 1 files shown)


ericniebler:
@sithhell Thanks for this. Do you mind running this change through clang-format? Thanks.

/ok to test

ericniebler (changes requested):
~~Needs to be clang-formatted~~

ericniebler (approved):

sithhell:
@ericniebler OK. I guess I have to fix my editor. I actually formatted before submitting :/
Sorry for the extra trouble.

lewissbaker:
@sithhell Can you elaborate on what the potential race would be here with parallel insertions?

The insertions use a compare-exchange so that they ensure that the item that is pushed has the correct next-pointer.
But acquiring the entire list with an exchange-with-nullptr seems (on the surface) like that is all that is necessary for the dequeue operation.

PR #1185, opened 2024-01-02
Add recursive fibonacci example/benchmark (RFC)

igchor:
I've noticed there are not too many benchmarks in this repo. I wanted to port some of the existing ones so a comparison can be made with 'native' implementations. This benchmark is based on https://github.com/oneapi-src/oneTBB/tree/master/examples/migration/recursive_fibonacci 

This is, of course, a very inefficient way to calculate Fibonacci numbers, but it is quite a good benchmark for use cases where new tasks are created recursively. I've implemented this using a custom sender. This is probably not the most idiomatic way but I couldn't figure out how to achieve the same performance by just using existing algorithms (and without start_detached) - perhaps you have some ideas? Any feedback is welcomed.

When running on my computer (on i9-7940X) I get:
```
# ./examples/example.benchmark.fibonacci 16 46 10 static
Avg time: 4377ms. Result: 1836311903
# ./examples/example.benchmark.fibonacci 16 46 10 tbb
Avg time: 763ms. Result: 1836311903
```

When running the 'native' tbb example (from the tbb repo) I get:
```
Fibonacci two tasks impl N = 1836311903 Avg time = 311 ms
Fibonacci single task impl N = 1836311903 Avg time = 310 ms
```


maikel:
I think this highlights that the any sender template class should probably have a way to customize the allocator. There are two allocations here, one for the sender and one for the operation.

I havent looked into this particular benchmark but I have implented and profiled this example stdexec in the past and the allocations dominate everything.

Note, IIRC tbb tasks use an internal pooling to allocate memory

maikel:
I think this highlights that the any sender template class should probably have a way to customize the allocator. There are two allocations here, one for the sender and one for the operation. I propose to use a template argument to specify the allocator for the type erased sender and maybe consider the allocator in the receivers environment to chose the allocator for the resulting operation state.

I havent looked into this particular benchmark but I have implented and profiled this example stdexec in the past and the allocations dominate everything.

Note, IIRC tbb tasks use an internal pooling to allocate memory.

ericniebler:
I can confirm your numbers. The `static_thread_pool` performs very badly here relative to the TBB thread pool. I wonder if @maikel is interested in this benchmark.

ericniebler:
@igchor if you agree with the changes in this PR, pls merge. https://github.com/igchor/stdexec/pull/1

igchor:
> @igchor if you agree with the changes in this PR, pls merge. [igchor#1](https://github.com/igchor/stdexec/pull/1)

Done. Interestingly, after those changes (specifically calling 'stdexec::on' for children vs calling it for when_all) the difference between tbb and static thread pools is lower (tbb sheduler is 1.5x slower) but I think this is because the existing implementation is not really optimal. I will probably submit some improvements to the tbb sheduler after some more testing.

maikel:
> > @igchor if you agree with the changes in this PR, pls merge. [igchor#1](https://github.com/igchor/stdexec/pull/1)
> 
> Done. Interestingly, after those changes (specifically calling 'stdexec::on' for children vs calling it for when_all) the difference between tbb and static thread pools is lower (tbb sheduler is 1.5x slower) but I think this is because the existing implementation is not really optimal. When tweaking how the tasks are submitted in tbb sheduler I can achieve ~450 ms for the configuration I mentioned in the first comment. I will probably submit the improvement to the tbb sheduler after some more testing.
> 
> Also, I think it confirms what Eric wrote, that the performance difference is not necessarily related to the allocations (but probably to the scheduling algorithm/overhead).

Indeed. There is some logic regarding the submission queues, which depends on the thread id. Each scheduler caches a pointer to an internal SPSC queue and upon submission the thread id will be used to check whether the current thread is allowed to use this submission queue. If the cached pointer does not match the current thread id the correct SPSC will be searched for, which might cause the overhead here.

This benchmark is helpful for me to improve on the algorithm. Thank you for this.

maikel:
Hi,

I took a closer look today and from what I see the problem is somewhere else. For this benchmark the load balancer is basically not working and work is being done on 2 or 1 threads independent on the number of threads in the pool.

One simple solution is to change the `bwos_params` in the constructor of `static_thread_pool`. The default values are 

```cpp
struct bwos_params {
  std::size_t num_blocks = 8,
  std::size_t block_size = 1024
};
``` 

In the bwos algorithm one has num_blocks many buckets and the owner of the workstealing bwos queues claims ownership of one bucket at the time. While having this ownership no other threads can steal from this bucket.

The depth of the fibonacci tree is not big enough (it is around 15-16) until the serial algorithm kicks in. So the buckets never get full and no workstealing can happen.

I ask you to play around with the numbers. I tried `num_blocks = 32` and `block_size = 4` and I got good results.

igchor:
@maikel I've tried the parameters you suggested, and the performance is indeed much better. (better than the existing tbb thread pool and similar to a version with my tweaks). I think it would be good to have a few more benchmarks/examples to cover different parallel patterns and see what's the performance like for them.

maikel:
> @maikel I've tried the parameters you suggested, and the performance is indeed much better. (better than the existing tbb thread pool and similar to a version with my tweaks). I think it would be good to have a few more benchmarks/examples to cover different parallel patterns and see what's the performance like for them.

Thats a great idea.

ericniebler:
The numbers are looking good. I'd like to merge this. Objections?

maikel:
approved

PR #1206, opened 2024-01-10
change default bwos_params for static_thread_pool

ericniebler:


include/exec/static_thread_pool.hpp
```diff
@@ -42,8 +42,8 @@
 
 namespace exec {
   struct bwos_params {
-    std::size_t numBlocks{8};
-    std::size_t blockSize{1024};
+    std::size_t numBlocks{32};
+    std::size_t blockSize{4};
```
ericniebler, line 46:
```suggestion
    std::size_t blockSize{8};
```

PR #1211, opened 2024-01-17
Add basic implementation for `system_context`

lucteo:
This adds the basic implementation of `system_context` according to P2079R3.

Making Lee's implementation work in stdexec, adding a few extra tests.


include/exec/system_scheduler.hpp
```diff
@@ -0,0 +1,663 @@
...
+  __exec_system_bulk_operation_state_impl& operator= (__exec_system_bulk_operation_state_impl&&) = delete;
+
+
+  void start() noexcept override {
+    stdexec::start(pool_operation_state_);
+  }
+
+  __exec_system_receiver recv_;
+  __exec_system_bulk_function_object bulk_function_;
+  decltype(__exec_pool_operation_state(std::declval<__exec_system_bulk_operation_state_impl*>(), std::declval<__exec_pool_sender_t>(), __exec_system_bulk_shape{}, __exec_system_bulk_function_object{}))
```
ericniebler, line 246:
```suggestion
stdexec::__result_of<__exec_pool_operation_state, __exec_system_bulk_operation_state_impl*, __exec_pool_sender_t, __exec_system_bulk_shape, __exec_system_bulk_function_object>
```

include/exec/system_scheduler.hpp
```diff
@@ -0,0 +1,663 @@
...
+// TODO: Potentially move to custom vtable implementations
+struct __exec_system_context_interface {
+  virtual __exec_system_scheduler_interface* get_scheduler() noexcept = 0;
+};
+
+// bulk function for scheduler to transmit from, will wrap actual function stub stored in real type
+using __exec_system_bulk_shape = long;
+using __exec_system_bulk_fn = void(void*, __exec_system_bulk_shape);
+struct __exec_system_bulk_function_object {
+  void* fn_state = nullptr;
```
ericniebler, line 40: we should decide if we're ugly-ifying identifiers in this header or not and stick to it. I would prefer all identifiers to be ugly-ified.

include/exec/system_scheduler.hpp
```diff
@@ -0,0 +1,663 @@
...
+
+
+
+
+
+
+// Low-level APIs
+// Phase 2 will move to pointers and ref counting ala COM
+// Phase 3 will move these to weak symbols and allow replacement in tests
+// Default implementation based on static_thread_pool
```
ericniebler, line 80: Are phases 2 and 3 planned for this PR, or will they happen in a future PR?

include/exec/system_scheduler.hpp
```diff
@@ -0,0 +1,663 @@
...
+struct __exec_system_sender_impl : public __exec_system_sender_interface {
+  __exec_system_sender_impl(
+        __exec_system_scheduler_impl* scheduler, __exec_pool_sender_t&& pool_sender) :
+      scheduler_{scheduler}, pool_sender_(std::move(pool_sender)) {
+
+  }
+
+  __exec_system_operation_state_interface* connect(__exec_system_receiver recv) noexcept override {
+    return
+      new __exec_system_operation_state_impl(std::move(pool_sender_), std::move(recv));
```
ericniebler, line 179: there are a slew of `new`'s and no `delete`s or `unique_ptr`s. are all the allocations intended to leak?

ericniebler (changes requested):
i'm excited for this. is this PR still evolving (phase 2, phase 3 as in the comments)? also, i'd like a plan about adding queries without breaking ABI.

and per request from @sean-parent, what is the plan for integrating this with system event loops? imagine a single-threaded embedded system running an event loop. how does that interoperate with this?

finally, i'd like to add phase 4: back this with windows thread pool on windows, or io_uring (?) on Linux, or GCD on MacOS. :-)))

lucteo:
> i'm excited for this. is this PR still evolving (phase 2, phase 3 as in the comments)? also, i'd like a plan about adding queries without breaking ABI.
> 
> and per request from @sean-parent, what is the plan for integrating this with system event loops? imagine a single-threaded embedded system running an event loop. how does that interoperate with this?
> 
> finally, i'd like to add phase 4: back this with windows thread pool on windows, or io_uring (?) on Linux, or GCD on MacOS. :-)))

My plan for this PR was just to push what Lee had already done + some fixes + adding more tests to cover the requirements from the paper. Phases 2+ will be coming in following PRs

include/exec/system_scheduler.hpp
lucteo: In a future PR. I'll try to keep the PRs small and focused, and this one is already pretty big.

include/exec/system_scheduler.hpp
lucteo: I would mark all these as things still left to be done. The implementation is somehow between Phase 1 and Phase 2 (using Lee's plan).

include/exec/system_scheduler.hpp
```diff
@@ -0,0 +1,663 @@
...
+// Phase 1 implementation, single implementation
+// TODO: Make a weak symbol and replace in a test
+static __exec_system_context_impl* __get_exec_system_context_impl() {
+  static __exec_system_context_impl impl_;
+
+  return &impl_;
+}
+
+// TODO: Move everything above here to a detail header and wrap in a
+// namespace to represent extern "C"
```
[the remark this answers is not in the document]
lucteo: If that's ok with you, I'll fix all these TODOs in later PRs

include/exec/system_scheduler.hpp
```diff
@@ -0,0 +1,663 @@
...
+        std::move(rec),
+        [](auto& op){
+          __exec_system_receiver receiver_impl{
+            &op.recv_,
+            [](void* cpp_recv){
+              stdexec::set_value(std::move(*static_cast<R*>(cpp_recv)));
+            },
+            [](void* cpp_recv){
+              stdexec::set_stopped(std::move(*static_cast<R*>(cpp_recv)));
+            }};
```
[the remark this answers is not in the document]
lucteo: `set_error` is not promised by the paper. There was an error here and I removed this case. It seems that I should have removed other occurrences of `set_error`; will remove them now.

include/exec/system_scheduler.hpp
ericniebler: I'm fine leaving phases 2 and 3 to later. i'm not fine with known memory leaks in `main`.

include/exec/system_scheduler.hpp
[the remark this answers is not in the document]
LeeHowes: I forget the deal with set_error in the paper. I think I just didn't have a good reason to put it here, but the expectation would be that a receiver would be capable of receiving errors. Whether we want to define some specific potential failure types or not I don't know.

include/exec/system_scheduler.hpp
```diff
@@ -0,0 +1,663 @@
...
+
+inline void tag_invoke(stdexec::set_stopped_t, __exec_system_bulk_pool_receiver&& recv) noexcept {
+  __exec_system_receiver &system_recv = recv.os_->recv_;
+  recv.os_->recv_.set_stopped(&(system_recv.cpp_recv_));
+}
+
+inline void tag_invoke(stdexec::set_error_t, __exec_system_bulk_pool_receiver&& recv, std::exception_ptr ptr) noexcept {
+  __exec_system_receiver &system_recv = recv.os_->recv_;
+  recv.os_->recv_.set_error(&(system_recv.cpp_recv_), &ptr);
+}
```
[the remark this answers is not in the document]
LeeHowes: My implementation was initial and hacky, I wasn't aiming to follow the latest design guidelines.

LeeHowes:
Yes the point I left this at was really early. As I wasn't writing it in the main repo I felt it was less work to leave memory leaks in and fix the APIs properly, rather than to make something correct but suboptimal first and evolve that to optimal correctness. The constraints may be different in stdexec.

LeeHowes:
> i'm excited for this. is this PR still evolving (phase 2, phase 3 as in the comments)? also, i'd like a plan about adding queries without breaking ABI.
> 
> and per request from @sean-parent, what is the plan for integrating this with system event loops? imagine a single-threaded embedded system running an event loop. how does that interoperate with this?

This should fall out of the replacement mechanism. We swap out the entire implementation, with no API changes, with an event loop. Driving of the event loop is something you do directly on the implementation because it's very specific, and not
exposed through this API. 

Now, even loop in the sense of supporting timers etc is slightly different, and I think there we need to be clear that we can evolve the ABI safely behind the plugin mechanism. This is where I lost confidence in my ability to spend the time on it to get it right.

> finally, i'd like to add phase 4: back this with windows thread pool on windows, or io_uring (?) on Linux, or GCD on MacOS. :-)))

Yes, same applies. I don't think that is at all complicated from the POV of the paper. Once we have a plugin API, we just hide all that inside. Demonstrating that we can do so though is important. It can be just a test case in the repo.

sean-parent:
If this is to support P2079R3, then I'd like to see an implementation layered earlier on some of the possible system thread pools. There is so much in that proposal that is not specified. What requirements are placed on the system scheduler? Most system schedulers support priorities and are elastic (i.e., blocking in the system thread pool will spin up additional threads to some limit).

> If the underlying system_context is unable to make progress on work created through system_scheduler instances, and the sender retrieved from the scheduler is connected to a receiver that supports the get_delegatee_scheduler query, work may scheduled on the scheduler returned by get_delegatee_scheduler at the time of the call to start, or at any later point before the work completes.

If I understand correctly, this is saying tasks can be stolen and executed on the current thread when work is scheduled. This may mean that work expected to happen concurrently on another thread is executed in the same thread context. In practice, this is a recipe for deadlocks on odd reentrancy bugs of code holding a lock while scheduling work. I did this in the portable STLab scheduler and had to pull it out and instead make the pool elastic. Unfortunately, the answer to "don't hold a mutex while scheduling work" wasn't viable in a large existing code base.

This approach also caused "Aw Snaps" on a web app where tasks intended to be scheduled async would get stolen in the main js run loop and exceed the browser time limit before returning to the main thread.

There is a good overview of the Window thread pool design [here](https://youtu.be/Z86b3Rd09sE?si=U4095X-OISWqejAj). Apple's GCD is a very similar model.

include/exec/system_scheduler.hpp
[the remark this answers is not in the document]
lucteo: After thinking some more on this, I think we need to have `set_error` here. Otherwise, how would the the implementation trigger some runtime errors (lack of memory, failure to allocate resources, etc.)

@ericniebler : at some point, long long time ago, there was a push to remove requirements for needing `set_error`. I take it that we do not want in P2300 to remove `std_error` as a requirement for any receiver, right?

include/exec/system_scheduler.hpp
```diff
@@ -0,0 +1,685 @@
...
+// For the default implementation, test will override
+#include "exec/static_thread_pool.hpp"
+
+
+struct __exec_system_scheduler_interface;
+struct __exec_system_sender_interface;
+struct __exec_system_scheduler_impl;
+struct __exec_system_sender_impl;
+
+// Virtual interfaces to underlying implementations for initial simplicit
```
ericniebler, line 28:
```suggestion
// Virtual interfaces to underlying implementations for initial simplicity
```

include/exec/system_scheduler.hpp
```diff
@@ -0,0 +1,685 @@
...
+
+  __exec_system_sender_interface*
+    bulk(__exec_system_bulk_shape __shp, __exec_system_bulk_function_object __fn) override;
+
+  stdexec::forward_progress_guarantee get_forward_progress_guarantee() const override {
+    return stdexec::forward_progress_guarantee::parallel;
+  }
+
+  bool equals(const __exec_system_scheduler_interface* __rhs) const override {
+    auto __rhs_impl = dynamic_cast<const __exec_system_scheduler_impl*>(__rhs);
```
ericniebler, line 102: future wishlist: support compiling without RTTI

include/exec/system_scheduler.hpp
```diff
@@ -0,0 +1,685 @@
...
+
+struct __exec_system_pool_receiver {
+  using receiver_concept = stdexec::receiver_t;
+
+  friend void tag_invoke(stdexec::set_value_t, __exec_system_pool_receiver&&) noexcept;
+
+  friend void tag_invoke(stdexec::set_stopped_t, __exec_system_pool_receiver&&) noexcept;
+
+  friend void
+    tag_invoke(stdexec::set_error_t, __exec_system_pool_receiver&&, std::exception_ptr) noexcept;
```
ericniebler, line 119: These should be defined in situ to remain hidden friends.

```suggestion
  void set_value() noexcept;
  void set_stopped() noexcept;
  void set_error(std::exception_ptr) noexcept;

  friend void tag_invoke(stdexec::set_value_t, __exec_system_pool_receiver&& __self) noexcept {
    __self.set_value();
  }

  friend void tag_invoke(stdexec::set_stopped_t, __exec_system_pool_receiver&& __self) noexcept {
    __self.set_stopped();
  }

  friend void
    tag_invoke(stdexec::set_error_t, __exec_system_pool_receiver&& __self, std::exception_ptr __ptr) noexcept {
      __self.set_error(std::move(__ptr));
    }
```

include/exec/system_scheduler.hpp
```diff
@@ -0,0 +1,685 @@
...
+  void start() noexcept override {
+    stdexec::start(__pool_operation_state_);
+  }
+
+  __exec_system_receiver __recv_;
+  stdexec::connect_result_t<__exec_pool_sender_t, __exec_system_pool_receiver>
+    __pool_operation_state_;
+};
+
+inline void tag_invoke(stdexec::set_value_t, __exec_system_pool_receiver&& __recv) noexcept {
```
ericniebler, line 153:
```suggestion
inline void __exec_system_pool_receiver::set_value() noexcept {
```

include/exec/system_scheduler.hpp
```diff
@@ -0,0 +1,685 @@
...
+  __exec_system_receiver& __system_recv = __recv.__os_->__recv_;
+  __system_recv.set_value(__system_recv.__cpp_recv_);
+}
+
+inline void tag_invoke(stdexec::set_stopped_t, __exec_system_pool_receiver&& __recv) noexcept {
```
ericniebler, line 158:
```suggestion
inline void __exec_system_pool_receiver::set_stopped() noexcept {
```

include/exec/system_scheduler.hpp
```diff
@@ -0,0 +1,685 @@
...
+  __exec_system_receiver& __system_recv = __recv.__os_->__recv_;
+  __recv.__os_->__recv_.set_stopped(&(__system_recv.__cpp_recv_));
+}
+
+inline void tag_invoke(
+  stdexec::set_error_t,
+  __exec_system_pool_receiver&& __recv,
+  std::exception_ptr __ptr) noexcept {
```
ericniebler, line 166:
```suggestion
inline void __exec_system_pool_receiver::set_error(
  std::exception_ptr __ptr) noexcept {
```

include/exec/system_scheduler.hpp
```diff
@@ -0,0 +1,685 @@
...
+  using receiver_concept = stdexec::receiver_t;
+
+  friend void tag_invoke(stdexec::set_value_t, __exec_system_bulk_pool_receiver&&) noexcept;
+
+  friend void tag_invoke(stdexec::set_stopped_t, __exec_system_bulk_pool_receiver&&) noexcept;
+
+  friend void tag_invoke(
+    stdexec::set_error_t,
+    __exec_system_bulk_pool_receiver&&,
+    std::exception_ptr) noexcept;
```
ericniebler, line 205: Similarly define these in situ.

include/exec/system_scheduler.hpp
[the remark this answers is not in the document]
ericniebler: i don't want to merge to main any symbols in the global namespace.

include/exec/system_scheduler.hpp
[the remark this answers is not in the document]
ericniebler: there is no longer a requirement for a receiver to have a set_error.

ericniebler (changes requested):
it's ok for the other changes to wait, but the identifiers in the global namespace and the memory leaks need to be fixed before i can feel good about merging this.

ericniebler:
The nvc++ test failure is odd:

```
test 607
        Start 607: a copied scheduler is equal to the original

607: Test command: /home/coder/stdexec/build/test/test.stdexec "a copied scheduler is equal to the original"
607: Working Directory: /home/coder/stdexec/build/test
607: Test timeout computed to be: 60
607: Filters: a copied scheduler is equal to the original
607: ===============================================================================
607: All tests passed (1 assertion in 1 test case)
607: 
607: pure virtual method called
607: terminate called without an active exception
607/715 Test #607: a copied scheduler is equal to the original ...................................................................................Subprocess aborted***Exception:   0.27 sec
```

The runtime failure is "`pure virtual method called`", but there are no `virtual` functions in the implementation. @lucteo can you think of anything that would cause this?

include/exec/system_context.hpp
```diff
@@ -0,0 +1,407 @@
...
+
+  /// A scheduler that can add work to the system context.
+  class system_scheduler {
+   public:
+    system_scheduler() = delete;
+
+    /// Returns `true` iff `*this` refers to the same scheduler as `__rhs`.
+    bool operator==(const system_scheduler& __rhs) const noexcept {
+      return __scheduler_ == __rhs.__scheduler_;
+    }
```
ericniebler, line 74:
```suggestion
    bool operator==(const system_scheduler&) const noexcept = default;
```

include/exec/system_context.hpp
```diff
@@ -0,0 +1,407 @@
...
+    /// Schedules new bulk work, calling `__fun` with the index of each chunk in range `[0, __size]`,
+    /// and the value(s) resulting from completing `__previous`; returns a sender that completes
+    /// when all chunks complete.
+    template <stdexec::sender __S, std::integral __Size, class __Fn>
+    friend system_bulk_sender<__S, __Size, __Fn> tag_invoke(
+      stdexec::bulk_t,
+      const system_scheduler& __self,
+      __S&& __previous,
+      __Size __size,
+      __Fn __fun) noexcept;
```
ericniebler, line 98: We need to switch to the new algorithm customization scheme, giving the `system_scheduler` a domain and customizing `bulk` there.

ericniebler (commented):
This is looking very good.

include/exec/system_context.hpp
```diff
@@ -0,0 +1,433 @@
...
+    /// Invoked when the previous sender completes with a value to trigger multiple operations on the system scheduler.
+    template <class... __As>
+    friend void tag_invoke(
+      stdexec::set_value_t,
+      __bulk_intermediate_receiver&& __self,
+      __As&&... __as) noexcept {
+      // Heap allocate input data in shared state as needed
+      auto __size = sizeof(std::tuple<__As...>);
+      __self.__state_.__arguments_data_ = new char[__size];
+      new (__self.__state_.__arguments_data_) std::tuple<__As...>{__as...};
```
ericniebler, line 228: why are we heap-allocating this? why can't it be part of the operation state?

include/exec/system_context.hpp
```diff
@@ -0,0 +1,433 @@
...
+      // TODO: figure out from the completion signal the size of the input data, and preallocate this in the shared state.
+
+      auto __type_erased_fn = [](void* __state_, long __idx) {
+        auto* __state = static_cast<__bulk_state_t*>(__state_);
+        std::apply(
+          [&](auto&&... __args) { __state->__snd_.__fun_(__idx, __args...); },
+          *reinterpret_cast<std::tuple<__As...>*>(__state->__arguments_data_));
+      };
```
ericniebler, line 236: so there are going to be `shape` indirect function calls? i'm curious if there's a better way.

include/exec/system_context.hpp
```diff
@@ -0,0 +1,433 @@
...
+  /// The sender used to schedule bulk work in the system context.
+  template <stdexec::sender __Previous, std::integral __Size, class __Fn>
+  struct system_bulk_sender {
+    /// Marks this type as being a sender; not to spec.
+    using sender_concept = stdexec::sender_t;
+    /// Declares the completion signals sent by `this`.
+    using completion_signatures = stdexec::completion_signatures<
+      stdexec::set_value_t(),
+      stdexec::set_stopped_t(),
+      stdexec::set_error_t(std::exception_ptr) >;
```
ericniebler, line 323: This isn't correct. The `bulk` sender should send the values and errors of its predecessor.

include/exec/system_context.hpp
ericniebler: Also, we should be storing the decayed values in the tuple.

lucteo: We shouldn't allocate here; we should be able to tell size of all the possible `set_value` arguments when we build this object. This is on my list of TODOs, but didn't get to it yet.

On the decayed values, I didn't noticed that. Thank you for pointing it out.

include/exec/system_context.hpp
lucteo: I doubt there is a better way. In the common case, each call will be made on a different thread.

In general, to schedule work, we need 1 + N calls across the external interface (possible to a different library), where N is the number of threads we need.

include/exec/system_context.hpp
lucteo: Indeed, see Lee's TODO 2 lines below.

However, my understanding that this sender will always complete with `set_value()` with no arguments, regardless of the values sent by the previous senders; these values are just used to pass in to the given user function. Is my understanding wrong?

On the error side, my plan is to forward all errors given by the previous sender, plus `std::exception_ptr`, so that failure to schedule on the external scheduler can be reported as an error.

ericniebler: > However, my understanding that this sender will always complete with `set_value()` with no arguments, regardless of the values sent by the previous senders

Not so. `bulk` is specified to move the stored values into the value channel after the bulk work is done.

lucteo:
Dear race conditions and heisenbugs. I get it that you are angry with me, but now you really pissed me off. I've tuned already my profiler on you. Have a happy very-short life!

lucteo:
> Dear race conditions and heisenbugs. I get it that you are angry with me, but now you really pissed me off. I've tuned already my profiler on you. Have a happy very-short life!

Found ya!
image

lucteo:
@ericniebler : can you please give an /ok to test ?
(I took the latest bugs to be a personal insult ;) )

ericniebler (approved):

ericniebler:
@lucteo are you planning any more work on this PR?

PR #1214, opened 2024-01-19
task: mark class as nodiscard

williamspatrick:
If you have a function that returns a task and forget to `co_await` on
it nothing happens.  By adding a nodiscard on basic task, this ensures
that any missed `co_await` is caught as a compile error.  This gives a
quick alert to a trivial coding bug.


ericniebler:
This is a good idea, thanks

PR #1220, opened 2024-01-22
cause `spawn`-ed work to terminate with the active exception

### LINKED ISSUE #1213: async_scope set_error calls std::terminate and exception is lost
If you spawn an `exec::task` to the async_scope, and that task throws an exception you end up with a fairly non-useful log due to std::terminate:
```
terminate called without an active exception
```

I see in the code applicable code this BUGBUG:
```cpp
        // BUGBUG NOT TO SPEC spawn shouldn't accept senders that can fail.
        template <same_as<set_error_t> _Tag>
        [[noreturn]] friend void
          tag_invoke(_Tag, __t&&, const std::exception_ptr&) noexcept {
          std::terminate();
        }
```

I'm not sure what the intended solution would be here.  It sounds like this shouldn't have even compiled?

Would it be reasonable to change the code to something like this?
```cpp
        template <same_as<set_error_t> _Tag>
        [[noreturn]] friend void
          tag_invoke(_Tag, __t&&, const std::exception_ptr& e) noexcept {
          try { std::rethrow_exception(e); } catch (...) { std::terminate(); }
        }
```

ericniebler:
fixes #1213

### DIFF
--- a/include/exec/async_scope.hpp
+++ b/include/exec/async_scope.hpp
@@ -655,8 +655,8 @@ namespace exec {
 
         // BUGBUG NOT TO SPEC spawn shouldn't accept senders that can fail.
         template <same_as<set_error_t> _Tag>
-        [[noreturn]] friend void tag_invoke(_Tag, __t&&, const std::exception_ptr&) noexcept {
-          std::terminate();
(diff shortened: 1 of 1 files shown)


williamspatrick:
LGTM.

PR #1226, opened 2024-01-30
Update #includes in exec/ and stdexec/ based on clangd symbol uses

laramiel:
A few of these were build failures in my config.
This also attempts to maintiain/improve include order consistency: the system libraries were typically included last in the .hpp files

### DIFF
--- a/include/stdexec/execution.hpp
+++ b/include/stdexec/execution.hpp
@@ -44,7 +44,6 @@
 #include <tuple>
 #include <type_traits>
 #include <variant>
-#include <coroutine>
 #include <cstddef>
 #include <exception>
 #include <utility>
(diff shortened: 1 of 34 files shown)


include/stdexec/execution.hpp
```diff
@@ -15,33 +15,40 @@
...
+#include <condition_variable>
+#include <stdexcept>
+#include <memory>
+#include <mutex>
+#include <optional>
+#include <system_error>
+#include <tuple>
+#include <type_traits>
+#include <variant>
+#include <coroutine>
```
laramiel, line 47: Note: This was suggested, but perhaps it should be omitted here, or guarded with the proper #ifdef

ericniebler:
wow, thanks!

PR #1227, opened 2024-01-30
Create CUDA events with cudaEventDisableTiming

caugonnet:
CUDA events enable timing by default, which may introduce superfluous synchronizations when timing is not needed


ericniebler:
Thanks!

PR #1228, opened 2024-01-30
Additional #include cleanups

laramiel:
Additional #include cleanups from clangd.

Also, in execution.hpp convert one instance of std::move to std::forward.
Typically stdexec uses an old-style cast `(_Sender&&)sender`, however here std::move was used when it should be forward (https://clang.llvm.org/extra/clang-tidy/checks/bugprone/move-forwarding-reference.html). There is another instance of this clangd check around line 2900 `auto __notify_visitor(`... which clangd incorrectly tags.

I did not apply these to the rest of the tests subdirectory; there are quite a few `#include`s still missing there.

### DIFF
--- a/include/stdexec/execution.hpp
+++ b/include/stdexec/execution.hpp
@@ -4521,7 +4521,7 @@ namespace stdexec {
       auto operator()(_Sender&& __sndr) const {
         auto __domain = __get_early_domain(__sndr);
         return stdexec::transform_sender(
-          __domain, __make_sexpr<into_variant_t>(__(), std::move(__sndr)));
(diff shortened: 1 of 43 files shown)


include/stdexec/execution.hpp
```diff
@@ -4521,7 +4521,7 @@ namespace stdexec {
       auto operator()(_Sender&& __sndr) const {
         auto __domain = __get_early_domain(__sndr);
         return stdexec::transform_sender(
-          __domain, __make_sexpr<into_variant_t>(__(), std::move(__sndr)));
+          __domain, __make_sexpr<into_variant_t>(__(), std::forward<_Sender>(__sndr)));
```
ericniebler, line 4524: :+1:

ericniebler:
thanks again!

PR #1230, opened 2024-02-01
use stdexec::unstoppable_token in place of std::unstoppable_token

laramiel:
Also add pragma once / update include in __xorshift


ericniebler:
nice catch, thanks

PR #1232, opened 2024-02-02
add `exec::reschedule()` for rescheduling onto the current scheduler

ericniebler:


include/exec/reschedule.hpp
```diff
@@ -0,0 +1,89 @@
+/*
+ * Copyright (c) 2022 NVIDIA Corporation
```
ericniebler, line 2:
```suggestion
 * Copyright (c) 2024 NVIDIA Corporation
```

PR #1242, opened 2024-02-14
Restrict redundant STDEXEC_UNREACHABLE calls to NVHPC

vasama:


ericniebler:
A developer of nvc++ pointed out to me that the bugs which necessitated the `STDEXEC_UNREACHABLE()`s in the first place was fixed a few releases ago. In other words, we shouldn't need _any_ `STDEXEC_UNREACHABLE()` at all.

Could you try removing them all?

PR #1245, opened 2024-02-16
add limited early type checking of sender expression

ericniebler:


include/stdexec/__detail/__basic_sender.hpp
```diff
@@ -444,10 +461,10 @@ namespace stdexec {
     static constexpr auto connect = __detail::__connect;
     static constexpr auto start = __detail::__start;
     static constexpr auto complete = __detail::__complete;
-    static constexpr auto get_completion_signagures = __detail::__get_completion_signagures;
+    static constexpr auto get_completion_signagures = __detail::__get_completion_signatures;
```
villevoutilainen, line 464: Might want to fix this "signagures" to "signatures", too.

lewissbaker:
Instead of having to explictily opt-in to an `is_dependent` trait on the sender, would it be better to allow the `get_completion_signatures` cpo to have two different overloads?

- `get_completion_signatures(sender)` - only supported by non-dependent senders
- `get_completion_signatures(sender, env)` - supported by all senders. forwards to `tag_invoke(get_completion_signatures, sender, env)` if valid, otherwise to `get_completion_signatures(sender)` if valid.

This way a non-dependent sender could just define `tag_invoke(get_completion_signatures_t, sender&&)` and not need to also specify `is_dependent`.

Sender adaptors would need to define both, however, since if they want to be able to be non-dependent if input is non-dependent they also need to support the overload that doesn't take an env.
However, this can be done in a single function-template by defining it to have a variadic number of envs.

e.g. rough sketch
```
template<same_as<then_sender> Self, queryable... Env>
auto tag_invoke(get_completion_signatures_t, const Self&, const Env&...)
  -> compute_derived_signatures<completion_signatures_of_t<WrappedSender, Env...>>;
```

Also, the downside of the `is_dependent` query is that a lot of sender adaptors are going to need to conditionally define the `is_dependent` query.
e.g. then-sender is dependent sender if its input sender is a dependent sender

If `is_dependent` is a type alias that you test for presence or not it's going to be more difficult to conditionally define it - need to play shenanigans with base-class mix-ins or similar.
Better would be to make it a `static constexpr bool is_dependent` and if not present then defaults to `true`?

ericniebler:
There already is a way for a sender type to advertise its completions without an `Env`: with a nested `Sender::completion_signatures` type alias. We could additionally add support for `get_completion_signatures(sndr)`, but an adaptor would need to constrain it, obviously. And the constraint is exactly the boolean that `Sender::__is_non_dependent()` is returning. So IMO it's easier to just advertise the boolean instead of using it to constrain an extra overload of `get_completion_signatures`.

And you misread this change. The `is_dependent` trait is in the exposition-only `__sexpr_impl` specializations, and the only senders that need to set it are the ones who are dependent even when their children are not.

This PR actually gives no public way for third-party senders to opt-in to non-dependent-ness apart from defining the nested `::completion_signatures` alias. We strictly don't need to do more, but we might want to for the reasons you mention.

ericniebler:
@maikel do you have any ideas why this early type-checking would be causing the `transform_each` tests to fail to compile?

maikel:
> @maikel do you have any ideas why this early type-checking would be causing the `transform_each` tests to fail to compile?

This commit detected an error in the constraint of the subscribe method of `transform_each`. That's great!

include/stdexec/__detail/__basic_sender.hpp
ericniebler:
```suggestion
    static constexpr auto get_completion_signatures = __detail::__get_completion_signatures;
```

lewissbaker:
> There already is a way for a sender type to advertise its completions without an `Env`: with a nested `Sender::completion_signatures` type alias. We could additionally add support for `get_completion_signatures(sndr)`, but an adaptor would need to constrain it, obviously. And the constraint is exactly the boolean that `Sender::__is_non_dependent()` is returning. So IMO it's easier to just advertise the boolean instead of using it to constrain an extra overload of `get_completion_signatures`.

For a lot of adaptors, the constraint on whether or not the parent sender is non-dependent is going to be based on whether or not the child is non-dependent. i.e. if you can query the completion-signatures from the child in a non-dependent way then you should be able to query the completion-signatures from the parent in a non-dependent way, too. In a lot of cases, the parent is going to just be performing a transformation of the completion-signatures of the child operation(s) and that transformation is going to be the same regardless of whether you are querying with a particular concrete env, or querying without an env (non-dependent).

Being able to write a single definition of `get_completion_signatures()` that handles both dependent and non-dependent cases seems like a net win over having to define separate `get_completion_signatures()` and then conditionally  define a `::completion_signatures` alias based on whether the child operation has a `::completion_signatures` member.

Currently:
```c++
template<typename InnerSignatures>
using some_sender_adaptor_signatures_t = /* ... */;

template<typename Inner, typename = void>
struct some_sender_adaptor_base {};

template<typename Inner>
struct some_sender_adaptor_base<Inner, std::void_t<typename Inner::completion_signatures>> {
  using completion_signatures = some_sender_adaptor_signatures_t<typename Inner::completion_signatures>;
};

template<typename Inner>
struct some_sender_adaptor : some_sender_adaptor_base<Inner> {
  template<typename Env>
  auto get_completion_signatures(Env) -> some_sender_adaptor_signatures_t<
    completion_signatures_of_t<Inner, Env>>;
};
```
versus, say:
```
template<typename Inner>
struct some_sender_adaptor {
  template<typename... Env>
  requires (sizeof...(Env) <= 1)
  auto get_completion_signatures(Env...) -> some_sender_adaptor_signatures_t<
    completion_signatures_of_t<Inner, Env...>>;
};
```

> And you misread this change. The `is_dependent` trait is in the exposition-only `__sexpr_impl` specializations, and the only senders that need to set it are the ones who are dependent even when their children are not.

👍 

> This PR actually gives no public way for third-party senders to opt-in to non-dependent-ness apart from defining the nested `::completion_signatures` alias. We strictly don't need to do more, but we might want to for the reasons you mention.

I think that trying to conditionally define a nested `::completion_signatures` alias based on some  constraint is going to be a pain for users.
If it was a template alias then you might be able to get away with a requires-clause on the template, but not for a plain alias.

PR #1250, opened 2024-02-18
[DRAFT] CUDA Scan implementation

ZelboK:
Just an initial skeleton of a scan implementation for CUDA. For brevity I just used the reduce test spec to test my changes. Obviously it would need it's own spec.

### DIFF
--- a/include/nvexec/stream/scan.cuh
+++ b/include/nvexec/stream/scan.cuh
@@ -1,0 +1,172 @@
+/*
+ * Copyright (c) 20224 NVIDIA Corporation
+ *
+ * Licensed under the Apache License Version 2.0 with LLVM Exceptions
+ * (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ *   https://llvm.org/LICENSE.txt
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#pragma once
+
+#include "../../stdexec/execution.hpp"
+#include <type_traits>
+#include <ranges>
+
+#include <cuda/std/type_traits>
+
+#include <cub/device/device_scan.cuh>
+
+#include "algorithm_base.cuh"
+#include "common.cuh"
+#include "../detail/throw_on_cuda_error.cuh"
+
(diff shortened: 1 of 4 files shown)


include/nvexec/stream/scan.cuh
```diff
@@ -0,0 +1,172 @@
...
+
+          status = STDEXEC_DBG_ERR(cudaFreeAsync(d_temp_storage, stream));
+          self.op_state_.defer_temp_storage_destruction(d_out);
+          if (status == cudaSuccess) {
+            int* host_out = new int[10];
+            cudaMemcpy(host_out, d_out, 10 * sizeof(int), cudaMemcpyDeviceToHost);
+            for (size_t i = 0; i < 10; ++i) {
+              std::cout << host_out[i] << std::endl;
+            }
+            self.op_state_.propagate_completion_signal(stdexec::set_value, d_out);
```
ZelboK, line 114: If you change `d_out` to `*d_out` you can confirm that the data is scanning properly. But it won't compile I imagine because of the comcpletion signatures being wrong.

include/nvexec/stream/scan.cuh
```diff
@@ -0,0 +1,172 @@
...
+#include "../detail/throw_on_cuda_error.cuh"
+
+namespace nvexec {
+  namespace STDEXEC_STREAM_DETAIL_NS {
+    namespace scan_ {
+
+
+      template <class SenderId, class ReceiverId, class InitT, class Fun>
+      struct receiver_t
+        : public __algo_range_init_fun::receiver_t<
```
ZelboK, line 37: This reuses the `algorithm_base.cuh` . The `ExclusiveScan` API I used was the one that allows you to specify an initial value, so I could easily reuse this base. 

The difference between the reduce is that it returns a single value where as a scan is to return an array of data so it is very similar.

include/nvexec/stream/scan.cuh
```diff
@@ -0,0 +1,172 @@
...
+
+        template <class Range>
+        using _set_value_t = completion_signatures<set_value_t(
+          ::std::add_lvalue_reference_t<
+            typename __algo_range_init_fun::binary_invoke_result_t<Range, InitT, Fun>>)>;
+
+        // template <class Range>
+        // using _set_value_t = completion_signatures<set_value_t(
+        //   std::vector<typename __algo_range_init_fun::binary_invoke_result_t<Range, InitT, Fun>>)>;
+
```
ZelboK, line 137: Not sure how to get the completion signatures right. Hoping to get some guidance

ZelboK:
@gevtushenko Can you take a look?

Issue #1253, opened 2024-02-20
Can not destructure a sender with multiple value


SamGaaWaa:
[godbolt](https://godbolt.org/z/5sereTEcG)

```c++
exec::task<void> test(){
    auto [x, ptr] = co_await ex::just(2, std::make_unique<int>(2));
    std::cout << x << ' ' << *ptr << '\n';
}
```
```shell
AddressSanitizer:DEADLYSIGNAL
=================================================================
==1==ERROR: AddressSanitizer: SEGV on unknown address 0x000000000000 (pc 0x000000404d66 bp 0x7ffc00bd50d0 sp 0x7ffc00bd5050 T0)
==1==The signal is caused by a READ memory access.
==1==Hint: address points to the zero page.
    #0 0x404d66 in test /app/example.cpp:10
    #1 0x410a3f in std::__n4861::coroutine_handle<void>::resume() const /opt/compiler-explorer/gcc-12.2.0/include/c++/12.2.0/coroutine:135
    #2 0x424911 in void stdexec::__as_awaitable::tag_invoke<stdexec::__receivers::set_value_t, int, std::unique_ptr<int, std::default_delete<int> > >(stdexec::__receivers::set_value_t, stdexec::__as_awaitable::__receiver_base<std::tuple<int&&, std::unique_ptr<int, std::default_delete<int> >&&> >&&, int&&, std::unique_ptr<int, std::default_delete<int> >&&) /opt/compiler-explorer/libs/stdexec/trunk/include/stdexec/execution.hpp:1520
[...]

PR #1255, opened 2024-02-22
fix lifetime of multiple values in senders

zyctree:
it will fix https://github.com/NVIDIA/stdexec/issues/1253


ericniebler:
thanks!

PR #1266, opened 2024-03-01
Implement receiver_ref for let_value

maikel:
This PR adds a type-erased receiver to the `let_` algorithms. Its purpose is to transparently compute whether the delayed call to `connect` will throw or not. 

The implementation strategy is as follows:

1. First compute whether the second call to connect between every resulting sender and the type erased `receiver_ref <ResultSenderSigs, Env, Scheduler>` will throw.

2. In the final operation state check whether connecting with the resulting Receiver improves or retains noexceptness. If yes, then connect with that one and otherwise we use the indirection via `receiver_ref` to keep `noexcept` for the call to `connect`.

For well-behaved senders and receivers it is expected to practically never need to really connect to the receiver ref. 
This strategy can be used in every algorithm that manages a variant operation states with multiple stages.

This is related to https://github.com/cplusplus/sender-receiver/issues/88

### DIFF
--- a/include/stdexec/execution.hpp
+++ b/include/stdexec/execution.hpp
@@ -3467,7 +3467,7 @@ namespace stdexec {
   //////////////////////////////////////////////////////////////////////////////
   // [exec.let]
 
-  namespace __any {
+  namespace __any_ {
     template <class _Sig>
     struct __receiver_signature_for;
 
@@ -3487,7 +3487,7 @@ namespace stdexec {
         _Tag{}(static_cast<_Receiver&&>(*__rcvr), static_cast<_Args&&>(__args)...);
       };
     }
-  } // namespace __any
+  } // namespace __any_
 
   namespace __let {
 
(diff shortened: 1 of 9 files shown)


ericniebler:
There is a `__receiver_ref` in `any_sender_of.hpp`. Can we reuse it?

maikel:
> There is a `__receiver_ref` in `any_sender_of.hpp`. Can we reuse it?

Yes, I intend to make some changes in `exec/any_sender_of.hpp`, too and refer to the bits in `execution.hpp`.

include/stdexec/execution.hpp
```diff
@@ -3466,7 +3511,96 @@ namespace stdexec {
...
+      friend void tag_invoke(_Tag, _Self&& __self, _As&&... __as) noexcept {
+        (*__self.__vtable_)(__self.__receiver_, _Tag(), static_cast<_As&&>(__as)...);
+      }
+
+     private:
+      const __receiver_vtable_for<_Sigs, _Env>* __vtable_;
+      void* __receiver_;
+      STDEXEC_ATTRIBUTE((no_unique_address))
+      _SchedulerOrNoneSuch __scheduler_;
+    };
```
ericniebler, line 3602: I'm a bit surprised to see all this stuff in the `__let` namespace instead of `__any_`. What's the reason?

ericniebler (commented):
Looks very good. Just a few nits. And I think it may be better to factor the `receiver_ref` stuff out into a `__detail/` header, which is pulled in from both `execution.hpp` and `any_sender_of.hpp`.

include/stdexec/execution.hpp
maikel: This receiver ref has a slightly different flavor.

1. completion functions are type erased 
2. call to get_env is type erased
3. we pass an additional scheduler

I think the third point is not well implemented and could be hidden behind the type erased call to get env. I will try to address that.

The receiver_ref in `any_sender` type-erases a given set of queries in addition to the completion functions, i.e. the environment itself is type-erased, too. I think those two `receiver_refs` are different. :thinking: 

...
Maybe something like `receiver_ref<Sigs, any_env<Queries...>>` might be possible (even a template specialization). What do you think?

ericniebler:
the nvc++ problem is fixed, but now it seems that something has changed to cause gcc-11+ASAN to take forever linking test.stdexec. the problem is happening on `main` as well.

PR #1275, opened 2024-03-07
Fixed swappable concept

koloshmet:
### Tiny fix of swappable concept

Issue: ```stdexec::swappable<int>``` evaluates to false (it must not) [godbolt](https://godbolt.org/z/oKPKcxTfb)

As a result exec::now (from timed scheduler) failures with any reasonable implementation

### DIFF
--- a/test/exec/test_io_uring_context.cpp
+++ b/test/exec/test_io_uring_context.cpp
@@ -118,6 +118,16 @@ namespace {
   }
 
   TEST_CASE(
+    "io_uring_context Call now(io_uring) is running clock",
+    "[types][io_uring][schedulers]") {
+    io_uring_context context;
+    io_uring_scheduler scheduler = context.get_scheduler();
+    auto start = now(scheduler);
+    std::this_thread::sleep_for(10ms);
+    CHECK(start + 10ms <= now(scheduler));
+  }
+
+  TEST_CASE(
     "io_uring_context Call io_uring::run_until_empty with sync_wait",
     "[types][io_uring][schedulers]") {
     io_uring_context context;
--- a/include/stdexec/concepts.hpp
+++ b/include/stdexec/concepts.hpp
@@ -159,7 +159,9 @@ namespace stdexec {
 
   template <class _Ty>
   concept swappable = //
-    swappable_with<_Ty, _Ty>;
+    requires(_Ty& a, _Ty& b) {
+      swap(a, b);
+    };
 
   template <class _Ty>
   concept movable =               //


test/exec/test_io_uring_context.cpp
```diff
@@ -117,6 +117,16 @@ namespace {
...
+  TEST_CASE(
+    "io_uring_context Call now(io_uring) is running clock",
+    "[types][io_uring][schedulers]") {
+    io_uring_context context;
+    io_uring_scheduler scheduler = context.get_scheduler();
+    auto start = now(scheduler);
+    std::this_thread::sleep_for(10ms);
+    CHECK(start + 10ms <= now(scheduler));
+  }
+
```
ericniebler, line 129: Is this change somehow related to the change to `swappable`?

koloshmet: After the fix, ```now(io_uring_scheduler)``` compiles.
Seems like the only usage of movable and therefore swappable concept is the static assertion in ```now_t``` CPO of timed scheduler and the only implementation of timed scheduler is io_uring

maikel: Let me change the fact that there is only one platform-dependent timed scheduler. I will add a `timed_thread_scheduler` next and fix some outstanding issues with the current one.

koloshmet: It would be great, but tests of my own timed scheduler is currently blocked by the fact that I can't implement ```now```. Are you going to design some completely different api for timers?

maikel: IMHO, the general design for timers is good. I don't quite understand yet why `swappable<int>` is failing and I think it shouldn't. I'm not totally convinced yet that your changes are solving the issue. Shouldn't we fix `swappable_with` instead?

koloshmet: ```std::swap``` accepts 2 lvalue references, but arguments of requires clause of ```swappable_with``` are universal references, so the only way to satisfy it is to pass 2 lvalue references as concept arguments. This is ok, according to [standard](https://timsong-cpp.github.io/cppwp/n4861/concept.swappable). But requires clause of swappable accepts 2 lvalue references of initial types, therefore according to reference stickiness, argument type of ```swappable``` can be any of ```T```, ```T&``` and ```T&&```
I may not be very strict, but I hope it explains something 😅
P.S. My changes is a copy of standard definition, so, I hope it works

maikel: Yup, I have the same error for #1277

ericniebler: Thanks for clarifying. I'll merge this today, after tests pass.

PR #1277, opened 2024-03-09
Implement timed_thread_scheduler

### LINKED ISSUE #685: make `run_loop` implementation lock-free

### LINKED ISSUE #847: `schedule_after` should start the work `duration` time after the work is *started*
Currently, it starts the work `duration` time after `schedule_after` is called.

maikel:
Depends on #1275 

Fixes #847

Note, scheduling on this scheduler is wait-free*. the intrusive mpsc queue can be used to address issue #685 

* it is wait-free for producers on the fast-path only

### DIFF
--- a/include/exec/__detail/intrusive_heap.hpp
+++ b/include/exec/__detail/intrusive_heap.hpp
@@ -1,0 +1,221 @@
+/*
+ * Copyright (c) 2024 Maikel Nadolski
+ * Copyright (c) 2024 NVIDIA Corporation
+ *
+ * Licensed under the Apache License Version 2.0 with LLVM Exceptions
+ * (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ *   https://llvm.org/LICENSE.txt
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#pragma once
+
+#include "../../stdexec/__detail/__config.hpp"
+
+#include <cstddef>
+#include <bit>
+
+namespace exec {
+#if defined(__cpp_lib_int_pow2) && __cpp_lib_int_pow2 >= 202002L
+  namespace detail {
+    using std::bit_ceil;
+  } // namespace detail
+#else
+#  if defined(__has_builtin) && __has_builtin(__builtin_clzll)
+  namespace detail {
+    inline std::size_t bit_ceil(std::size_t n) noexcept {
+      const int leading_zeros = __builtin_clzll(n);
+      return std::size_t{1} << (sizeof(std::size_t) * 8 - leading_zeros + 1);
+    }
+  } // namespace detail
+#  else
+  namespace detail {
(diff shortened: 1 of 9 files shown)


include/exec/__detail/intrusive_heap.hpp
```diff
@@ -0,0 +1,221 @@
...
+      m |= m >> 16;
+      m |= m >> 32;
+      return m + 1;
+    }
+  } // namespace detail
+#  endif
+#endif
+
+  template <auto Key, auto Prev, auto Left, auto Right>
+  class intrusive_heap;
```
koloshmet, line 54: I apologize for the invasion, but I worked on a very similar thing and came to the thought, that erasures from heap are too expensive and therefore cancelation is also. Isn't it better to implement intrusive ordered set based on any balanced binary tree? To my shame I never implemented this idea, but maybe you can find it interesting.

maikel: I think erasure here is O(log N) (I set the time point value to 0 and call bottom-up-heapify). Thats pretty good.

For a real production version of this I have seen that @lewissbaker or @kirkshoop mentioned some hash based priority queue. But I don't know how that works.

koloshmet: Yes, you're right, I was too caught up with the thought about multiple erasures that time, sorry

maikel: Note that this is not a generic ordered set. Deleting an element modifies the key value which is unusual. More over it assumes that a zero initialized key value yields a minimum. More over bottom up heapify will pass the node up even if the parent has a zero timeout, too. That way it is ensured that after calling bottom up the node to be erased is at root and can be popped back.

This is a data structure that works for timeouts but is probably not good for a generic setting

include/exec/timed_thread_scheduler.hpp
```diff
@@ -0,0 +1,343 @@
...
+      timed_thread_schedule_operation_base(
+        std::chrono::steady_clock::time_point time_point,
+        void (*set_stopped)(timed_thread_operation_base*) noexcept,
+        void (*set_value)(timed_thread_operation_base*) noexcept) noexcept
+        : timed_thread_operation_base{set_value, command_type::schedule}
+        , time_point_{time_point}
+        , set_stopped_{set_stopped} {
+      }
+
+      std::chrono::steady_clock::time_point time_point_;
```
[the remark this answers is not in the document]
maikel: Is that so.. hm yes I see how thats currently not guaranteed. I will address that

maikel:
I think the current failure in TSAN is due to https://github.com/google/sanitizers/issues/1259 and would be solved if we upgrade gcc to version 12. That works at least locally for me.

I still have to investigate why the tests on windows are failing.

include/exec/timed_thread_scheduler.hpp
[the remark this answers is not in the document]
maikel: Please have another look :pray: I've added the `when_type` But I only convert the time point to the when type when the operation is started and not at construction. Is this the correct timing to increase the counter?

Should I somehow address the issue that the static counter wraps around or should I assume that `std::size_t` is practically large enough to never wrap around?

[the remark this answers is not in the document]
maikel: I think I found a better place for the counter. The worker thread of the context can do it without any synchronization. The order of the mpsc queue will define the fifo order. The when type will be created right before inserting it into the internal heap data structure

ericniebler:
c

PR #1282, opened 2024-03-12
use cccl instead of the old libcudacxx repo

ericniebler:


robertmaynard (approved):

cmake/versions.json
```diff
@@ -1,8 +1,8 @@
 {
   "packages" : {
-    "libcudacxx" : {
-      "version" : "1.9.0",
-      "git_url" : "https://github.com/NVIDIA/libcudacxx.git",
+    "cccl" : {
+      "version" : "v2.2.0",
+      "git_url" : "https://github.com/NVIDIA/cccl.git",
```
robertmaynard, line 5:
```suggestion
    "cccl" : {
      "version" : "v2.2.0",
      "git_url" : "https://github.com/NVIDIA/cccl.git",
```
This needs to look like:
```
    "CCCL" : {
      "version" : "2.2.0",
      "git_url" : "https://github.com/NVIDIA/cccl.git",
      "git_tag" : "v${version}"
    }
```
As the version string should be numeric only, and the name is case sensitive ( needs to match the `find_package` name ).

robertmaynard (changes requested):

cmake/versions.json
robertmaynard: It currently works due to your entry being ignored, and the default `CCCL` entry being used.

cmake/versions.json
```diff
@@ -1,8 +1,9 @@
 {
   "packages" : {
-    "libcudacxx" : {
-      "version" : "1.9.0",
-      "git_url" : "https://github.com/NVIDIA/libcudacxx.git",
+    "CCCL" : {
+      "version" : "2.2.0",
+      "git_url" : "https://github.com/NVIDIA/cccl.git",
+      "git_tag" : "v${version}"
       "git_tag" : "${version}"
```
ericniebler, line 7:
```suggestion
    "CCCL" : {
      "version" : "2.2.0",
      "git_url" : "https://github.com/NVIDIA/cccl.git",
      "git_tag" : "v${version}"
```

PR #1290, opened 2024-03-16
System context replaceability

lucteo:
Add ability to replace the default implementation of system context.

system_context can be delivered in two ways:
* with a static library implementing the default behaviour (this is the default)
* header-only, for convenience (user needs to specify with a define)

The static library comes with the getter of the system context specified with weak linking. This allows users to overwrite the default implementation. Adding a test that shows that this behaviour can be changed at user level. The test needs a new executable, and links the default system_context library.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -332,6 +332,8 @@ target_include_directories(system_context PRIVATE
                            $<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}>
                            )
 target_compile_features(system_context PRIVATE cxx_std_20)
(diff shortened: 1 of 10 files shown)


lucteo:
I have no clue why https://github.com/NVIDIA/stdexec/actions/runs/8366919773/job/22908184313?pr=1290 fails system_context tests with `pure virtual method called`. The implementation of system_context doesn't use virtual functions.

@ericniebler : do you have any ideas?

ericniebler:
I suspect bad codegen. I have seen this happen on other PR's as well.

PR #1294, opened 2024-03-20
Update `ci.{c,g}pu.yml` and devcontainers to clang18, CUDA 12.4, and nvhpc24.3

trxcllnt:
* update `ci.cpu.yml` to cuda12.4
* update `ci.gpu.yml` to clang18-cuda12.0, clang18-cuda12.4, and nvhpc24.3
* devcontainers to clang18-cuda12.0, clang18-cuda12.4, and nvhpc24.3


ericniebler:
Thank you! Do you know if this will fix the gcc-11 Release ASAN build? It's been suffering from a weird linker error for weeks.

trxcllnt:
> Do you know if this will fix the gcc-11 Release ASAN build?

I don't know, but we'll see here shortly. We can also try other versions of GCC if we need.

trxcllnt:
@ericniebler I think it's just complaining because I put `[skip ci]` in the last commit. This PR should be good to merge.

PR #1308, opened 2024-04-23
Remove cruft from retry.hpp

camio:


ericniebler:
I often insert trailing comments like that to keep clang-format from doing something weird. Have you tried running clang-format over that file after the change?

camio:
Ah, I didn't realize this was there to work around clang-format's output. Sure enough, if I remove the comment clang-format makes this less readable. Closing this.

PR #1310, opened 2024-04-24
Fix `split`/`ensure_started` with non-stdexec senders

msimberg:
It looks like the computation of the completion signatures for `split`/`ensure_started` assumed that the sender and env types were ADL-hidden when they in fact aren't. This happens to compile with stdexec-senders (that have a `__t` typedef), but doesn't compile with user-provided senders. I've added tests with "custom senders", and removed the un-hiding of types. Or should it be the converse: the sender and env types should actually be hidden? It's unclear to me if the types will show up in a `tag_invoke` call somewhere.

### DIFF
--- a/include/stdexec/execution.hpp
+++ b/include/stdexec/execution.hpp
@@ -3279,13 +3279,13 @@ namespace stdexec {
       }
     };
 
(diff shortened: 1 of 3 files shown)


ericniebler:
thank you! well-spotted. this is the correct fix.

PR #1313, opened 2024-04-27
Disable tests and installing Catch2 if in a subproject

planetmarshall:
Catch2 is still installed if stdexec is used as a subproject. This can cause issues if Catch2 is already installed by the parent project. Also disable the building of tests by default if using as a subproject.


ericniebler:
Thanks! I'm grateful for any help I can get, but especially CMake help.

planetmarshall:
> Thanks! I'm grateful for any help I can get, but especially CMake help.

No problem. I hope to make more substantial contributions in the future but still getting to grips with the sender/receiver pattern.

PR #1332, opened 2024-05-17
Remove `__destruct_schedule_operation` from system_context ABI

lucteo:
The ABI for system_context contained the un-needed `__destruct_schedule_operation` operation (both for `schedule` and for `bulk_schedule`). This is not needed, as we can destroy the operation state after the completion signal was sent.


ericniebler:
@lucteo could you look into the gcc-11 release mode failure in the `"simple chain task on system context"` test?

ericniebler (approved):

Issue #1276, opened 2024-03-08
Very deep template instantiation stacks


AndreyG:
Just ["hello world"](https://github.com/NVIDIA/stdexec/blob/main/examples/hello_world.cpp)  example requires template depth limit be 229 for Clang 17.0.1 and 157 for GCC 13.2: https://gcc.godbolt.org/z/K45fxG75T. I think this is too much and cannot work in real world production code.
Do you have ideas what could be improved in this sense and why there are so deep template instantiation stacks?
Maybe something could be optimized in `__meta.hpp`?

ericniebler:
i haven't optimized for instantiation depth. yeah, there is probably some fat to cut in `__meta.hpp`.

ericniebler:
things are modestly better now. see your original link: https://gcc.godbolt.org/z/K45fxG75T. more needs to be done.

AndreyG:
Yes, it's a little bit better now, depth limit for Clang is 205 and for GCC is 141 (https://gcc.godbolt.org/z/9bG7PTnYs).

PR #1333, opened 2024-05-18
bring down the template instantiation depth a tad

ericniebler:
refs #1276


ericniebler:
with this change `hello_world.cpp` goes from a template instantiation depth of 230 to 218 (clang-17).

PR #1337, opened 2024-05-20
Make completion signatures depend on the Receiver type

maikel:
This draft PR tracks my attempt to make `get_completion_signatures` take the receiver type instead of the env. If that works out it would solve some pending issues wrt to detecting whether connecting child senders throws or not.

### DIFF
--- a/include/stdexec/__detail/__ensure_started.hpp
+++ b/include/stdexec/__detail/__ensure_started.hpp
@@ -39,8 +39,11 @@ namespace stdexec {
     struct __ensure_started_t { };
 
     struct ensure_started_t {
+      template <class _CvrefSender, class _Env>
+      using __receiver_t = __t<__meval<__receiver, __cvref_id<_CvrefSender>, __id<_Env>>>;
+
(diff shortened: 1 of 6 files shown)


ericniebler:
Have you considered doing away with get_completion_signatures and just having connect, with the completion signatures as a nested typedef on the operation state?

maikel:
> Have you considered doing away with get_completion_signatures and just having connect, with the completion signatures as a nested typedef on the operation state?

Hm Im not sure whether computing completion sigs can require a complete definition of the operation state. I think it must require a complete definition of the receiver, tho.

ericniebler:
> Have you considered doing away with get_completion_signatures and just having connect, with the completion signatures as a nested typedef on the operation state?

i have made this change in the ericniebler/ustdex and it works well. this is the direction i would like to pursue for stdexec.

ericniebler:
@maikel do you have plans for this PR or should i close it?

maikel:
> @maikel do you have plans for this PR or should i close it?

Please wait until 13th of January :pray:

PR #1338, opened 2024-05-21
start_now() allows stoppable work to be started with no allocation

kirkshoop:
I think that there is an algorithm that works for the embedded use case that is also useful in many other cases. I do not think that this algorithm needs to allocate.

Here is an implementation of start_now() 

After we remove start_detached() and ensure_started() and add counting_scope - I think that this facility will be another important way to start senders.

`start_now()` returns an immovable object that contains the operation and that connects and starts the operation in its constructor.

The returned object has `request_stop()` and `get_token()` because it has an `inplace_stop_source` and it has provided that `inplace_stop_token` to the given sender through the receiver environment.  

The returned object has `join() -> sender` so that the destruction of the returned object can be ordered after the contained operation completes.

### DIFF
--- a/include/exec/async_scope.hpp
+++ b/include/exec/async_scope.hpp
@@ -31,12 +31,9 @@ namespace exec {
     struct __impl;
     struct async_scope;
 
-    template<class _A> 
-    concept __async_scope = 
-      !std::is_copy_constructible_v<_A> && !std::is_move_constructible_v<_A> && 
-      !std::is_copy_assignable_v<_A> && !std::is_move_assignable_v<_A> && 
-      requires (_A& __a) {
-      {__a.nest(stdexec::just())} -> sender_of<stdexec::set_value_t()>;
+    template <class _A>
+    concept __async_scope = requires(_A& __a) {
+      { __a.nest(stdexec::just()) } -> sender_of<stdexec::set_value_t()>;
     };
 
     struct __task : __immovable {
@@ -811,4 +808,8 @@ namespace exec {
   } // namespace __scope
 
   using __scope::async_scope;
+
+  template <class _AsyncScope, class _Sender>
+  using nest_result_t =
+    decltype(stdexec::__declval<_AsyncScope&>().nest(stdexec::__declval<_Sender&&>()));
 } // namespace exec
--- a/include/exec/async_scope.hpp
+++ b/include/exec/async_scope.hpp
@@ -46,10 +46,6 @@ namespace exec {
     using __env_t = make_env_t<_BaseEnv, with_t<get_stop_token_t, inplace_stop_token>>;
 
     struct __impl {
-      #if STDEXEC_NVHPC()
-      // part of a workaround for NVHPC codegen bug:
-      unsigned int const __cookie_ = 0xDEADBEEF;
-      #endif
       inplace_stop_source __stop_source_{};
       mutable std::mutex __lock_{};
       mutable std::ptrdiff_t __active_ = 0;
@@ -225,10 +221,6 @@ namespace exec {
 
         void start() & noexcept {
           STDEXEC_ASSERT(this->__scope_);
-          #if STDEXEC_NVHPC()
-          // work around for NVHPC codegen bug:
-          void(this->__scope_->__cookie_ == 0xDEADBEEF ? void() : std::terminate());
-          #endif
           std::unique_lock __guard{this->__scope_->__lock_};
           auto& __active = this->__scope_->__active_;
           ++__active;
--- a/include/exec/async_scope.hpp
+++ b/include/exec/async_scope.hpp
@@ -46,6 +46,10 @@ namespace exec {
     using __env_t = make_env_t<_BaseEnv, with_t<get_stop_token_t, inplace_stop_token>>;
 
     struct __impl {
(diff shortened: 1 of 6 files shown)


include/exec/start_now.hpp
```diff
@@ -0,0 +1,185 @@
...
+        if (__joiner) {__joiner->join();}
+      }
+
+      // Forward all receiever queries.
+      auto get_env() const noexcept -> _Env {
+        return __stgref_().__env_;
+      }
+    };
+
+    static inline const __joiner __empty_joiner_{};
```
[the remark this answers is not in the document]
kirkshoop: constexpr results in: `error: constexpr variable cannot have non-literal type 'const exec::__start_now_::__joiner'` 

If I get rid of the virtual base this will be possible..

include/exec/async_scope.hpp
```diff
@@ -31,6 +31,14 @@ namespace exec {
     struct __impl;
     struct async_scope;
 
+    template<class _A> 
+    concept __async_scope = 
+      !std::is_copy_constructible_v<_A> && !std::is_move_constructible_v<_A> && 
+      !std::is_copy_assignable_v<_A> && !std::is_move_assignable_v<_A> &&
```
ericniebler, line 37: if a type has a `.nest()` member that returns a `sender_of<set_value_t()>`, do you really want the algorithms to reject the type just because it's movable? how is immovability an essential property of async scopes?

include/exec/start_now.hpp
```diff
@@ -75,107 +75,143 @@ namespace exec {
...
-        , __op_state_(connect(static_cast<_Sender&&>(__sndr), __receiver_t{__ref(*this)})) {
-        start(__op_state_);
+      __storage(_Env __env, _AsyncScope& __scope, stdexec::__t<_SenderId>... __sndr)
+        : __joiner_(&__empty_joiner_)
+        , __pending_(0)
+        , __source_()
+        , __env_(stdexec::__env::__join(stdexec::__env::__with{__source_.get_token(), stdexec::get_stop_token}, std::move(__env)))
+        , __op_state_(stdexec::__conv{__construct(__scope.nest(__sndr))}...) {
+        __pending_ = sizeof...(_SenderId);
+        stdexec::__apply([](auto&... __op_state) noexcept { bool arr[]{(stdexec::start(__op_state), true)...}; (void)arr; }, __op_state_);
```
ericniebler, line 160: you can fold over the comma operator like this:

```suggestion
        stdexec::__apply([](auto&... __op_state) noexcept { (stdexec::start(__op_state),...); }, __op_state_);
```

test/CMakeLists.txt
```diff
@@ -77,6 +77,7 @@ set(stdexec_test_sources
     exec/async_scope/test_dtor.cpp
     exec/async_scope/test_spawn.cpp
     exec/async_scope/test_spawn_future.cpp
+    exec/async_scope/test_start_now.cpp
```
ericniebler, line 80: looks like you forgot to commit this file

include/exec/start_now.hpp
```diff
@@ -0,0 +1,221 @@
...
+
+      mutable std::atomic<const __joiner*> __joiner_;
+      mutable std::atomic<int> __pending_;
+
+      STDEXEC_ATTRIBUTE((no_unique_address))
+      stdexec::inplace_stop_source __source_;
+      STDEXEC_ATTRIBUTE((no_unique_address))
+      __env_t __env_;
+      STDEXEC_ATTRIBUTE((no_unique_address))
+      stdexec::__decayed_tuple<stdexec::connect_result_t<__nested_t<stdexec::__t<_SenderId>>, __receiver_t>...> __op_state_;
```
ericniebler, line 146: if you change this to `stdexec::__tup::__tuple_for` here then ....

include/exec/start_now.hpp
```diff
@@ -0,0 +1,221 @@
...
+      auto __construct(_Sender&& __sndr) noexcept {
+        return [&, this](){return stdexec::connect(static_cast<_Sender&&>(__sndr), __receiver_t{__ref(*this)});};
+      }
+
+      __storage(_Env __env, _AsyncScope& __scope, stdexec::__t<_SenderId>... __sndr)
+        : __joiner_(&__empty_joiner_)
+        , __pending_(0)
+        , __source_()
+        , __env_(stdexec::__env::__join(stdexec::__env::__with{__source_.get_token(), stdexec::get_stop_token}, std::move(__env)))
+        , __op_state_(stdexec::__conv{__construct(__scope.nest(__sndr))}...) {
```
ericniebler, line 158: ... here you don't have to use `__conv`; you can just direct-initialize the tuple with a pack of calls to `stdexec::connect`. that works because `stdexec::__tup::__tuple` is an aggregate.

when you make this change, you will need to replace calls like `__apply(fn, static_cast<Tup&&>(tup))` to `tup.apply(fn, static_cast<Tup&&>(tup))`. (`__tuple::apply` is a static member; once stdexec can assume support for deducing-this, we can replace it everywhere with `tup.apply(fn)`.)

ericniebler:
this seems to have uncovered a possible codegen issue in nvc++. i'm working on reducing the fault in the hopes to find a workaround.

PR #1342, opened 2024-05-31
empty tuple members should not take up space

ericniebler:
also, stop using `std::integer_sequence` because of a nasty nvc++ bug.

### DIFF
--- a/include/stdexec/__detail/__concepts.hpp
+++ b/include/stdexec/__detail/__concepts.hpp
@@ -140,14 +140,14 @@ namespace stdexec {
 
   // Avoid using libstdc++'s object concepts because they instantiate a
   // lot of templates.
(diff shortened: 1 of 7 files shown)


include/stdexec/__detail/__concepts.hpp
```diff
@@ -160,17 +160,14 @@ namespace stdexec {
...
-#if STDEXEC_HAS_BUILTIN(__is_constructible)
-  template <class _Ty, class... _As>
-  concept constructible_from = //
-    destructible<_Ty> &&       //
-    __is_constructible(_Ty, _As...);
-#else
+#if !defined(STDEXEC_IS_CONSTRUCTIBLE)
+#error here
+#endif
+
```
ericniebler, line 166:
```suggestion
```

PR #1347, opened 2024-06-11
Refactor task_context construction

vasama:
* Delay construction of the task context until await.
* Add `__parent_promise_t` to disambiguate the context constructor.

This makes it easier to define task contexts which require querying the environment.

### DIFF
--- a/include/exec/task.hpp
+++ b/include/exec/task.hpp
@@ -88,6 +88,8 @@ namespace exec {
       __sticky
     };
 
+    struct __parent_promise_t {};
+
     template <__scheduler_affinity _SchedulerAffinity = __scheduler_affinity::__sticky>
     class __default_task_context_impl {
       template <class _ParentPromise>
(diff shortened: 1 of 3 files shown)


include/exec/task.hpp
```diff
@@ -406,10 +389,10 @@ namespace exec {
         using __context_t = typename _Context::template promise_context_t<__promise>;
 
         auto get_env() const noexcept -> const __context_t& {
-          return __context_;
+          return *__context_;
         }
 
-        __context_t __context_;
+        std::optional<__context_t> __context_;
```
ericniebler, line 395: pls replace this use of `std::optional` with `__optional` from `__detail/__optional.hpp`. it is easier on the compiler.

ericniebler (approved):
small nit, otherwise looks good.

vasama:
This CI failure looks very weird. I could not reproduce it locally, even after running the tests in a loop hundreds of times. Would you mind retrying it to see if maybe it was some fluke on the CI machine?

ericniebler:
i've been equally mystified by these msvc errors. i'm following up internally to see if there's something "special" about the msvc containers.

PR #1348, opened 2024-06-12
Update Conan support to Conan 2.0

vasama:
* Rename the package to `p2300` according to [Conan recommendations](https://docs.conan.io/1/migrating_to_2.0/general.html#lowercase-references).
* Change the package versioning to use git revision by default.
  Can be overridden using `conan create ./stdexec --version x.y.z`.
* Execute unit tests during package creation, unless opted out.
  Opt out using `conan create ./stdexec -c tools.build:skip_test=True`.
* Specify the CMake file name as `P2300` (`find_package(P2300)`).
* Specify the CMake target name as `P2300::P2300`.

The existing Conan support is fairly antiquated, but it exists so someone must be (or must have been) using it.
Thus I'm tagging anyone who touched the existing Conan support: @lucteo @paulbendixen @trxcllnt @ericniebler

### DIFF
--- a/conanfile.py
+++ b/conanfile.py
@@ -1,35 +1,51 @@
-from conans import ConanFile, CMake, tools
-import requests
-import re
-from os import path
+from conan import ConanFile
+from conan.tools.build.cppstd import check_min_cppstd
+from conan.tools.cmake import CMake, cmake_layout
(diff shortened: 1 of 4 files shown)


ericniebler:
Thanks!

PR #1352, opened 2024-06-16
Fix system_context linking with Conan package

vasama:
* The Conan package now includes the system context.
* The Conan test package now uses the system context.

The Conan package is created using CMake install. This requires the `system_context` library install to be configured. I added it where the `stdexec` library install is configured, but I'm not sure if that's where it should be.

When testing on WSL Ubuntu 22.04 with Clang 18, my `test_package` executable (see [here](https://github.com/vasama/stdexec/blob/450a55358a9aedf2e43edff982e73b529698780e/test_package/test.cpp)) would segfault when calling `__get_exec_system_context_impl`. Removing the `STDEXEC_ATTRIBUTE((weak))` from the declaration [here](https://github.com/NVIDIA/stdexec/commit/450a55358a9aedf2e43edff982e73b529698780e#diff-125f4090fb987bfec99abcee661b93ad7e229f8a3108bab6954f65f268079ba2) solved the issue. I'm not sure why.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -409,7 +409,7 @@ endif()
 
 include(CPack)
 
-install(TARGETS stdexec
+install(TARGETS stdexec system_context
         DESTINATION ${CMAKE_INSTALL_LIBDIR}
         EXPORT stdexec-exports)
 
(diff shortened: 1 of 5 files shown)


ericniebler (changes requested):
the `weak` attribute is an important feature of the system context. it makes it so that the context can be replaced at link time like the system allocator. we'll need to find a different fix.

ericniebler (approved):

PR #1369, opened 2024-07-10
fix Static Initialization Order Fiasco causing flakiness in `static_thread_pool` shutdown

ericniebler:
this intentionally leaks one memory block at shutdown. clang's and gcc's leak sanitizers do not flag it.


ericniebler:
@maikel can you comment on this PR? is it problematic for the node-to-thread-index vector to be global instead of thread-local? can the numa policies also be global instead of thread-local?

PR #1372, opened 2024-07-12
Add missing includes

laramiel:
These files can use an #include "__senders.hpp"

Otherwise __well_formed_sender declaration is missing.


ericniebler (approved):

ericniebler:
Thanks!

PR #1376, opened 2024-07-17
change the system context to be replaceable at runtime

ericniebler:
also, fix static initialization order fiasco issues wrt the system context at shutdown.

this design for the system context has a few nice properties:

* lazily constructed at first use
* accessible by default-constructing a `system_context` object, which grabs a ref-count to the current global system context in its ctor and releases it in its dtor.
* is usable at _all_ points during the execution of the program, even before/after `main`, with no Static Initialization Order Fiasco.
* is replaceable at runtime as many times as you like.
* replacing the global system context does not affect any existing `system_context` objects.
* shuts down cleanly with no leaks

this commit changes the ABI of the system context interfaces, so it bumps the version number. it also reserves space in the `system_context_interface` and `system_scheduler_interface` for additional "virtual" functions.

adds two helper types: `system_context_base` and `system_scheduler_base`, which automate the population of the `system_context_interface` and `system_scheduler_interface` "vtables", respectively.

it also adds `static_system_context_instance` to make it simple to safely put a custom system context into static storage.

### DIFF
--- a/include/exec/__detail/__system_context_default_impl.hpp
+++ b/include/exec/__detail/__system_context_default_impl.hpp
@@ -15,31 +15,41 @@
  */
 #pragma once
 
-#include "__system_context_if.h"
 #include "stdexec/execution.hpp"
 #include "exec/static_thread_pool.hpp"
 
-namespace exec::__system_context_default_impl {
-  using namespace stdexec::tags;
+STDEXEC_PRAGMA_PUSH()
+// "warning: empty struct has size 0 in C, size 1 in C++"
+// The intended use of these two structs is as a base class for other structs.
+// When used like that, the size occupied by these structs is 0 because of the
+// empty base class optimization.
+STDEXEC_PRAGMA_IGNORE_GNU("-Wextern-c-compat")
 
-  using __pool_scheduler_t = decltype(std::declval<exec::static_thread_pool>().get_scheduler());
+#include "__system_context_if.h"
+
+STDEXEC_PRAGMA_POP()
+
+namespace exec::__detail {
+  using stdexec::__declval;
+  using __pool_scheduler_t =
+    stdexec::__decay_t<decltype(__declval<exec::static_thread_pool&>().get_scheduler())>;
 
   /// Receiver that calls the callback when the operation completes.
-  template <class _Sender>
+  template <class _Sender, bool _IsBulk>
   struct __operation;
 
-  template <class _Sender>
+  template <class _Sender, bool _IsBulk>
   struct __recv {
     using receiver_concept = stdexec::receiver_t;
 
     /// The callback to be called.
-    __exec_system_context_completion_callback_t __cb_;
+    system_context_completion_callback __cb_;
 
     /// The data to be passed to the callback.
     void* __data_;
 
     /// The owning operation state, to be destructed when the operation completes.
-    __operation<_Sender>* __op_;
+    __operation<_Sender, _IsBulk>* __op_;
 
     void set_value() noexcept {
       __cb_(__data_, 0, nullptr);
@@ -54,10 +64,11 @@ namespace exec::__system_context_default_impl {
     }
   };
 
-  template <typename _Sender>
-  struct __operation {
+  template <class _Sender, bool _IsBulk>
+  struct __operation
+    : stdexec::__if_c<_IsBulk, system_bulk_operation_state, system_operation_state> {
     /// The inner operation state, that results out of connecting the underlying sender with the receiver.
-    stdexec::connect_result_t<_Sender, __recv<_Sender>> __inner_op_;
+    stdexec::connect_result_t<_Sender, __recv<_Sender, _IsBulk>> __inner_op_;
     /// True if the operation is on the heap, false if it is in the preallocated space.
     bool __on_heap_;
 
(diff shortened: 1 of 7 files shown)


ericniebler:
@lewissbaker for your consideration.

include/exec/__detail/__system_context_default_impl.hpp
```diff
@@ -66,123 +77,109 @@ namespace exec::__system_context_default_impl {
...
-      __operation<decltype(stdexec::schedule(std::declval<__pool_scheduler_t>()))>;
+   public:
+    static constexpr stdexec::forward_progress_guarantee forward_progress_guarantee =
+      stdexec::forward_progress_guarantee::parallel;
+
+    explicit __system_scheduler_impl(exec::static_thread_pool& __pool) noexcept
+      : exec::system_scheduler_base(this)
+      , __pool_scheduler_{__pool.get_scheduler()} {
+      auto sched = __pool.get_scheduler();
+      __pool_scheduler_ = sched;
```
ericniebler, line 132: remove debugging code
```suggestion
```

include/exec/__detail/__system_context_default_impl_entry.hpp
```diff
@@ -23,23 +23,121 @@
...
+      __detail::__unlock();
+      __new_instance->destroy_fn(__new_instance);
+      return __instance;
+    }
+
+    __detail::__unlock();
+    return __new_instance;
+  }
+
+  /// Gets the default system context implementation.
```
ericniebler, line 110:
```suggestion
  /// Releases a reference to the specified system context object.
  /// If the ref count drops to zero, then
  ///   - if the specified object is the active global object, set
  ///     the active global object to null.
  ///   - destroy the instance
```

include/exec/__detail/__system_context_default_impl_entry.hpp
```diff
@@ -23,23 +23,111 @@
...
+      }
+    };
+
+    struct __system_context_base {
+      __system_context_base() {
+        // This static will create the system context, and decrement its
+        // ref count after main exits.
+        static __system_context_keep_alive __s_keep_alive;
+      }
+    };
```
lewissbaker, line 60: The `__system_context_base` class doesn't seem to be used anywhere. What is it's purpose?

Also, the `__system_context_keep_alive` class is only used by `__system_context_base` and so seems to be dead-code.

include/exec/__detail/__system_context_default_impl_entry.hpp
```diff
@@ -23,23 +23,111 @@
...
+    // is the current system context from the time the first system
+    // context object is created until after main exits.
+    struct __system_context_keep_alive {
+      __system_context_keep_alive() {
+        (void) get_system_context_instance();
+      }
+
+      ~__system_context_keep_alive() {
+        release_system_context_instance(__system_context_instance.load());
+      }
```
lewissbaker, line 51: If this class were to actually be used, I would have concerns about it potentially incrementing the ref-count on one system-context object and then decrementing the ref-count on another system-context object if anyone ever called `set_new_system_context_handler` during the lifetime of this object.

include/exec/system_context.hpp
```diff
@@ -83,19 +100,118 @@ namespace exec {
     template <class _Sender>
     using __sender_data_t = decltype(stdexec::sync_wait(std::declval<_Sender>()).value());
 
+    /// A very simple spin lock. It is used in leu of a mutex because
```
lewissbaker, line 103: typo: leu -> lieu

include/exec/system_context.hpp
```diff
@@ -83,19 +100,118 @@ namespace exec {
...
+    struct __spin_lock {
+      constexpr __spin_lock() noexcept = default;
+
+      void __lock() noexcept {
+        bool __is_locked = false;
+        // spin until we get the lock:
+        while (!__locked_.compare_exchange_strong(__is_locked, true)) {
+          std::this_thread::yield();
+          __is_locked = false;
+        }
```
lewissbaker, line 115: You probably want to use `atomic::wait()` and `atomic::notify_all()` to do the appropriate backoff here. Alternatively, you can use a binary_semaphore as a kind of spin-wait mutex.
Just initialize to have a count of 1 and then lock() calls acquire() and unlock() calls release().

include/exec/system_context.hpp
```diff
@@ -83,19 +100,118 @@ namespace exec {
...
 
   class system_scheduler;
   class system_sender;
-  template <stdexec::sender _S, std::integral _Size, class _Fn>
+  template <stdexec::sender _Sndr, std::integral _Size, class _Fn>
   class system_bulk_sender;
 
+  /// @brief A base class that helps populate the system_context_interface
+  ///        using members of the derived class.
+  class system_context_base : public system_context_interface {
```
lewissbaker, line 134: I don't think this type should be part of the public interface.
Can we name it with underscores instead? i.e. `__system_context_base`

include/exec/__detail/__system_context_default_impl_entry.hpp
```diff
@@ -23,23 +23,111 @@
...
+    auto* __instance_copy = __instance;
+    (void) __detail::__system_context_instance.compare_exchange_strong(__instance_copy, nullptr);
+
+    __detail::__system_context_mutex.__unlock();
+    __instance->destroy_fn(__instance);
+  }
+
+  /// Sets the default system context implementation.
+  extern "C" STDEXEC_SYSTEM_CONTEXT_INLINE new_system_context_handler
+    set_new_system_context_handler(new_system_context_handler __handler) {
```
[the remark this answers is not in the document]
lewissbaker: Actually, I noticed that instead you have the ref-count initialized to 1 instead of to 0.

This instead means that nothing will ever destroy the system context object as nothing ever releases the last ref-count (assuming that `system_context` objects have matching calls to `get_system_context_instance()` and `release_system_context_instance()`).

include/exec/__detail/__system_context_default_impl_entry.hpp
lewissbaker: Also, if the intention is that the `__system_context_base` is intended to be a base-class of the system-context object then having the `__system_context_keep_alive` constructor call `get_system_context_instance()` is going to end up always calling the system-context construction function twice.

The first cal to `get_system_context_instance()` by the user is going to construct the default system-context, then the constructor of that system-context object is going to call `get_system_context_instance()` again, which is then going to try to construct a second system-context object as we're not yet finished constructing the first one, which is going to then call into the constructor of the `__system_context_base` again.

Best case, this will redundantly construct two system-context objects, install one and then have the second one replace the first, destroying the first. Worst-case, you'll hit a terminate() because you are reentrantly trying to initialize a function-local static object.

lewissbaker (changes requested):
I've made an initial pass over the logic for run-time replaceability. I haven't yet looked in detail at the user-facing part of the API. I have a set of other issues  with that side of things but will leave that for another forum.
There are a lot of issues in the runtime replacement logic, though.

As @kirkshoop mentioned elsewhere, this kind of initialization is very difficult to get right unless you really restrict what is allowed and is likely to be a bug-farm, both for implementers and for people implementing replacement system-contexts.

include/exec/__detail/__system_context_default_impl_entry.hpp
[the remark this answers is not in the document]
ericniebler: That is what keep_alive is for. It releases the final ref count.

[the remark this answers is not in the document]
lewissbaker: If it is the intent of the keep_alive to release the final ref-count then I don't think that the keep_alive constructor should be trying to increment the ref-count by calling `get_system_context_instance()` as you are already initializing the ref-count to 1.

However, it's hard to evaluate exactly where keep_alive constructor is being invoked as that code doesn't seem to be called from anywhere at the moment.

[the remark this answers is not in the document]
ericniebler: `system_context` is meant to inherit from `__system_context_base`, so that the when the system context object is first created, its ref-count is `2` with the second ref-count held by the keep-alive until after main.

include/exec/__detail/__system_context_default_impl_entry.hpp
```diff
@@ -23,23 +23,111 @@
...
+    // if it is not null, increment the reference count and return the ptr
+    if (__instance != nullptr) {
+      ++__instance->ref_count;
+      __detail::__system_context_mutex.__unlock();
+      return __instance;
+    }
+    __detail::__system_context_mutex.__unlock();
+
+    // create a new instance:
+    auto* __new_instance = __detail::__new_system_context_handler.load()();
```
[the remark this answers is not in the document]
ericniebler: This complexity is why I added `static_system_context_instance`.

include/exec/__detail/__system_context_default_impl_entry.hpp
ericniebler: That was an oversight. `system_context` should inherit from `__system_context_base`. I have fixed that.

include/exec/__detail/__system_context_default_impl_entry.hpp
ericniebler: > I would have concerns about it potentially incrementing the ref-count on one system-context object and then decrementing the ref-count on another system-context object

that is precisely the design. `set_new_system_context_handler` is responsible for fixing up the ref-counts when the system context object changes. i've said this several times already.

include/exec/__detail/__system_context_default_impl_entry.hpp
```diff
@@ -23,23 +23,111 @@
...
+  namespace __detail {
+    STDEXEC_SYSTEM_CONTEXT_INLINE constinit __spin_lock __system_context_mutex{};
+
+    STDEXEC_SYSTEM_CONTEXT_INLINE constinit std::atomic<system_context_interface*>
+      __system_context_instance{nullptr};
+
+    STDEXEC_SYSTEM_CONTEXT_INLINE constinit std::atomic<new_system_context_handler>
+      __new_system_context_handler{[]() -> system_context_interface* {
+        return new __detail::__system_context_impl{};
+      }};
```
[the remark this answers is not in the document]
ericniebler: they are modified outside of the lock in some places.

include/exec/__detail/__system_context_default_impl.hpp
```diff
@@ -15,31 +15,41 @@
...
 #include "exec/static_thread_pool.hpp"
 
-namespace exec::__system_context_default_impl {
-  using namespace stdexec::tags;
+STDEXEC_PRAGMA_PUSH()
+// "warning: empty struct has size 0 in C, size 1 in C++"
+// The intended use of these two structs is as a base class for other structs.
+// When used like that, the size occupied by these structs is 0 because of the
+// empty base class optimization.
+STDEXEC_PRAGMA_IGNORE_GNU("-Wextern-c-compat")
```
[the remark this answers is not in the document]
ericniebler: mostly laziness. because the PRAGMA macros are defined in a C++ header that I don't want to include in `__system_context_if.h`.

include/exec/__detail/__system_context_default_impl_entry.hpp
```diff
@@ -23,23 +23,111 @@
...
+    set_new_system_context_handler(new_system_context_handler __handler) {
+    auto __old_handler = __detail::__new_system_context_handler.exchange(__handler);
+    auto* __new_instance = __handler();
+
+    auto* __instance = __detail::__system_context_instance.exchange(__new_instance);
+
+    if (__instance != nullptr) {
+      release_system_context_instance(__instance);
+    }
+    return __old_handler;
```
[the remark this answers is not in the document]
ericniebler: i don't care about the complexity. i care about correctness, and i don't see a reason to limit how many times the system context is replaced.

include/exec/__detail/__system_context_if.h
```diff
@@ -21,84 +21,121 @@
 #include <stdint.h>
 
 #ifdef __cplusplus
-extern "C" {
+#  define STDEXEC_SYSTEM_CONTEXT_NOEXCEPT noexcept
+namespace exec {
+  extern "C" {
```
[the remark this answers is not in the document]
ericniebler: it was part of @lucteo's original design and i opted to keep it. i'm not attached.

include/exec/__detail/__system_context_if.h
```diff
@@ -21,84 +21,121 @@
...
+  ///    []() -> exec::system_context_interface* {
+  ///      return new my_system_context_impl{};
+  ///    });
+  new_system_context_handler set_new_system_context_handler(new_system_context_handler /*handler*/);
+
+  /// Callback to be called by the scheduler when new work can start.
+  typedef void (*system_context_completion_callback)(
+    void*,  // data pointer passed to scheduler
+    int,    // completion type: 0 for normal completion, 1 for cancellation, 2 for exception
+    void*); // If completion type is 2, this is the exception pointer.
```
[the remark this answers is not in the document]
ericniebler: changing the "virtual" interfaces of the system context isn't within the scope of this PR

ericniebler:
i think i have a bug where if `set_new_system_context_handler` is called after the global system context object has been cleaned up, a new system context object is correctly installed, but nobody will decrement the ref count so it will never be cleaned up.

include/exec/system_context.hpp
ericniebler: no, it is intended to be a public part of the API. it is a utility to ensure that the "virtual table" is filled out correctly, the version number is set appropriately, and that the ref count is properly set to 1.

lewissbaker: So are users implementing their own system-context _required_ to use this?
Or are we defining the semantics enough that it is well-specified how to initialize a `system_context_interface` structure directly?

include/exec/__detail/__system_context_if.h
```diff
@@ -21,84 +21,121 @@
...
+  union __system_context_interface_reserved {
+    void* __object_ptr;
+    void* (*__function_ptr)();
+    uintptr_t __int;
+  };
+
+  /// Interface that allows interaction with the system context, allowing scheduling work on the system.
+  struct system_context_interface {
+    /// The supported version of the system context interface, in the form YYYYMM.
+    uint32_t version;
```
lewissbaker, line 43: It's not clear to me how this version number is planned to be used.

What will the client do if it sees a backend implementation that has a newer version than the version currently supported? Are all future versions guaranteed to be backwards compatible with older versions?
What if the client sees a backend implementation that is older that the current version? Will it need to introduce additional branches everywhere to check the version on every field access for fields which were added after the first version?

include/exec/system_context.hpp
ericniebler: > So are users implementing their own system-context required to use this?

no

include/exec/__detail/__system_context_if.h
ericniebler: > Are all future versions guaranteed to be backwards compatible with older versions?

yes, so that client code compiled against version N can still work with a system context from version N+1.

> What if the client sees a backend implementation that is older that the current version? Will it need to introduce additional branches everywhere to check the version on every field access for fields which were added after the first version?

Yes, new client code needs to do a version number check before using any field that was added after v1.

ericniebler:
superseded by #1394

Issue #1371, opened 2024-07-10
Configuring CMake fails


stephenberry:
Configuring CMake with stdexec as a dependency works the first time entirely from scratch (on Mac/Windows). However, subsequent calls to CMake will produce the following error:
```
CMake Error at build/_deps/stdexec-src/test/CMakeLists.txt:140 (include):
  include could not find requested file:

    /contrib/Catch.cmake


CMake Error at build/_deps/stdexec-src/test/CMakeLists.txt:142 (catch_discover_tests):
  Unknown CMake command "catch_discover_tests".
```

Deleting the cache seems to work, except on Linux which seems to have this problem more persistently.

We're configuring with STD_BUILD_TESTS turned off.

ericniebler:
stdexec doesn't use `STD_BUILD_TESTS` for anything. it builds tests if either of `STDEXEC_BUILD_TESTS` or `BUILD_TESTING` is set to a truthy value.

i have never seen this error, but then again i don't typically build stdexec as a dependency. i'll see if i can repro when i have a chance.

wsberry:
I am finding the bug to be intermittent. Two observations when this error occurs:
1.  the catch repo was not included in the `CMAKE_BINARY_DIR/_deps` folder (it was not downloaded)
2. the path error indicates `contrib/Catch.cmake` when it should be something line `${CMAKE_BINARY_DIR}/_deps/catch2-src/contrib/Catch.cmake`

wsberry:
I avoided this error by not using CPM in my project:

```cmake
function (fetch_stdexec)
  set(branch_or_tag "main")
  set(url "https://github.com/NVIDIA/stdexec.git")
  set(target_folder "${CMAKE_BINARY_DIR}/_deps/stdexec-src")

  if (NOT EXISTS ${target_folder})
    execute_process(
        COMMAND git clone --depth 1 --branch "${branch_or_tag}" --recursive "${url}" "${target_folder}"
        WORKING_DIRECTORY ${CMAKE_BINARY_DIR}
        RESULT_VARIABLE exec_process_result
        OUTPUT_VARIABLE exec_process_output
    )
    if(NOT exec_process_result EQUAL "0")
        message(FATAL_ERROR "Git clone failed: ${exec_process_output}")
    else()
        message(STATUS "Git clone succeeded: ${exec_process_output}")
    endif()
  endif()

[...]

PR #1380, opened 2024-07-22
Prevent CMake from attempting to configure test executables without Catch2

ednolan:
Previously, the CMake would check whether STDEXEC_BUILD_TESTS was set when determining whether CPM should add Catch2, but it would then check BUILD_TESTING when determining whether to add the CMake for the test subdirectory. If STDEXEC_BUILD_TESTS was enabled, this would work, because the CMake would enable BUILD_TESTING whenever STDEXEC_BUILD_TESTS was enabled. However, if BUILD_TESTING was enabled but STDEXEC_BUILD_TESTS was disabled, then the CMake would attempt to configure the tests without having added Catch2, causing the build configuration to fail.

This was a particular problem when adding stdexec as a submodule of another project, since if that project had an include(CTest) command invocation before it added stdexec, and it did not explicitly enable STDEXEC_BUILD_TESTS, the build configuration would fail, since STDEXEC_BUILD_TESTS is disabled by default when stdexec is a submodule, and include(CTest) automatically sets BUILD_TESTING.

This commit addresses the issue by setting BUILD_TESTING to whatever STDEXEC_BUILD_TESTS was set to, and then relying on BUILD_TESTING to decide both whether to add Catch2 and whether to configure the test executables. It also adds a check for BUILD_TESTING when determining how to set the default for STDEXEC_BUILD_TESTS when the user doesn't configure it. The outcome is that, if the user explicitly enables or disables STDEXEC_BUILD_TESTS, then the CMake will use it to override the preexisting BUILD_TESTING setting; but otherwise, the CMake will enable tests only if BUILD_TESTING is set or if stdexec is not a submodule.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -79,13 +79,24 @@ rapids_cmake_build_type(Release)
 list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/cmake/Modules")
 
 # Don't build tests if configuring stdexec as a submodule of another
-# CMake project, unless they explicitly set STDEXEC_BUILD_TESTS=TRUE
-if (CMAKE_PROJECT_NAME STREQUAL PROJECT_NAME)
+# CMake project, unless they explicitly set STDEXEC_BUILD_TESTS=TRUE,
+# or they enabled CTest's BUILD_TESTING option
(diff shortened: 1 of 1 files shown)


ednolan:
This addresses issue 1371: https://github.com/NVIDIA/stdexec/issues/1371#issuecomment-2241949203

ericniebler:
Thanks!

PR #1386, opened 2024-07-28
add sender factory `exec::just_from`

ericniebler:
`just_from(fn)` creates a sender that completes inline by passing a "sink" function to
`fn`. Calling the sink function with arguments sends the arguments as values to the receiver.

The sink function passed to `fn` must be called exactly once.

Example:
```cpp
// The following sender is equivalent to just(42, 3.14):
auto sndr = exec::just_from([](auto sink) { return sink(42, 3.14); });
```

The function passed to `just_from` must return an instance of a specialization of
`stdexec::completion_signatures<>` that describes the ways the sink function might be
invoked. The sink function returns such a specialization of `stdexec::completion_signatures<>`
corresponding to the arguments passed to it, but if your function uses the sink function
in several different ways, you must specify the return type explicitly.

Example:
```cpp
auto sndr = exec::just_from(
  [](auto sink) {
    if (some-condition) {
      sink(42);
    } else {
      sink(3.14);
    }
    return stdexec::completion_signatures<stdexec::set_value_t(int),
                                          stdexec::set_value_t(double)>{};
  });
```

### DIFF
--- a/include/exec/just_from.hpp
+++ b/include/exec/just_from.hpp
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2022 NVIDIA Corporation
+ * Copyright (c) 2024 NVIDIA Corporation
  *
  * Licensed under the Apache License Version 2.0 with LLVM Exceptions
  * (the "License"); you may not use this file except in compliance with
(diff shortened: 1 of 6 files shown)


include/exec/just_from.hpp
```diff
@@ -0,0 +1,235 @@
+/*
+ * Copyright (c) 2022 NVIDIA Corporation
```
ericniebler, line 2:
```suggestion
 * Copyright (c) 2024 NVIDIA Corporation
```

PR #1393, opened 2024-08-02
Revert "Remove `__destruct_schedule_operation` from system_context ABI (#1332)"

ericniebler:
This reverts commit bfad35d6f6acc41387b78ead351e20ba7eff52be.

@lucteo Looks like the system context tests got flaky again after landing this. Reverting.


lucteo:
@ericniebler : Sorry, I wanted to say that we need to abandon this. I'm starting to move to Lewis' replaceability API (it would be getting easier consensus than what I had previously)

lucteo:
After changing the "replaceability API", I'm still getting frequent crashes when trying to destroy the operation state (somewhere deep in static_thread_pool). I will complete my API work and then come back to this issue.

PR #1394, opened 2024-08-05
Define "replaceability API" in terms of C++ interfaces.

lucteo:
- Get rid of old C interface
- using `query_system_context` to be able to retrieve different backend implementations
- allows both run-time and link-time replaceability


ericniebler:
This looks good, thanks @lucteo!

include/exec/system_context.hpp
```diff
@@ -241,47 +210,109 @@ namespace exec {
...
+      /// Storage for the arguments received from the previous sender.
+      alignas(__storage_t) unsigned char __arguments_data_[sizeof(__storage_t)];
+    };
+
+    /// Derived class that properly forwards the arguments received from `_Previous` to the receiver methods.
+    /// Uses the storage defined in the base class. No extra data is added here.
+    template <class _Previous, class _BulkState, class... _As>
+    struct __typed_forward_args_receiver : __forward_args_receiver<_Previous> {
+      using __base_t = __forward_args_receiver<_Previous>;
+      using __rcvr_t = _BulkState::__rcvr_t;
```
ericniebler, line 251:
```suggestion
      using __rcvr_t = typename _BulkState::__rcvr_t;
```

include/exec/__detail/__system_context_replaceability_api.hpp
```diff
@@ -0,0 +1,87 @@
...
+    /// Called for each item of a bulk operation, possible on different threads.
+    virtual void start(uint32_t) noexcept = 0;
+  };
+
+  /// Describes a storage space.
+  /// Used to pass preallocated storage from the frontend to the backend.
+  struct storage {
+    void* __data;
+    uint32_t __size;
+  };
```
lewissbaker, line 61: I am still not convinced that the API needs to allow callers to pass in a potentially dynamic amount of storage that the implementation can use.

The callers are all going to be compiled with a fixed amount of storage defined by whatever the public operation-state types defined by the stdlib contain.

It is not clear to me what the use-case is that you are trying to support by making the amount of storage dynamic here.

If you want to be able to change the amount of storage reserved by the calling code then you'll need to presumably set some #define on the command-line which changes the size of the reserved buffer in the operation-state. However, this would also require the recompilation of all code that uses the system_scheduler to avoid ODR-violations.

Is the intent here to support the case where you recompile all client code to reserve more/less storage in the operation-states, but _don't_ recompile the implementation of the system-context interface at the same time?

If this is the case, then we should discuss the trade-offs.

One concern I have with the approach outlined below is that it forces the implementation to store the additional parameters in the storage provided.
e.g. with the `bulk_schedule()` example below you are passing the storage, and then also passing the count, receiver pointer and bulk_item_receiver pointers.
If the implementation wants to avoid dynamic allocation then it will need to store the values for these additional parameters inside the 'storage' provided by the user, before it even starts to think about storing additional internal state (e.g. linked list items).

However, the user-facing operation-state class probably already needs to store these things. It at least needs to store the count somewhere so it can pass it into the `bulk_schedule()` call in `bulk_op_state::start()`.

The `receiver` and `bulk_item_receiver` objects (or at least the concrete types that inherit from these interfaces) can just be sub-objects of the `bulk_op_state` class. However, since they are both virtual, you will be storing two vtable pointers in the `bulk_op_state`.

Further, if you need to implement stop-token based cancellation, then you will need to construct a stop-callback object in start() and destroy it when one of the receiver methods is called. However, this will probably mean that the `receiver` object will need to get access to the `bulk_op_state` object that owns the `receiver` object, which it can't do in general (it's not allowed to cast from the address of a data-member to the address of the containing class except in very limited circumstances), so you'll probably need to store a pointer to the op-state in the receiver passed in, adding another pointer of storage to the op-state class.

i.e. it would look something like:
```
template<typename Rcvr, typename F>
struct bulk_op_state {
  struct rcvr_t : system_context_replaceability::receiver {
    // implicit vtable*
    bulk_op_state* op;
    void set_value() noexcept {
      op->stop_callback.reset();
      op->parent_rcvr.set_value();
    }
    // etc. for set_error/set_stopped
  };
  struct bulk_rcvr_t : system_context_replaceability::bulk_item_receiver {
    // implicit vtable*
    bulk_op_state* op;
    
    void start(uint32_t idx) noexcept {
      op->func(idx);
    }
  };
  
  system_scheduler* scheduler;
  Rcvr parent_rcvr;
  uint32_t count;
  F func;
  rcvr_t rcvr;
  bulk_rcvr_t bulk_rcvr;
  
  // 'storage' needs to hold storage for:
  // - &rcvr
  // - &bulk_rcvr
  // - count
  // - next/prev linked list pointers
  uintptr_t storage[5];
};
```

The net result of all of this is that the op_state class ends up being up to 7 pointers larger than it needs to be.
- 2 vtable pointers for 'rcvr' and 'bulk_rcvr'
- 2 op-state pointers for 'rcvr' and 'bulk_rcvr'
- extra storage in 'storage' for '&rcvr', '&bulk_rcvr', 'count'

Further, in your implementation you are storing the 'parent_rcvr' object as a sub-object of the 'rcvr' object. However, this inihibits the optimization for operation-states that allows avoiding storing the parent receiver by instead recomputing it from the address of the parent-operation-state (adding overhead of an additional pointer).

With different design choices you should be able to reduce all of this down to a single vtable pointer and also allow the optimization that avoids needing to store the parent-receiver, potentially saving 56 bytes in the operation-state for a bulk operation.

Let's assume a stateless lambda for 'f' and that the parent receiver can be optimized away.
Then the total operation-state size should be able to get down to:
- pointer to scheduler interface
- count
- storage for internal use (say 2 pointers for next/prev linked list)
This is a total of 4 pointers in size, compared to 11 pointers in size with the approach suggested in this PR.

The implementation would look something (roughly) like this:
```c++
struct system_context_bulk_op_base {
  explicit system_context_bulk_op_base(uint32_t count) noexcept
  : count(count) {
     for (auto& x : internal_storage) x = 0;
  }
  virtual void set_value() noexcept = 0;
  virtual void set_error(std::exception_ptr) noexcept = 0;
  virtual void set_stopped() noexcept = 0;
  virtual void execute_bulk_item(uint32_t index) noexcept = 0;
  uintptr_t internal_storage[2];
  uint32_t count;
};

template<typename Rcvr, typename Func>
struct system_context_bulk_op final
: public receiver_storage<system_context_bulk_op<Rcvr>, Rcvr> // handles storing Rcvr (or not)
, private system_context_bulk_op_base {
  system_context_bulk_op(system_scheduler_if* sched, Func func, Rcvr rcvr)
  : receiver_storage<system_context_bulk_op<Rcvr>, Rcvr>(std::move(rcvr))
  , system_context_bulk_op_base(count)
  , sched(sched)
  , func(std::move(func))
  {}
  
  void start() noexcept {
    sched->bulk_start(this);
  }
  
private:
  void set_value() noexcept override {
    this->get_receiver().set_value();
  }
  
  void set_error(std::exception_ptr e) noexcept override {
    this->get_receiver().set_error(std::move(e));
  }
  
  void set_stopped() noexcept override {
    this->get_receiver().set_stopped();
  }
  
  void execute_bulk_item(uint32_t index) noexcept {
    std::invoke(func, index);
  }
  
  system_scheduler_if* sched;
  [[no_unique_address]] Func func;
};

struct system_scheduler_if {
  virtual void bulk_start(system_context_bulk_op_base* op) noexcept = 0;
  // ...
};
```

I really do think we need to be more careful about the data-structure API design decisions we make here as they can potentially have large impacts on the overhead of the resulting system, particularly in terms of operation-state size, but also in terms of extra operations needed to populate all of these unnecessary extra fields.

include/exec/__detail/__system_context_replaceability_api.hpp
```diff
@@ -0,0 +1,87 @@
...
+#ifndef STDEXEC_SYSTEM_CONTEXT_REPLACEABILITY_API_H
+#define STDEXEC_SYSTEM_CONTEXT_REPLACEABILITY_API_H
+
+#include "stdexec/__detail/__execution_fwd.hpp"
+
+#include <typeindex>
+
+
+/// Implementation-defined mechanism of querying a system context interface of type `id`.
+extern void* __query_system_context_interface(std::type_index id) noexcept;
```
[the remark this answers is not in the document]
lucteo: On the usage of `std::type_index`.

Fair point. Please note that this is an QOI issue, and not something we want to specify in the proposal. I could imagine implementations that would use simple enums here.
I will change our implementation to reflect the schema with UUIDs.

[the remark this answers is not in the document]
lucteo: On constness:

If the replaceability API is with C++ classes, then `const` doesn't make sense. Operations exposed on the interface may not be `const`.
If we are moving back to a C-like interface, then it probably makes sense to have it `const`.

include/exec/__detail/__system_context_replaceability_api.hpp
```diff
@@ -0,0 +1,87 @@
...
+    /// Schedule work on system scheduler, calling `__r` when done and using `__s` for preallocated memory.
+    virtual void schedule(storage __s, receiver* __r) noexcept = 0;
+    /// Schedule bulk work of size `__n` on system scheduler, calling `__br` for each item, calling `__r` when done and using `__s` for preallocated memory.
+    virtual void
+      bulk_schedule(uint32_t __n, storage __s, receiver* __r, bulk_item_receiver* __br) noexcept = 0;
+    /// Returns the maximum concurrency supported by the system scheduler.
+    virtual uint32_t max_concurrency() noexcept = 0;
+    /// Get the forward progress guarantee promised by the system scheduler.
+    virtual stdexec::forward_progress_guarantee get_forward_progress_guarantee() noexcept = 0;
+  };
```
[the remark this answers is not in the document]
lucteo: While I agree with the argument, it feels a bit over-engineering. The cost of an extra load is extremely small compared to the cost of scheduling work on other threads.

include/exec/__detail/__system_context_replaceability_api.hpp
lucteo: I may have missed some of the details from this description, but I hope I got the main message.

I would still argue that we need a storage that the host part need to pass to the backend part.

The backend needs to have some implementation details for bulk (e.g., `static_thread_pool` bulk's operation state), and need to store that data somewhere. It can't store it locally to the `bulk_schedule` function, as this terminates before the operation completes. Thus, in the absence of extra storage, the backend is forced to have a heap memory allocation in every case. This is unacceptable from my point of view.

The fact that the storage has a dynamic size provides the following benefits:
* allows the same backend code to work with multiple frontends; for example different standard lib implementations may choose different sizes; we can't hardcode the right size in the API, and the backends need a way to know that size
* allow different parts of the same stdlib frontend to use different sizes; I can easily see extensions of stdlib that allows customizing the amount of storage to be used; I was thinking of implementing something like that in stdexec.

I will go again over the replaceability API and backend to ensure that we don't store data that we don't need. You pointed out some places in which extra data is stored.

include/exec/__detail/__system_context_replaceability_api.hpp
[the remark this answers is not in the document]
lewissbaker: > On constness

I agree - if the pattern is to return a struct containing function-pointers then returning a pointer-to-const makes sense. If on the other hand, you are returning a pointer to an object with virtual functions, some of which are logically non-const, then I agree it does not makes sense to return pointer-to-const here.

> I could imagine implementations that would use simple enums here.

One of the use-cases I can imagine people may want to support is the ability to add implementation-specific or user-defined extensions. e.g. to be able to query an interface onto the system context that gives access to qt-specific, or libdispatch-specific facilities, or possibly to query interfaces onto algorithm-specific optimization interfaces. e.g. to allow algorithms to dispatch to an optimized implementations instead of some default implementation in terms of schedule() or bulk().

include/exec/__detail/__system_context_replaceability_api.hpp
lewissbaker: > I would still argue that we need a storage that the host part need to pass to the backend part.

I am not disagreeing with you there. In my code example above, it is still passing in storage that the backend can use for storing additional data. The `internal_storage` member is reserved for use by the backend for whatever purpose it requires. With the idea that the standard guarantees at least some minimum number of bytes/alignment here, although implementations are free to reserve more.

> allows the same backend code to work with multiple frontends; for example different standard lib implementations may choose different sizes; we can't hardcode the right size in the API, and the backends need a way to know that size

Won't the backend need to be compiled separately for each possible stdlib implementation it is to be used against anyway? If the backend uses _any_ standard library facilities then compiling it against one stdlib and linking the backend lib against user-code compiled with a different stdlib just seems like it's asking for ODR violations. e.g. at a minimum, we might be passing std::error_code or std::exception_ptr objects across the API boundary here.

And if you are compiling your backend differently for each stdlib then it seems like it would be reasonable to use that stdlib's definitions of types that define how much storage is reserved for the implementation. A backend implementation can query statically how much storage is provided and then determine which strategy it is going to use (e.g. using inline storage vs heap-allocating).

If one of the goals of this API is that we can compile a system_context backend once and link it against code compiled with different stdlib implementations then I think this is something that needs to be called out in the paper explicitly and the implications discussed - I am not convinced that this is a goal that is achievable looking superficially at it.

lucteo: The cost of passing a size for the storage is minimal: an extra parameter to the `schedule` function. This isn't stored the operation state.

Making the backend depend on the compilation of the frontend seems a design red-flag.

As I said above, I can envision that different parts on the client-side might use different sizes for preallocated storage. Thus, there isn't just one frontend for the backend to depend on. 

If we have multiple modules A, B, C that all use system_context, I don't necessarily need to build them all at once to be able to replace the backend for system context.

> With the idea that the standard guarantees at least some minimum number of bytes/alignment here, although implementations are free to reserve more.
Having a guaranteed greater than zero may inhibit optimization in some cases. Reserving storage always has costs: the operation-state becomes bigger. If an application uses specialized allocators to ensure that the memory allocation cost is small, and allocating on the heap is more frequent, then having that extra storage is just a cost.

I see no reason of mandating this in the paper. Let the implementations choose the best strategy.

lucteo: Yesterday I started looking at ways to reduce the size of the operation states. I found a few cases, but none of them were related to the definition of the API; they were all implementation-specific issues. Will continue on this front...

lewissbaker: > I can envision that different parts on the client-side might use different sizes for preallocated storage. Thus, there isn't just one frontend for the backend to depend on.

Can you elaborate on how this might be achieved given the user-facing API we will be exposing? How will different client-code cause the operation-state for a system_scheduler schedule operation to reserve different amounts of storage? And given that it will (presumably) invoke the same backend function, what would be the benefit in passing different amounts of operation-state-reserved storage here?

I feel like I'm missing some use-case/requirement/assumption that you have.

In your example about modules A, B, C and D being built at separate times - I would still expect each of these modules to reserve the same amount of storage for the same operation-state object type for a schedule operation - to do otherwise would be an ODR violation.
I'm not saying that you need to recompile user code every time you change the backend - only if you use a new stdlib (or new configuration of a stdlib that chooses a different reserved storage size).

lewissbaker: Is the idea that we would define the system_scheduler operation-states such that if the receiver environment provides an allocator then we don't reserve storage in the operation-state and instead call the allocator to allocate some storage for use internally?

lucteo: I envision that each implementation will have its own constants for operation sizes & alignment. Moreover, I envision that the standard library will provide implementation-defined ways of changing these constants. That is, if the standard implementation requires 100 bytes of storage, I may replace the backend with something that requires 120 bytes, thus the library should allow me to change these constants. Probably changing this will be at compile-time.

Thus, I may set different values to modules A, B and C. If the replacement of these constant will not not create ODR problems if these details are not exposed to the linker.

If the `get_system_scheduler()` function is inline, and uses these constants to produce different types (e.g., template classes parameterized by op-state size), then there isn't any ODR violation when linking the modules together.

Extending this rationale, we can have different values of these sizes in the same module. Imagine the following definition for getting the system scheduler:
```c++
template <size_t _Size = 100, size_t _Alignment = 8>
inline auto get_system_scheduler() {
    return __impl_defined_system_scheduler<_Size, _Alignment>{__get_system_scheduler_backend()};
}
```

This can be used in the same module with different values.

One example in which this might happen:
* we replace the system scheduler backend, and the new backend requires a bigger size
* we have places the uses system scheduler in the standard library -- these would use the default values
* we use system scheduler in some places that we fully control, and we use appropriate sizes for these places.

-----

I was not thinking before that this reserved storage (needed at compile-time) would work with receiver-provided allocators. But I think you make a good point. We can make it work with allocators. We can make the frontend part allocate size for the the operation state, and pass it to the backend. But, I'm not sure if this would actually be something that will be beneficial in practice.

lewissbaker: > Thus, I may set different values to modules A, B and C.
Just to clarify, are you talking about "modules" as in shared libraries? Or "modules" as in separately compiled C++20 modules?

lucteo:
@lewissbaker : please see the latest patch for an analysis of the operation-state size. I've made some improvements; the final size of the bulk op state almost halved. There may be other improvements possible, but we are reaching the law of diminishing returns.

include/exec/__detail/__system_context_replaceability_api.hpp
lucteo: modules as being shared libraries. Sorry about the confusion.

PR #1395, opened 2024-08-11
Reproduce split race/crash

ccotter:
https://gist.github.com/ccotter/c0504455c7d54240758a34dc99d620f0 is the ASAN report.

From https://github.com/NVIDIA/stdexec/blob/54b38c99e8cbb348c1dd3508fa540728882247db/include/stdexec/__detail/__shared.hpp#L247 the shared state becomes a dangling pointer if if https://github.com/NVIDIA/stdexec/blob/54b38c99e8cbb348c1dd3508fa540728882247db/include/stdexec/__detail/__shared.hpp#L346 frees the memory from another thread.

I'm not quite sure how to fix this, but this PR manages to reproduce this bug reliably with ASAN in the "split forwards results from a different thread"" test.

### DIFF
--- a/include/stdexec/__detail/__shared.hpp
+++ b/include/stdexec/__detail/__shared.hpp
@@ -342,6 +342,7 @@ namespace stdexec {
 
         // Set the "completed" bit in the ref count. If the ref count is 1, then there are no more
         // waiters. Release the final reference.
+        usleep(1000);
         if (__count(this->template __set_bit<__completed_bit>()) == 1) {
           this->__dec_ref(); // release the extra ref count, deletes this
         }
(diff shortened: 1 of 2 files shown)


ccotter:
I naively tried to make a copy of the `intrusive_ptr` inside `__detach` to keep the ref count positive long enough to ensure the object is kept alive, but I wasn't able to get that to work.

ericniebler:
A million thanks for this, Chris. I'll investigate as soon as I can, but probably not before next week. Big presentation coming up.

ericniebler:
I think #1460 fixes this

PR #1396, opened 2024-08-14
System context should use libdispatch if available.

lucteo:
We are using two backends for system context:
- libdispatch
- default (based on static_thread_pool)

Even if libdispatch is used, the default backend is also implicitly used in replaceability tests.


lucteo:
This won't put my changes on top of my previous branch... dropping this PR, and waiting for the first PR to be merged.

PR #1400, opened 2024-08-21
Proof of concept: Relacy tests

ccotter:
[Relacy](https://www.1024cores.net/home/relacy-race-detector) is a runtime race detector preceding ThreadSanitizer, developed by the same author @dvyukov. Relacy simulates threads as fibers and uses a special scheduling algorithm to force as many different interleavings as possible. Relacy is well suited for unit testing concurrent data structures.

I originally found #1395 by using a modified TSAN that changed `pthread_mutex_unlock` to insert a random `usleep` for up to 1000 milliseconds. A colleague pointed me to Relacy as far superior tool to hit rare thread interleavings, so as an exercise, I wanted to see if I could catch the same bug with Relacy. This draft PR contains instructions below for catching the same bug in #1395.

Relacy's output for the `split_bug` shows the full history of the variables that it is able to track, with a limited snippet shown below, including commentary added by me:

```
// Thread 0 decrements the state ref count from 2 to 1
// note: the first two bits of the atomic are reserved bits, so value=9 -> ref count 2, value=5 -> ref count 1
[363] 0: <0x71e130> fetch_sub , prev=9, arg=4, new=5, order=acq_rel, in __dec_ref_, stdexec/include/stdexec/__detail/__intrusive_ptr.hpp(111)

// Thread 2 decrements the ref count from 1 to 0, triggering the deallocation of the shared state
[365] 2: <0x71e130> fetch_sub , prev=7, arg=4, new=3, order=acq_rel, in __dec_ref_, stdexec/include/stdexec/__detail/__intrusive_ptr.hpp(111)
[366] 2: <0x71e080> atomic load, value=^@, order=relaxed, in ~inplace_stop_source, stdexec/include/stdexec/__detail/../stop_token.hpp(244)
[367] 2: memory deallocation: addr=0x71e080, in , (0)

// Thread 0 tries to access the state object which was just freed by thread 2 above.
[368] 0: <0x71e094> access to freed memory, in __try_lock_unless_stop_requested_, stdexec/include/stdexec/__detail/../stop_token.hpp(306)
[369] 0: ACCESS TO FREED MEMORY (access to freed memory), in __try_lock_unless_stop_requested_, stdexec/include/stdexec/__detail/../stop_token.hpp(306
```

Much of Relacy is macro based and hijacks standard library names with tricks like defining aliases from `std::atomic` to `rl::atomic`. As such, it's not as robust as ThreadSanitizer for instance. For concurrent data structures/algorithms like those in stdexec, I think it'd be worth having a build that can support Relacy tests (e.g., replacing `std::lock_guard lock(__mutex_)` with `STDEXEC_LOCK_GUARD(lock, __mutex_)` that conditionally expands to the Relacy version of the atomic/mutex/lock guards etc). If we do it right, the tests could verify a full search path of all possible thread interleavings (depending on the complexity of the unit test and algorithm), or at the very least, trigger a large sample of randomly chosen interleavings. Let me know what you think and I can work on a polished PR.

### Modifications to stdexec

~For example, `cv.notify_one()` is replaced with `cv($).notify_one()`, where `$` is a macro that expands to an object containing file/line information for Relacy's runtime. I wonder if a defaulted `std::source_location` could actually replace the need for much of the source code changes to stdexec?~

I was able to modify Relacy in https://github.com/dvyukov/relacy/pull/20 to minimize the number of changes to stdexec to allow writing Relacy tests. The changes to stdexec are
 - Remove stop_token's asserts, since they are fired in the Relacy test depending on which thread interleavings are tried first, to allow Relacy' runtime to print the interleaving history when it detects the use after free.
 - Replacing implicit `atomic::operator T()` and `int operator--()` with explicit calls to `load`/`fetch_sub`, which is able to leverage defaulting `std::source_location` to provide file/line info to the Relacy runtime.
 - Refactoring a cv notify (this is a limitation in Relacy which can easily be fixed in Relacy itself).

Note: Ideally, ThreadSanitizer (or a peer clang -fsanitize option) could do compiler based instrumentation. Relacy has not been in active development for a long time, with effort shifted towards TSAN. That said, Relacy still appears to offer a lot of value in being able to hit rare interleavings that TSAN itself does not support.

I think I found another bug with `async_scope` in `async_scope_bug`, which I'll provide more details (and try for a fix as well).

## Instructions

 > [!CAUTION]
Only verified to work on Linux based x86_64 with GCC-11.
Below gives me a segfault on Mac with Apple Clang (not sure why).

```
git clone -b stdexec https://github.com/ccotter/relacy && cd relacy
git clone -b relacy https://github.com/ccotter/stdexec
CXX_STD=c++20 make build/test/stdexec/{async_scope_bug,split_bug}
./build/test/stdexec/split_bug
```

### DIFF
--- a/test/rrd/Makefile
+++ b/test/rrd/Makefile
@@ -1,0 +1,59 @@
+# User-customizable variables:
+CXX ?= c++
+CXX_STD ?= c++20
+CXXFLAGS ?= -I relacy -I relacy/relacy/fakestd -O1 -std=$(CXX_STD) -I ../../include -I ../../test -g
+DEPFLAGS ?= -MD -MF $(@).d -MP -MT $(@)
+build_dir = build
+
+.SECONDARY:
+
+test_programs = split async_scope
+
+test_exe_files = $(foreach name,$(test_programs),$(build_dir)/$(name))
+
+exe_files = $(test_exe_files)
+o_files = $(exe_files:=.cpp.o)
+
+ansi_term_csi = [
+ansi_term_bold = $(ansi_term_csi)1m
+ansi_term_green = $(ansi_term_csi)32m
+ansi_term_red = $(ansi_term_csi)31m
+ansi_term_reset = $(ansi_term_csi)m
+
+COMPILE.cpp = $(CXX) $(DEPFLAGS) $(CXXFLAGS) -c
+LINK.cpp = $(CXX) $(CXXFLAGS)
+
+.PHONY: all
+all: tests
+
+.PHONY: tests
+tests: $(test_exe_files)
+
+$(build_dir)/%.check-result: $(build_dir)/% always-run
+	@ \
+	printf '%s%s ...%s\n' $(ansi_term_bold) $(*) $(ansi_term_reset) >&2; \
+	$(<); \
+	status="$${?}"; \
+	printf %d "$${status}" >$(@); \
+	if [ "$${status}" -eq 0 ]; then \
+		printf '%s%s %s%s\n' $(ansi_term_green) $(*) OK $(ansi_term_reset); \
+	else \
+		printf '%s%s %s%s\n' $(ansi_term_red) $(*) FAIL $(ansi_term_reset); \
+	fi >&2; \
+	exit "$${status}"
+
+$(build_dir)/%: $(build_dir)/%.cpp.o
+	$(LINK.cpp) $(^) -o $(@)
+
+$(build_dir)/%.cpp.o: %.cpp
+	@mkdir -p $(dir $(@))
+	$(COMPILE.cpp) -o $(@) $(<)
+
+.PHONY: clean
+clean:
+	rm -fr -- $(build_dir)/
+
+.PHONY: always-run
+always-run:
+
+-include $(o_files:=.d)
--- a/include/exec/__detail/__bwos_lifo_queue.hpp
+++ b/include/exec/__detail/__bwos_lifo_queue.hpp
@@ -152,7 +152,7 @@ namespace exec::bwos {
       block_type(block_size, allocator),
       allocator_of_t<block_type>(allocator))
     , mask_(blocks_.size() - 1) {
-    blocks_[owner_block_].reclaim();
+    blocks_[owner_block_.load()].reclaim();
   }
 
   template <class Tp, class Allocator>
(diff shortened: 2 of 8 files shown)


include/stdexec/stop_token.hpp
```diff
@@ -26,9 +26,9 @@
 #include <thread>
 #include <concepts>
 
-#if __has_include(<stop_token>) && __cpp_lib_jthread >= 201911
-#  include <stop_token>
-#endif
+//#if __has_include(<stop_token>) && __cpp_lib_jthread >= 201911
+//#  include <stop_token>
+//#endif
```
ericniebler, line 31: why did this get commented out?

ericniebler (approved):
hey chris! thanks for this, and sorry for the delay. relacy sounds very cool. i would love to have more confidence in the thread-safety of stdexec.

include/stdexec/stop_token.hpp
ccotter: No longer needed with https://github.com/dvyukov/relacy/pull/23

Relacy replaces standard includes at compile time, but `stop_token` is new compared to relacy.

ccotter:
I commented out (but reverted the change) the `STDEXEC_ASSERT`s in `inplace_stop_source`'s destructor, since the asserts were triggered and prevented relacy from showing the thread interleaving and memory access history log output. With the assertions enabled, the `split_bug` test reliably crashed quickly:

```
split_bug: stdexec/include/stdexec/__detail/../stop_token.hpp:244: stdexec::inplace_stop_source::~inplace_stop_source(): Assertion `(__state_.load(std::mo_relaxed, ::rl::debug_info(__FUNCTION__, "stdexec/include/stdexec/__detail/../stop_token.hpp", 244)) & __locked_flag_) == 0' failed.
```

test/rrd/Makefile
```diff
@@ -0,0 +1,59 @@
+# User-customizable variables:
```
ccotter, line 1: Ideally, this would be cmake along with a clone of the Relacy source code in one step. Leaving this here for now, as I haven't chosen to integrate the build into CI (some rationale in the README).

ccotter:
@ericniebler if you're good with it, this is good for review/merge. The tests here aren't ready to be integrated into CI yet, but I think the two tests I've added (both of which expose races) can be used as a model to add additional test to verify correctness.

PR #1403, opened 2024-08-26
[Enhancement] generalize tbb's thread_pool_base

DavidElesTheFox:
### Why?
I would like to use the library with another thread pool than tbb. This PR contains a proposed implementation for this enhancement and solves some issues.

### Current Issues
For this purpose the `tbbexec::thread_pool_base` looks like a perfect abstraction. But currently it has two issues:
 - It is in the `tbb_thread_pool.hpp`, therefore it has a dependency with tbb.
 - It has a friend declaration `friend struct DerivedPoolType::tbb_thread_pool::scheduler;` where the `tbb_thread_pool` actually bounding it to the tbb implementation and it is also unnecessary. I.e.: `friend struct DerivedPoolType::scheduler;` would be also good.

### Proposal
My proposal to having a more generic pool hierarchy in the stdexec. Currently it looks like:
```
 tbbexec
    | - tbb_thread_pool.hpp
```
I would go with adding 3 implementation for 3rd party thread pools (3 because I like the [Rule Of Three](https://en.wikipedia.org/wiki/Rule_of_three_(computer_programming))):
```
 execpools
  | - boost
  |    | - boost_thread_pool.hpp
  | - taskflow
  |    | - taskflow_thread_pool.hpp
  | - tbb
  |    | - tbb_thread_pool.hpp
  | - thread_pool_base.hpp
```
I choose these 3 3rd-party library, because I think these are the most popular threading libraries.

### About the PR
 - I've tried to organize the commit properly, making easier to follow the changes.
 - I used the `llvm18-cuda12.4` dev-container to test the changes.
 - I formatted the changed files with clang-format
 - Added two new cmake option: `STDEXEC_ENABLE_BOOST`, `STDEXEC_ENABLE_TASK_FLOW`. These should control the dependencies and fetches those when it is necessary.


DavidElesTheFox:
I've rebased the PR and also implemented @mathisloge comment.

I have a bit of a feeling to using the adding the standalone ASIO adds a bit of a complexity to the whole architecture and I'm up for reverting if it causes maintenance issues. The source of the complexity:
 - Standalone Asio isn't based on CMake, thus importing it even with CPM is not so simple. (Needed to add a cmake script)
 - Because the asio implementation depends on the configuration a generated header file needed to be added (asio_config.h.in). Though the generated file is not part of the source tree It ads another thing what needs to be maintained.

If you have any better idea to integrate the standalone asio let me know.

ericniebler:
apologies for the delay. i'll get to this soon i hope.

include/execpools/thread_pool_base.hpp
```diff
@@ -0,0 +1,485 @@
...
+  //! and
+  //! * template<F> auto execute(F &&f) -> decltype(f())
+  //!
+  //! See https://spec.oneapi.io/versions/1.0-rev-3/elements/oneTBB/source/task_scheduler/task_arena/task_arena_cls.html
+  using namespace stdexec::tags;
+
+  template <class PoolType, class ReceiverId>
+  struct operation {
+    using Receiver = stdexec::__t<ReceiverId>;
+    struct __t;
```
BenFrantzDale, line 36: Aside: I keep seeing this pattern, along with `stdexec::__t<Foo>`. I see what it's doing, but it's not clear to me why, other than that `__t<Foo>` is nicer than `typename Foo::__t`. If someone could give an explanation, I'd gladly add a doxygen comment to
```cpp
  template <class _Tp>
  using __t = typename _Tp::__t;
```
in `__meta.hpp`.

include/execpools/thread_pool_base.hpp
```diff
@@ -0,0 +1,485 @@
...
+        return sender{*pool_};
+      }
+    };
+
+    [[nodiscard]]
+    auto get_scheduler() noexcept -> scheduler {
+      return scheduler{static_cast<DerivedPoolType&>(*this)};
+    }
+
+    /* Is this even needed? Looks like no.
```
BenFrantzDale, line 428: I left this comment in here I think... I didn't get an answer to the question and forgot to remove it.

include/execpools/thread_pool_base.hpp
```diff
@@ -0,0 +1,485 @@
...
+    friend class thread_pool_base<PoolType>;
+
+    PoolType& pool_;
+    Receiver rcvr_;
+
+    explicit __t(PoolType& pool, Receiver rcvr)
+      : pool_(pool)
+      , rcvr_(std::move(rcvr)) {
+      this->__execute =
+        [](task_base* t, std::uint32_t /* tid What is this needed for? */) noexcept {
```
BenFrantzDale, line 466: I never got an answer to what the uint32 is for.

include/execpools/tbb/tbb_thread_pool.hpp
```diff
@@ -0,0 +1,58 @@
...
+    //! Constructor forwards to tbb::task_arena constructor:
+    template <class... Args>
+      requires stdexec::constructible_from<tbb::task_arena, Args...>
+    explicit tbb_thread_pool(Args&&... args)
+      : arena_{std::forward<Args>(args)...} {
+      arena_.initialize();
+    }
+
+    [[nodiscard]]
+    auto available_parallelism() const -> std::uint32_t {
```
BenFrantzDale, line 38: This isn't part of P2300 is it? It's just a nice-to-have, I think?

include/execpools/tbb/tbb_thread_pool.hpp
```diff
@@ -0,0 +1,58 @@
...
+    }
+
+    friend thread_pool_base<tbb_thread_pool>;
+
+    template <class PoolType, class ReceiverId>
+    friend struct operation;
+
+    void enqueue(task_base* task, std::uint32_t tid = 0) noexcept {
+      arena_.enqueue([task, tid] { task->__execute(task, /*tid=*/tid); });
+    }
```
BenFrantzDale, line 54: When I was writing this, it wasn't clear to me where the end of the P2300 concepts are and where things are specialized. I think (?) this `pool.enqueue(task)` isn't part of P2300, and is instead an implementation detail of the operation state for the sender for the pool's scheduler.

BenFrantzDale (commented):
I love this! When I wrote the tbb thread pool (with the help of Eric and others), I factored it like I did with exactly this in mind: That fundamentally an execution context should just need an `enqueue` member function and that for the most part everything else will be shared across most thread-pool-like resources.

I left a few comments about my old questions when I wrote it that are still nagging me.

ericniebler (changes requested):
we need to leave a forwarding `tbbexec/tbb_thread_pool.hpp` header so we don't break people's builds. also, we need to consider that folks are currently linking to the `STDEXEC::tbbexec` target. that should continue working.

also, can `thread_pool_base.hpp` be used to simplify `exec/static_thread_pool.hpp`?

include/execpools/thread_pool_base.hpp
```diff
@@ -0,0 +1,485 @@
...
+
+    void enqueue() noexcept {
+      pool_.enqueue(this);
+    }
+
+    void start() & noexcept {
+      enqueue();
+    }
+  };
+} // namespace execpools
```
BenFrantzDale, line 485: A good test/useage-example of this would be like an `AnyThreadPool` where tasks are submitted as `std::function<void()>`s.

include/execpools/thread_pool_base.hpp
ericniebler: > other than that `__t<Foo>` is nicer than `typename Foo::__t`.

that's it, really. i don't like littering my code with `typename` everywhere. not a hill to die on.

BenFrantzDale: Makes sense. With that in mind, why is it seemingly everywhere? Like
```cpp
    class static_thread_pool_ {
      template <class ReceiverId>
      struct operation {
        using Receiver = stdexec::__t<ReceiverId>;
        class __t;
      };
```
That's saying that `static_thread_pool_`'s `operation<ReceiverId>` struct has a `Receiver`, which it gets from the `ReceiverId`, and then has a `__t` class which is the operation state? That gets used later to do `template <class Receiver> using operation_t = stdexec::__t<operation<stdexec::__id<Receiver>>>;` Why that indirection there? (If you can explain it, I'll add `//!` to relevant places!) To me it feels like a round-about way of saying that `operation_t<Receiver>` is the operation-state type: You get the `__id` of the receiver, get `operation<that>` which basically says `using Receiver = __t<__id<Receiver>>;`...

I'm sure the indirection is there for a reason, I just don't see it. Is there some more-universal pattern going on that this is just a simple case of?

include/execpools/thread_pool_base.hpp
```diff
@@ -0,0 +1,485 @@
...
+
+namespace execpools {
+  //! This is a P2300-style thread pool wrapping tbb::task_arena, which its docs describe as "A class that represents an
+  //! explicit, user-managed task scheduler arena."
+  //! Once set up, a tbb::task_arena has
+  //! * template<F> void enqueue(F &&f)
+  //! and
+  //! * template<F> auto execute(F &&f) -> decltype(f())
+  //!
+  //! See https://spec.oneapi.io/versions/1.0-rev-3/elements/oneTBB/source/task_scheduler/task_arena/task_arena_cls.html
```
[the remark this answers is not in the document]
DavidElesTheFox: I have to admit it was one of the copy paste errors what I made unfortunately.
Correcting all occurrences.

include/execpools/thread_pool_base.hpp
DavidElesTheFox: If I don't misunderstand the question - so feel free to correct me if I do  - which is why the indirection is made with the `__t` class and what is the reason behind this indirection. Then for me this [doc](https://github.com/NVIDIA/stdexec/blob/519f931ef9d44ca6cc8b2c8a3b3cb760a3523d4f/MAINTAINERS.md#class-template-parameters) pretty much explains it.
Briefly as I understand: It makes the class type less tight coupled from its template arguments (I.e.: It depends on the Id as argument not on the specific type)

Is it correct @ericniebler? Please correct me when it's wrong.

On the other hand I'm curious whether c++-modules will make changes on it one day or not, but it is just a personal curiosity.

include/execpools/thread_pool_base.hpp
DavidElesTheFox: When I checked the code I was also wondering. For me it looks like used during bulk tasks this is used to propagate/report the error in the proper into the proper thread. Thus, I thought I'm keeping this comment here, because I'm not 100% sure.

include/execpools/tbb/tbb_thread_pool.hpp
DavidElesTheFox: I couldn't find any reference about it. But from my personal point of view it is nice to have but also I can imagine that not 'all' of the threading libraries has this information or easy to access. I.e.: In asio_thread_pool the `_executor` member is required due to fulfill this functional requirement.)
Thus, if we removing it then the thread pool will be more generic.

include/execpools/tbb/tbb_thread_pool.hpp
DavidElesTheFox: I fully agree. I mean I also don't now and I also have the same opinion. Should I add a documentation about it somewhere to the [structural description](https://github.com/NVIDIA/stdexec?tab=readme-ov-file#structure)?

include/execpools/thread_pool_base.hpp
DavidElesTheFox: I agree, more example makes it more usable! I just would like to questioning back, to clarify, if you don't mind.

The idea is to have an `AnyThreadPool` that based on the `thread_pool_base` but doesn't have a `get_scheduler()` public function to schedule tasks on it, but would have an `enqueue(std::function<void()> callable)` function how one can submit task to the pool. Is that correct?

DavidElesTheFox:
I've pushed some quick update on it:
 - removing the copy-paste errors on the base thread pool (tbb references, dead code)
 - `on` and `transform` got deprecated, I replaced these function usage on the tests
 - adding back the `tbbexec` header. @ericniebler I've also added a warning to the file. My idea is that everyone who uses the old header got a notification about the changes with this warning. I mean I like to be backward compatible but it is also nice if there is a way to remove this later on and doesn't leave in the repo forwever.
 
@ericniebler I think I can see the way how to use the same `scheduler` and `sender` in static_thread_pool and the other pools. So, these can have a common base class. I'm a bit worried about the details what I can't see right now (the possible tiny differences between static- and tbb/base- thread_pool, but theoretically it should work. I'm working on it to make this happen really soon.
 
 @BenFrantzDale I'm gonna add as an example the `AnyThreadPool` I like the idea of having more example for the usage!

include/execpools/thread_pool_base.hpp
BenFrantzDale: Aah. If there are no objections I’ll add a comment to `__t` with breadcrumbs pointing to that doc. Thank you!

include/execpools/thread_pool_base.hpp
BenFrantzDale: Something like that.  When I wrote this originally I was coming at it with the notion that a thread-based execution context must be modeled by something g you throw `std::function<void()>`s at and that it promises to execute eventually. (Or it could be `void(*f)(void*)`s and corresponding `void* x` as a pair and it promises to have a thread call `f(x)`. That’s weaker and I think still fits.)

ericniebler:
>  think I can see the way how to use the same scheduler and sender in static_thread_pool and the other pools. [...] I'm working on it to make this happen really soon.

no rush, that can be a separate PR.

ericniebler (approved):

ericniebler:
thank you!

Issue #1389, opened 2024-08-01
Compile failures on gcc-14 on Ubuntu Noble


williamspatrick:
To recreate this:

```
docker run --rm -it ubuntu:noble
apt update
apt-get install g++-14 cmake git
update-alternatives --install /usr/bin/gcc gcc /usr/bin/gcc-14 14 \
  --slave /usr/bin/g++ g++ /usr/bin/g++-14 \
  --slave /usr/bin/gcov gcov /usr/bin/gcov-14 \
  --slave /usr/bin/gcov-dump gcov-dump /usr/bin/gcov-dump-14 \
  --slave /usr/bin/gcov-tool gcov-tool /usr/bin/gcov-tool-14
cd ~
git clone https://github.com/NVIDIA/stdexec && cd stdexec
cmake -S . -B build
cmake --build build -j1
```

Failure:
```
root@2660c3013321:~/stdexec# cmake --build build -j1
[  1%] Building CXX object CMakeFiles/system_context.dir/src/system_context/system_context.cpp.o
In file included from /root/stdexec/include/stdexec/__detail/__schedule_from.hpp:27,
                 from /root/stdexec/include/stdexec/__detail/__continue_on.hpp:25,
                 from /root/stdexec/include/stdexec/execution.hpp:26,
                 from /root/stdexec/include/exec/__detail/__system_context_default_impl.hpp:19,
                 from /root/stdexec/include/exec/__detail/__system_context_default_impl_entry.hpp:24,
                 from /root/stdexec/src/system_context/system_context.cpp:18:
/root/stdexec/include/stdexec/__detail/__schedulers.hpp: In instantiation of 'auto stdexec::__sched::schedule_t::operator()(_Scheduler&&) const [with _Scheduler = stdexec::__inln::__scheduler]':
/root/stdexec/include/stdexec/__detail/__schedulers.hpp:66:17:   required by substitution of 'template<class _Scheduler>  requires  scheduler<_Scheduler> using stdexec::schedule_result_t = stdexec::__call_result_t<stdexec::__sched::schedule_t, _Scheduler> [with _Scheduler = stdexec::__inln::__scheduler]'
   66 |       { schedule(static_cast<_Scheduler&&>(__sched)) } -> sender;
      |         ~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/root/stdexec/include/stdexec/__detail/__inline_scheduler.hpp:78:65:   required from here
[...]

PR #1405, opened 2024-08-29
Replace default lambda expression with explicitly lambda

emmett2020:
Try to fix this issue: https://github.com/NVIDIA/stdexec/issues/1389
But I'm not sure if using lambda as a template parameter has any other purpose.


ericniebler:
> But I'm not sure if using lambda as a template parameter has any other purpose.

It does have another purpose. When defined as a template parameter, the mangled name is much shorter. See the following:

image

https://godbolt.org/z/nnxoW9cbP

I haven't measured, but I would be very surprised if your proposed change didn't regress compile time performance.

ericniebler:
closing for now since #1389 no longer seems to repro.

PR #1421, opened 2024-10-08
I noticed P2300R10 renames get_delegatee_scheduler to get_delegation_schduler

BenFrantzDale:
This does that rename leaving deprecated names.


ericniebler:
thanks, ben!

PR #1422, opened 2024-10-09
Add maintainers.md breadcrumbs, and other similar doxy comments

BenFrantzDale:
LMK if this is too much or the style is wrong. These are the breadcrumbs I wish I'd had :-D

### DIFF
--- a/include/stdexec/__detail/__basic_sender.hpp
+++ b/include/stdexec/__detail/__basic_sender.hpp
@@ -500,8 +500,8 @@ namespace stdexec {
   using __get_attrs_fn =
     __result_of<__detail::__drop_front, __mtypeof<__sexpr_impl<_Tag>::get_attrs>>;
 
-  //////////////////////////////////////////////////////////////////////////////////////////////////
-  // __basic_sender
+  //! A dummy type used only for diagnostic purposes. 
+  //! See `__sexpr` for the implementation of P2300's _`basic-sender`_.
   template <class...>
   struct __basic_sender {
     // See MAINTAINERS.md#class-template-parameters for `__id` and `__t`.
@@ -510,7 +510,9 @@ namespace stdexec {
   };
 
   //! A struct template to aid in creating senders.
-  //! A `__sexpr<Desc>` 
+  //! This struct closely resembles P2300's [_`basic-sender`_](https://eel.is/c++draft/exec#snd.expos-24),
+  //! but is not an exact implementation.
+  //! Note: The struct named `__basic_sender` is just a dummy type and is also not _`basic-sender`_.
   template <auto _DescriptorFn, class = __anon>
   struct __sexpr {
     using sender_concept = sender_t;
--- a/include/stdexec/__detail/__basic_sender.hpp
+++ b/include/stdexec/__detail/__basic_sender.hpp
@@ -504,14 +504,18 @@ namespace stdexec {
   // __basic_sender
   template <class...>
   struct __basic_sender {
+    // See MAINTAINERS.md#class-template-parameters for `__id` and `__t`.
     using __id = __basic_sender;
     using __t = __basic_sender;
   };
 
+  //! A struct template to aid in creating senders.
+  //! A `__sexpr<Desc>` 
   template <auto _DescriptorFn, class = __anon>
   struct __sexpr {
     using sender_concept = sender_t;
 
+    // See MAINTAINERS.md#class-template-parameters for `__id` and `__t`.
     using __id = __sexpr;
     using __t = __sexpr;
     using __desc_t = decltype(_DescriptorFn());
@@ -595,6 +599,9 @@ namespace stdexec {
 
   //////////////////////////////////////////////////////////////////////////////////////////////////
   // __make_sexpr
+  //! A tagged function-object
+  //! Takes data and children and 
+  //! returns `__sexpr_t<_Tag, _Data, _Child...>{_Tag(), data, children...}`.
   namespace __detail {
     template <class _Tag>
     struct __make_sexpr_t {
--- a/include/stdexec/__detail/__meta.hpp
+++ b/include/stdexec/__detail/__meta.hpp
@@ -28,6 +28,9 @@
 #include "__utility.hpp"
 
 namespace stdexec {
+  //! Convenience metafunction getting the dependant type `__t` out of `_Tp`.
+  //! That is, `typename _Tp::__t`.
+  //! See MAINTAINERS.md#class-template-parameters for details.
   template <class _Tp>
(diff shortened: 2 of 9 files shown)


include/stdexec/__detail/__basic_sender.hpp
```diff
@@ -595,6 +599,9 @@ namespace stdexec {
 
   //////////////////////////////////////////////////////////////////////////////////////////////////
   // __make_sexpr
+  //! A tagged function-object
+  //! Takes data and children and 
+  //! returns `__sexpr_t<_Tag, _Data, _Child...>{_Tag(), data, children...}`.
```
BenFrantzDale, line 604: I didn't quite grok this `sexpr` stuff... Are there docs?

include/stdexec/__detail/__meta.hpp
```diff
@@ -410,6 +421,7 @@ namespace stdexec {
     using __f = _Fn;
   };
 
+  //! Metafunction takes a function and provides a checked version of it?
```
BenFrantzDale, line 424: The `?` is because I didn't grok this stuff... what's going on with all of these one-letter things?

ericniebler: `__q` is like [`mpl::quote`](https://www.boost.org/doc/libs/1_86_0/libs/mpl/doc/refmanual/quote.html). It turns an alias/class template into a metafunction.

it also propagates "meta-exceptions". all of the meta utilities recognize specializations of `stdexec::_ERROR_` as an error type. Error types short-circuit the evaluation of the metafunction and are automatically propagated, like an exception.

this lets me report type errors briefly at the library boundary, even if the actual error happens deep inside a meta-program.

ericniebler (approved):

include/stdexec/__detail/__basic_sender.hpp
BenFrantzDale: Regarding "docs", personally, I find comments like the ones I'm trying to add in this PR to be more useful than "docs", since go-to-definition lands on them directly. I'll try to paraphrase what you are saying here.

BenFrantzDale: Do you mean `basic-sender` is represented by `__sexpr`? Because it doesn't seem to match... Plus there's `__basic_sender` defined on lines I touched above. Which is which?

include/stdexec/__detail/__meta.hpp
```diff
@@ -321,6 +326,9 @@ namespace stdexec {
   template <class... _Args>
   concept _Ok = (STDEXEC_IS_SAME(__ok_t<_Args>, __msuccess) && ...);
 
+  //! If both are true:
+  //! Then __i<true, true>::__g<F, Args...> is an alias for F<Args...>
+  //! and __i<true, true>::__f<F> is an alias for F.
```
BenFrantzDale, line 331: I tried to explain what I saw it doing, but I didn't grok it. Is there a high-level description of what `__i` does?

include/stdexec/__detail/__basic_sender.hpp
ericniebler: `stdexec::__basic_sender` is a dummy type used only for diagnostic purposes. `__sexpr` is the implementation of P2300's _`basic-sender`_.

BenFrantzDale: Ok. Want me to rename things or in this PR just document that fact?

include/stdexec/__detail/__meta.hpp
ericniebler: it is the implementation of P2300's [_`META-APPLY`_](https://eel.is/c++draft/exec#util.cmplsig-5). 

> [Note [1](https://eel.is/c++draft/exec#util.cmplsig-note-1): The purpose of META-APPLY is to make it valid to use non-variadic templates as Variant and Tuple arguments to gather-signatures. — end note]

in addition to avoiding the dreaded "pack expanded into non-pack argument" error, it is part of the meta-error propagation mechanism. if any of the argument types are a specialization of `_ERROR_`, `__i` will short-circuit and return the error. `__minvoke` and `__meval` are implemented in terms of `__i`.

include/stdexec/__detail/__meta.hpp
```diff
@@ -844,7 +872,14 @@ namespace stdexec {
...
     using __f = _Id<_Ty>;
   };
+<<<<<<< HEAD
 
+=======
+  //! Metafunction mapping `_Ty` to either
+  //! * `typename _Ty::__id` if that exists, or to 
+  //! * `_Ty` (itself) otherwise.
+  //! See MAINTAINERS.md#class-template-parameters.
+>>>>>>> adcee031 (Add doxy comments to some of the metafunctions with breadcrumbs back to)
```
ericniebler, line 882: oops!

ericniebler (changes requested):
you accidentally committed a git diff marker.

include/stdexec/__detail/__meta.hpp
BenFrantzDale: 🤦

include/stdexec/__detail/__basic_sender.hpp
BenFrantzDale: Also, _`basic-sender`_ is
```cpp
  template<class Tag, class Data, class... Child>
  struct basic-sender : product-type<Tag, Data, Child...> {    // exposition only
    using sender_concept = sender_t;
```
but `__sexpr` is
```cpp
  template <auto _DescriptorFn, class = __anon>
  struct __sexpr {
    using sender_concept = sender_t;
```
so something doesn't make sense... I assume you are being imprecise and that something related to `__sexpr` corresponds to _`basic-sender`_?

ericniebler: i didn't mean to imply that `__sexpr` is exact clone of _`basic-sender`_. it does some things differently, but it largely follows the design of _`basic-sender`_.

no need to rename things. i like them as they are.

BenFrantzDale: Ok, if it isn’t quite _`basic-sender`_ I’ll say it’s similar to it. But I did comment that `__basic_sender` is not in fact providing _`basic-sender`_ which seems important to explicitly state.

ericniebler (approved):

ericniebler:
Thanks Ben! These comments are great.

BenFrantzDale:
> Thanks Ben! These comments are great.

Thank you for explaining things. There are other dark corners I haven’t grokked yet, but I’m very pleased with this start.

PR #1436, opened 2024-11-11
Improve EDG detection and workarounds

vasama:
* Add missing const in `sender_adaptor_closure`.
* Detect NVHPC as EDG first, add separate detection of NVHPC.
* Convert most NVHPC workaround conditions to check for EDG instead.
* Add detection of Visual Studio Intellisense (EDG).
* Add separate detection of the MSVC headers for Intellisense.
* Add workaround for Intellisense issue in `sender_adaptor_closure`.
* Add workarounds for EDG bugs that cropped up related to coroutines.
  It seems that currently the coroutine tests are not built in CI with EDG, because NVC++ is the only EDG based compiler and does not have coroutine support.

There is still an issue in `test_task.cpp` that I haven't been able to figure out. Once that's sorted as well, it would be nice to have another EDG-based compiler in CI, especially one with coroutine support.

### DIFF
--- a/include/exec/__detail/intrusive_heap.hpp
+++ b/include/exec/__detail/intrusive_heap.hpp
@@ -54,7 +54,7 @@ namespace exec {
 #  endif
 #endif
 
-  template <auto Key, auto Prev, auto Left, auto Right>
+  template <class Node, class KeyT, auto Key, auto Prev, auto Left, auto Right>
   class intrusive_heap;
 
   template <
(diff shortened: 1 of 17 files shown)


ericniebler:
thanks @vasama !

ericniebler:
Do you know if the pack indexing problem is my bug or clang's?

vasama:
I reported the suspected Clang bug here: https://github.com/llvm/llvm-project/issues/116105

PR #1444, opened 2024-12-09
Fix MSVC workaround in stdexec/coroutine.hpp

### LINKED ISSUE #1435: #if STDEXEC_MSVC() && _MSC_VER >= 1939 should read _MSC_VER < 1939
Just from reading the code and the comment following this line:

https://github.com/NVIDIA/stdexec/blob/main/include/stdexec/coroutine.hpp#L21

vasama:
Fixes #1435.

### DIFF
--- a/include/stdexec/coroutine.hpp
+++ b/include/stdexec/coroutine.hpp
@@ -18,7 +18,7 @@
 #include "__detail/__awaitable.hpp"
 #include "__detail/__config.hpp"
 
-#if STDEXEC_MSVC() && _MSC_VER >= 1939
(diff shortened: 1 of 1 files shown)


include/stdexec/coroutine.hpp
```diff
@@ -18,7 +18,7 @@
 #include "__detail/__awaitable.hpp"
 #include "__detail/__config.hpp"
 
-#if STDEXEC_MSVC() && _MSC_VER >= 1939
```
ericniebler, line 21: :facepalm:

ericniebler:
thanks!

PR #1445, opened 2024-12-11
Remove relative includes across components

vasama:
Includes of the form `"../stdexec/execution.hpp"` are converted to `"stdexec/execution.hpp"`. This allows packaging the component headers separately.

If the headers are packaged separately, the packages break due to the relative includes. Personally I think relative includes should always be avoided, but in this PR I only changed those that go across components (`stdexec`, `exec`, `nvexec`, ...).

### DIFF
--- a/include/exec/__detail/__atomic_intrusive_queue.hpp
+++ b/include/exec/__detail/__atomic_intrusive_queue.hpp
@@ -16,7 +16,7 @@
  */
 #pragma once
 
-#include "../../stdexec/__detail/__intrusive_queue.hpp"
+#include "stdexec/__detail/__intrusive_queue.hpp"
 
 #include <atomic>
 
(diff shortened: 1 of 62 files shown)


ericniebler:
i seem to recall that the HPC SDK wanted/needed stdexec to use relative includes, but i'm not 100% sure. @dkolsen-pgi, am i misremembering?

dkolsen-pgi:
> I seem to recall that the HPC SDK wanted/needed stdexec to use relative includes, but i'm not 100% sure.

You are correct.  In the HPC SDK the include paths with `nvc++ -stdpar` are set up by default so that all the includes need to contain `experimental`, as in `<experimental/stdexec/execution.hpp>`.  Without the `experimental` the includes won't work.  This change will break the HPC SDK.

We need to make a plan for transitioning stdexec as packaged in the HPC SDK from experimental to C++26 standard conforming (plus extensions).  Getting rid of the cross-component relative includes can be part of that plan.  But eliminating the relative includes can't happen yet without the HPC SDK locking down on an older version of stdexec.

PR #1465, opened 2025-01-07
Bug:  static_thread_pool::get_scheduler_on_thread causes SegFault

DavidElesTheFox:
### Issue
`nodemask_` is not initialized when `get_scheduler_on_thread` is called on a `static_thread_pool`

### Details

A new test is added. I haven't find existing tests for the `static_thread_pool`, thought starting one can be handy for later.

The test case creates a new pool and try to start tasks on different thread indexes. It fails with segmentation fault without the fix because the member pointer is not initialized.

With the fix we can observe that the pool scheduled tasks on different threads.

### DIFF
--- a/include/exec/static_thread_pool.hpp
+++ b/include/exec/static_thread_pool.hpp
@@ -407,7 +407,7 @@ namespace exec {
 
         static_thread_pool_* pool_;
         remote_queue* queue_;
-        const nodemask* nodemask_;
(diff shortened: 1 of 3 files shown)


ericniebler:
thank you @DavidElesTheFox !

DavidElesTheFox:
Thank you :), it's always a pleasure to collaborate! :)

PR #1466, opened 2025-01-08
Bug: Fix build issues during fetch_content usage

DavidElesTheFox:
### Context

When stdexec is used as a 3rd-party dependency and imported with fetch_content it is downloaded into the cmake build folder under the _deps folder.

A setup like that will results a build directory structure like 
```
build/_deps/stdexec-src
           /stdexec-build
```
in this context the `CMAKE_BINARY_DIR` is the `build` folder and the `CMAKE_CURRENT_BINARY_DIR` is the `build/_deps/stdexec-build`

### Issue 1:

Right now the build will place two files into the 'main' build folder:
 - `execution.bs`
 - `RAPIDS.cmake`

It might causes problems when the main cmake project uses RAPIDS with a different version.

### Issue 2:

There is no such a file `stdexec_version_config.hpp` under the main build tree. And it will cause a configuration error. So, the library can't be used. 

### Fix

Using `CMAKE_CURRENT_BINARY_DIR` instead of `CMAKE_BINARY_DIR`.
Consequences: 
 - When stdexec is built 'stand alone' the two variables store the same value. Thus, it will not cause any changes.
 - When stdexec is imported then the build of the library will stay in its local scope.


ericniebler:
thanks!

Issue #1475, opened 2025-02-14
[Question] Is it intentional that static_thread_pool is basically spinlocking when idle?


miwechner:
Hi, 

Thanks for your work, I am really enoying working with S&R.

I wanted to use `exec::static_thread_pool` in a GUI application and noticed that I have very high CPU usage when I create an exec::static_thread_pool without enqueuing anything. I think it comes down to this function which means one stealer thread is always kept alive and attempts to pop the queues, even though nothing has been enqueued.

https://github.com/NVIDIA/stdexec/blob/b888185d667f68b9a8bda5d0c81d03edf9ec3fe1/include/exec/static_thread_pool.hpp#L930-L934

I guess this is intentional to reduce latency, but when i started i have to say that this behaviour wasn't clear to me. I also could not find any documentation for static_thread_pool or issue here on Github talking about this. So i am asking just in case for me and for anyone else that would attempt the same.

Thanks,

Michael

maikel:
Lets say its at least a known issue. It was discussed in issue #1305 and your question is another motivational push to change that.

maikel:
Hi Michael,

I made a PR where the pool only wakes up new thiefs if at least one victim is awake. If it is the last thread it will be put to sleep.

Can you try that, whether it fits better?

Thank you for reporting the issue!

PR #1476, opened 2025-02-15
Make static thread pool sleep when empty

maikel:
Addresses #1475

### DIFF
--- a/include/exec/static_thread_pool.hpp
+++ b/include/exec/static_thread_pool.hpp
@@ -604,6 +604,8 @@ namespace exec {
         void notify_one_sleeping();
         void set_stealing();
         void clear_stealing();
+        void set_sleeping();
+        void clear_sleeping();
 
         bwos::lifo_queue<task_base*, numa_allocator<task_base*>> local_queue_;
         __intrusive_queue<&task_base::next> pending_queue_{};
@@ -620,7 +622,7 @@ namespace exec {
       void run(std::uint32_t index) noexcept;
       void join() noexcept;
 
-      alignas(64) std::atomic<std::uint32_t> numThiefs_{};
+      alignas(64) std::atomic<std::uint32_t> numActive_{};
       alignas(64) remote_queue_list remotes_;
       std::uint32_t threadCount_;
       std::uint32_t maxSteals_{threadCount_ + 1};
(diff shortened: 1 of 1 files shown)


include/exec/static_thread_pool.hpp
```diff
@@ -923,12 +926,30 @@ namespace exec {
...
 
+    inline void static_thread_pool_::thread_state::set_sleeping() {
+      pool_->numActive_.fetch_sub(1 << 16, std::memory_order_relaxed);
+    }
+
+    // wakeup a worker thread and maintain the invariant that we always one active thief as long as a potential victim is awake
+    inline void static_thread_pool_::thread_state::clear_sleeping() {
+      const std::uint32_t numActive = pool_->numActive_.fetch_add(1 << 16, std::memory_order_relaxed);
+      if (numActive == 0) {
+        notify_one_sleeping();
```
miwechner, line 937: Thanks for addressing this so quickly. I tried to run it on my Windows PC in debug mode project and got a crash on initialization here:

image

The `thread_states_` are still nullopt when this is constructed, but if i guard the empty optional access I no longer have any active threads just from constructing the thread pool.

maikel: @miwechner please try again :pray:

miwechner: @maikel works without problems now on my end, thank you!

ericniebler (approved):
thanks

include/exec/static_thread_pool.hpp
```diff
@@ -980,7 +1001,9 @@ namespace exec {
           if (result.task) {
             return result;
           }
+          set_sleeping();
           cv_.wait(lock);
+          clear_sleeping();
```
miwechner, line 1006: The thread sanitizer I think correctly identified a potential deadlock here as `clear_sleeping` calls `notify`, which in turn has to acquire the `_mut` lock of a different thread:

@maikel, As far as I see there is no reason to hold this lock when calling clear_sleeping so this should fix it:
```
diff --git a/include/exec/static_thread_pool.hpp b/include/exec/static_thread_pool.hpp
index e41d946e..d3c79c25 100644
--- a/include/exec/static_thread_pool.hpp
+++ b/include/exec/static_thread_pool.hpp
@@ -1003,9 +1003,12 @@ namespace exec {
           }
           set_sleeping();
           cv_.wait(lock);
+          lock.unlock();
           clear_sleeping();
         }
-        lock.unlock();
+        if(lock.owns_lock()){
+          lock.unlock();
+        }
         state_.store(state::running, std::memory_order_relaxed);
         result = try_pop();
       }
```

PR #1485, opened 2025-03-03
Parallel scheduler -- incorporate feedback from LEWG in Hagenberg

lucteo:
Incorporate the feedback from LEWG in Hagenberg, except the changes related to bulk (P3481), which will come as a later PR.

Mainly renames; minor changes to the API.


lucteo:
@ericniebler : this is ready for review / CI check.

(after this, I'm planning another PR that makes changes to `bulk`, which also affects this code)

include/exec/system_context.hpp
```diff
@@ -82,17 +82,17 @@ namespace exec {
...
-  class system_sender;
+  class parallel_scheduler;
+  class __parallel_sender;
   template <stdexec::sender _S, std::integral _Size, class _Fn>
-  class system_bulk_sender;
+  class __parallel_bulk_sender;
 
   /// Returns a scheduler that can add work to the underlying execution context.
-  auto get_system_scheduler() -> system_scheduler;
+  auto get_parallel_scheduler() -> parallel_scheduler;
```
ericniebler, line 91: can you pls add back a (deprecated) `get_system_scheduler()` function?

include/exec/system_context.hpp
```diff
@@ -220,44 +221,48 @@ namespace exec {
...
     /// Gets the environment of this sender.
     [[nodiscard]]
-    auto get_env() const noexcept -> __detail::__system_scheduler_env {
+    auto get_env() const noexcept -> __detail::__parallel_scheduler_env {
       return {__scheduler_};
     }
 
+    /// Value completion happens on the parallel scheduler.
+    parallel_scheduler
+      query(stdexec::get_completion_scheduler_t<stdexec::set_value_t>) const noexcept;
```
ericniebler, line 236: i recently moved to trailing return types for all of stdexec. if you have it, `clang-tidy --fix --fix-errors <file>` will apply the changes for you.

ericniebler (changes requested):

include/exec/system_context.hpp
lucteo: Done. thank you.

include/exec/system_context.hpp
lucteo: I did apply clang-tidy to these source files, but it also changed some other files. I only submitted changes to the "system_context"-related source files.

ericniebler (approved):

ericniebler:
@lucteo stdexec is currently broken on gcc-14/15 because the new system context is trying to `constinit` the singleton, which has a `std::shared_ptr` in it. could you pls have a look at that?

lucteo:
@ericniebler : can you please retrigger the the GPU build in the hope that it will pass? (it seems that there is some problems there, as this is not the first time I see the CI failing in the same way)

ericniebler:
> can you please retrigger the the GPU build in the hope that it will pass? (it seems that there is some problems there, as this is not the first time I see the CI failing in the same way)

i've been retrying it. this is attempt # 15. it is caused by a known internal nvidia networking issue (the gpu runners are hosted by nvidia). it's been like this for weeks. the issue is being worked on. until then, i'll keep re-running the tests until they pass, or just force-land it.

PR #1500, opened 2025-03-24
P3481 R2 - Incorporate changes from the "bulk issues" paper

lucteo:
See http://wg21.link/P3481R2

Main changes:
- add `bulk_chunked` and `bulk_unchunked`
- `bulk` takes an additional execution policy
- `bulk` is specified in terms of `bulk_chunked`


lucteo:
@ericniebler : this is not yet ready, but I wanted to push it up, so that you have a chance to see it earlier.

I'm planning to change the schedulers (static_thread_pool, libdispatch, parallel_scheduler) to be customized in terms of `bulk_unchunked`.

I'm also expecting some issues on the CUDA sides; I'm not sure how can I move forward on that side. Any help is greatly appreciated.

lucteo:
The GCC failure should be fixed. I'm not sure how to approach the Windows failure :(

ericniebler:
@lucteo you can disregard the GPU CI failures. they are infra-related. the msvc failures are real though.

lucteo:
@ericniebler : there are some issues on the GPU side as well caused by my patch. I will temporarily pause this for 2 weeks, as I need to finish my ACCU presentation for next week.

ericniebler:
@lucteo i have addressed the GPU build problem.

ericniebler (approved):
s'thing must've been amiss about my suggested edits. tests are failing now. :-/

lucteo:
The error is strange. First, it works perfectly fine on my Mac; I was expecting the CI Mac build to be similar to my builds, but apparently it isn't.
Second, it seems that only the first declaration of `operator()` from the base class is used in the derived class, while the second is not. I find this strange (especially that it happens on 3 compilers)

Fixed by moving the deprecation in the base class. The deprecation method should also apply to `bulk_chunked` w/o policy, but there shouldn't be any code affected by this, so, we should be fine.

PR #1526, opened 2025-05-04
Add missing install.

jehelset:
`thread_pool_base.hpp` used f.ex. by the tbb thread-pool, but is not installed by any other target.


ericniebler (approved):

ericniebler:
thanks for your contributions, @jehelset. it's much appreciated.

Issue #1523, opened 2025-05-01
STDEXEC_HAS_FEATURE(__cpp_lib_execution) fails with recent libstdc++.


jehelset:
I think https://github.com/NVIDIA/stdexec/commit/0242ad9a5ac172324ff51803a1c79facca36b229 broke NVIDIA/stdexec for at least gcc-15 and gcc-trunk (https://godbolt.org/z/9ez66G7vv).

it checks `__has_feature(__cpp_lib_execution)`, but in libstdc++ is a macro defined at https://github.com/gcc-mirror/gcc/blob/master/libstdc%2B%2B-v3/include/bits/version.h#L672.

should perhaps guard with

```
#if defined(__cpp_lib_execution)
#  define STDEXEC_HAS_EXECUTION_POLICY() 0
#elif __has_feature(__cpp_lib_execution)
#  define STDEXEC_HAS_EXECUTION_POLICY() 1
#else
#  define STDEXEC_HAS_EXECUTION_POLICY() 0
#endif
```

@lucteo

lucteo:
Appologize, I missed this. I can take a look when I get some spare time (these days are rather busy for me)

ericniebler:
fixed by #1559

PR #1528, opened 2025-05-04
More robust feature-checking.

jehelset:
Suggestion for closing https://github.com/NVIDIA/stdexec/issues/1523.

### DIFF
--- a/include/stdexec/__detail/__config.hpp
+++ b/include/stdexec/__detail/__config.hpp
@@ -402,7 +402,9 @@ namespace stdexec {
 #  define STDEXEC_HAS_STD_MEMORY_RESOURCE() 0
 #endif
 
-#if STDEXEC_HAS_FEATURE(__cpp_lib_execution)
+#if defined(__cpp_lib_execution)
+#  define STDEXEC_HAS_EXECUTION_POLICY() 0
+#elif STDEXEC_HAS_FEATURE(__cpp_lib_execution)
 #  define STDEXEC_HAS_EXECUTION_POLICY() 1
 #else
 #  define STDEXEC_HAS_EXECUTION_POLICY() 0


prlw1:
Your patch seems to do the logical opposite of what I think you intend: if the feature exists, you don't use it.

jehelset:
> Your patch seems to do the logical opposite of what I think you intend: if the feature exists, you don't use it.

i don't think https://github.com/NVIDIA/stdexec/commit/0242ad9a5ac172324ff51803a1c79facca36b229 was aimed at adding support for this on libstdc++-based builds, so here, as a quick fix i just wanted to make sure these still are able to build (not add support for this, since then i'd have to fix the underlying build failure, and i didn't have time for that).

ericniebler:
this seems wrong to me. @lucteo thoughts?

mathisloge:
It would be great if we find some solution to this, as the main branch currently does not compile on gcc-15.

lucteo:
Sorry, I've missed this. These days are a bit busy for me, but I'll look as soon as I have some time.

include/stdexec/__detail/__config.hpp
```diff
@@ -418,7 +418,9 @@ namespace stdexec {
 #  define STDEXEC_HAS_STD_MEMORY_RESOURCE() 0
 #endif
 
-#if STDEXEC_HAS_FEATURE(__cpp_lib_execution)
+#if defined(__cpp_lib_execution)
+#  define STDEXEC_HAS_EXECUTION_POLICY() 0
+#elif STDEXEC_HAS_FEATURE(__cpp_lib_execution)
```
lucteo, line 423: Sorry, I've seen the issue before the fix :(
I did post a possible fix in #1559.

Besides the problem reported in the issue, we also need to check the version of the feature flag, as the execution policies were added in 2 steps.

PR #1530, opened 2025-05-06
Work around NVC++ bug with partial spec of var template

dkolsen-pgi:
The development branch of NVC++ currently has a regression where the compiler sometimes gets confused about the type of a partial specialization of a variable template with a deduced type.  Work around this bug by changing the type of `__v<__mconstant<_Np>>` from `auto` to the equivalent `decltype(_Np)`.

### DIFF
--- a/include/stdexec/__detail/__meta.hpp
+++ b/include/stdexec/__detail/__meta.hpp
@@ -121,8 +121,9 @@ namespace stdexec {
   template <class _Tp, _Tp _Ip>
   inline constexpr _Tp __v<std::integral_constant<_Tp, _Ip>> = _Ip;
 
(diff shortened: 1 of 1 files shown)


ericniebler (approved):

include/stdexec/__detail/__meta.hpp
```diff
@@ -121,8 +121,9 @@ namespace stdexec {
   template <class _Tp, _Tp _Ip>
   inline constexpr _Tp __v<std::integral_constant<_Tp, _Ip>> = _Ip;
 
+  // `decltype(_Np)` instead of `auto` to work around NVHPC/EDG bug.
```
ericniebler, line 124:
```suggestion
  // `__mtypeof<_Np>` instead of `auto` to work around NVHPC/EDG bug.
```

PR #1532, opened 2025-05-11
Customize `bulk_chunked` and `bulk_unchunked` for `parallel_scheduler`

lucteo:
Doing this, we will align with the intent of P2079R7.
- the backend supports `shedule_bulk_chunked` and `schedule_bulk_unchunked` instead of `bulk_schedule`
- update replaceability API to match P2079R7
- add chunking logic
- if execution policy of bulk_chunked doesn't require parallelization, call only one chunk
- add more tests

### DIFF
--- a/.github/workflows/ci.cpu.yml
+++ b/.github/workflows/ci.cpu.yml
@@ -6,6 +6,7 @@ on:
       - main
       - "member-function-customization"
       - "pull-request/[0-9]+"
+      - "parallel-scheduler-backend-api"
 
 concurrency:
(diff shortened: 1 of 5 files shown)


lucteo:
@ericniebler : If you have some spare time, I would appreciate a review on this PR.

lucteo:
@ericniebler : can you please give it another round? I fixed the data race in tests.

ericniebler (approved):

ericniebler:
thanks @lucteo !

PR #1545, opened 2025-06-07
Ignore callback parameter in never_stop_token::__callback_type ctor.

jehelset:


jehelset:
Not used, potentially save template instantiations.

PR #1547, opened 2025-06-11
sequence: relax iterate to support ranges::input_range

petertran858:
Relaxes requirement on iterate from `forward_range` to `input_range` in order to support single-pass ranges.


petertran858:
CC: @maikel

ericniebler:
`iterate` actually _does_ require a forward range because of this line:

```c++
stdexec::set_value(static_cast<_ItemRcvr&&>(__rcvr_), *__parent_->__iterator_++);
```

the expression `*it++` does the following

1. makes a copy of `it`
2. increments `it`
3. dereferences the copy

for an input iterator, step (2) invalidates the copy made in step (1), leading step (3) into undefined-behavior land.

PR #1551, opened 2025-06-18
Silencing clang-tidy warnings

sithhell:
- Silencing `readability-static-accessed-through-instance` by removing static from function declaration
 - Silencing "Assigned value is garbage or undefined":
   - Properly initialize vtable
   - Disabling some code when run under clang-tidy

### DIFF
--- a/include/exec/task.hpp
+++ b/include/exec/task.hpp
@@ -320,16 +320,16 @@ namespace exec {
 
      private:
       struct __final_awaitable {
-        static constexpr auto await_ready() noexcept -> bool {
+        constexpr auto await_ready() noexcept -> bool {
           return false;
         }
 
-        static auto await_suspend(__coro::coroutine_handle<__promise> __h) noexcept
+        auto await_suspend(__coro::coroutine_handle<__promise> __h) noexcept
           -> __coro::coroutine_handle<> {
           return __h.promise().continuation().handle();
         }
 
-        static void await_resume() noexcept {
+        void await_resume() noexcept {
         }
       };
 
@@ -369,6 +369,7 @@ namespace exec {
           this->__data_.template emplace<1>(std::current_exception());
         }
 
+#ifndef __clang_analyzer__
         template <sender _Awaitable>
           requires __scheduler_provider<_Context>
         auto await_transform(_Awaitable&& __awaitable) noexcept -> decltype(auto) {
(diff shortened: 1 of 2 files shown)


include/exec/task.hpp
```diff
@@ -320,16 +320,16 @@ namespace exec {
...
 
-        static auto await_suspend(__coro::coroutine_handle<__promise> __h) noexcept
+        auto await_suspend(__coro::coroutine_handle<__promise> __h) noexcept
           -> __coro::coroutine_handle<> {
           return __h.promise().continuation().handle();
         }
 
-        static void await_resume() noexcept {
+        void await_resume() noexcept {
         }
```
ericniebler: the code is correct. these functions do not need to be passed a `this` pointer, hence the `static`. is there a way to suppress this warning locally?

include/exec/task.hpp
```diff
@@ -320,16 +320,16 @@ namespace exec {
 
      private:
       struct __final_awaitable {
-        static constexpr auto await_ready() noexcept -> bool {
+        constexpr auto await_ready() noexcept -> bool {
```
ericniebler:
```suggestion
        static constexpr auto await_ready() noexcept -> bool {
```

include/exec/task.hpp
```diff
@@ -320,16 +320,16 @@ namespace exec {
...
           return false;
         }
 
-        static auto await_suspend(__coro::coroutine_handle<__promise> __h) noexcept
+        auto await_suspend(__coro::coroutine_handle<__promise> __h) noexcept
```
ericniebler:
```suggestion
        static auto await_suspend(__coro::coroutine_handle<__promise> __h) noexcept
```

include/exec/task.hpp
```diff
@@ -320,16 +320,16 @@ namespace exec {
...
           -> __coro::coroutine_handle<> {
           return __h.promise().continuation().handle();
         }
 
-        static void await_resume() noexcept {
+        void await_resume() noexcept {
```
ericniebler:
```suggestion
        static void await_resume() noexcept {
```

include/exec/task.hpp
sithhell: You are of course completely right. I'm actually on the fence of just disabling the "static readability" check. Thanks for merging the remainder!

PR #1560, opened 2025-06-25
Bulk unchunked

lucteo:
Update `bulk_unchunked` to the specification of P3481R5.
This will give `bulk_unchunked` an execution policy and allows it to run on non-concurrent shcedulers.

Also fix customization of parallel_scheduler when we don't have a receiver environment.


lucteo:
@ericniebler : can you please take a look at this PR too?

Issue #1534, opened 2025-05-13
Boost.Beast unable to use `asioexec::use_sender`


SidneyCogdill:
```cpp
#include <asioexec/use_sender.hpp>
#include <exec/task.hpp>

#include <boost/beast.hpp>

auto foo(boost::beast::tcp_stream stream) -> exec::task<void> {
  boost::beast::flat_buffer buffer;

  boost::beast::http::request<boost::beast::http::string_body> req;
  auto bytes_read = co_await boost::beast::http::async_read(
      stream, buffer, req, asioexec::use_sender);
}
```

Error occurs within `boost/beast/http/impl/read.hpp`:

```cpp
// If you get an error on the following line it means
// that your handler does not meet the documented type
// requirements for the handler.

static_assert(
    beast::detail::is_invocable<ReadHandler,
    void(error_code, std::size_t)>::value,
    "ReadHandler type requirements not met");
...
```

<details>

<summary>Compile time error</summary>

```
[proc] Executing command: /home/user/.local/bin/cmake --build ./build --config Debug --target all  --
[build] [1/2   0% :: 0.010] Re-checking globbed directories...
[build] [1/8  12% :: 0.365] Scanning ./main.cpp for CXX dependencies
[build] [2/8  25% :: 0.373] Generating CXX dyndep file CMakeFiles/hello_world.dir/CXX.dd
[build] [3/4  75% :: 6.722] Building CXX object CMakeFiles/hello_world.dir/main.cpp.o
[build] FAILED: CMakeFiles/hello_world.dir/main.cpp.o 
[build] /usr/bin/clang++-21 -I./build/_deps/stdexec-src/include -I./build/_deps/stdexec-build/include -g -std=gnu++23 -MD -MT CMakeFiles/hello_world.dir/main.cpp.o -MF CMakeFiles/hello_world.dir/main.cpp.o.d @CMakeFiles/hello_world.dir/main.cpp.o.modmap -o CMakeFiles/hello_world.dir/main.cpp.o -c ./main.cpp
[build] In file included from ./main.cpp:5:
[build] In file included from /usr/include/boost/beast.hpp:16:
[build] In file included from /usr/include/boost/beast/http.hpp:29:
[build] In file included from /usr/include/boost/beast/http/read.hpp:843:
[...]

PR #1566, opened 2025-07-08
asioexec::completion_token: Lvalue Completion Handler Invocation

RobertLeahy:
As added by 35a3e31590e736fbb7dd55324b3a7f991a059ce3 the completion handler type used by the implementation of asioexec::completion_token only supported rvalue-qualified invocation. This was due to the fact:

- Invocation of a completion handler is consumptive in the same way that calling set_value, set_error, or set_stopped on a receiver is (note those three operations require an rvalue-qualified operand), and
- Asio has invoked completion handlers with rvalue qualification since 1.20.0 (standalone)/1.77 (Boost)

Unfortunately there are operations in the wild which do not rvalue qualify their completion handlers when invoking them. For example the operations provided by Boost.Beast (at least as of this writing).

Updated asioexec::completion_token to support the above-described operations by removing the rvalue qualification from the function call operator of its completion handler type.

### DIFF
--- a/include/asioexec/completion_token.hpp
+++ b/include/asioexec/completion_token.hpp
@@ -244,7 +244,7 @@ namespace asioexec {
       }
 
       template <typename... Args>
-      void operator()(Args&&... args) && noexcept {
+      void operator()(Args&&... args) noexcept {
         STDEXEC_ASSERT(self_);
         {
           const std::lock_guard l(self_->m_);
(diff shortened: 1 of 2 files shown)


RobertLeahy:
Fixes #1534.

ericniebler:
thanks!

RobertLeahy:
#1534 should be resolved, I don't have permissions to close it though.

PR #1578, opened 2025-07-20
Allow just to Throw on Connect

RobertLeahy:
Previously if connecting a just sender with a receiver threw due to a throwing move operation terminate was called due to an exception passing through noexcept.

### DIFF
--- a/include/stdexec/__detail/__basic_sender.hpp
+++ b/include/stdexec/__detail/__basic_sender.hpp
@@ -228,7 +228,7 @@ namespace stdexec {
 
       __op_base(_Sexpr&& __sndr, _Receiver&& __rcvr) noexcept(
         __nothrow_decay_copyable<_Receiver>
-        && __noexcept_of<__sexpr_impl<__tag_t>::get_state, _Sexpr, _Receiver&>)
+        && noexcept(__state_t(__sexpr_impl<__tag_t>::get_state(static_cast<_Sexpr&&>(__sndr), __rcvr_))))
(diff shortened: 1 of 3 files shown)


include/stdexec/__detail/__basic_sender.hpp
```diff
@@ -228,7 +228,7 @@ namespace stdexec {
 
       __op_base(_Sexpr&& __sndr, _Receiver&& __rcvr) noexcept(
         __nothrow_decay_copyable<_Receiver>
-        && __noexcept_of<__sexpr_impl<__tag_t>::get_state, _Sexpr, _Receiver&>)
+        && noexcept(__state_t(__sexpr_impl<__tag_t>::get_state(static_cast<_Sexpr&&>(__sndr), __rcvr_))))
```
ericniebler, line 231: Why is this change necessary? Isn't it the same as what it was before?

RobertLeahy: Since there are two commits in this PR the PR message is thin (I hate the Github workflow).

The commit which actually fixes this bug, rather than fixing `make_fun_receiver` (which was necessary for the unit tests), has information on why this is necessary:

https://github.com/NVIDIA/stdexec/pull/1578/commits/daf9a01a95cfec21124d848d3ae85f3c663a249b

TL;DR: The existing check ensures that the function invocation does not throw. But in the case of `just` the function returns an rvalue reference which must thereafter be used to initialize the `__state_t` member, an operation which can throw even if the invocation does not/cannot throw.

ericniebler:
Thanks Robert!

PR #1585, opened 2025-07-23
allow using __optional with -fno-exceptions

khangng-ampere:
Currently, including `<exec/async_scope.hpp>` will err:

```c++
include/stdexec/__detail/__optional.hpp:122:11: error: cannot use 'throw' with exceptions disabled
          throw __bad_optional_access();
```


ericniebler:
thanks!

PR #1595, opened 2025-07-28
upgrade the CCCL dependency to 3.0.1 and rapids-cmake to 12.10

ericniebler:


trxcllnt:
I'm adding llvm21 devcontainers [here](https://github.com/rapidsai/devcontainers/pull/569) to see if that fixes the compile error. I suspect the missing symbol error is due to rapids-cmake [disabling the namespace magic](https://github.com/rapidsai/rapids-cmake/blob/0b111489d1e6f8400e1fc88297623a2a9915fa77/rapids-cmake/cpm/cccl.cmake#L107-L111) which could be tickling a bug in clang.

trxcllnt:
Nevermind, I still see the test failure when I re-enable the namespace magic. Something must be up with `thrust::universal_vector` in CCCL v3.0.1.

Issue #1609, opened 2025-08-13
cuda_fwd.cuh cannot require that <cuda/atomic> has not yet been included


dkolsen-pgi:
The header `nvexec/detail/cuda_fwd.cuh` contains this code:
```c++
#ifdef _CUDA_ATOMIC
#  error This file must be included before <cuda/atomic> from libcudacxx
#endif
```

No.  Just no.  You cannot require that a public header from some other library has not yet been included.  User code could include `<cuda/atomic>` before including any stdexec headers.

Something changed recently in stdexec or CCCL so that `#include <nvexec/stream_context.cuh>` triggers this `#error` when compiled with `nvc++ -stdpar`.  (`nvexec/stream_context.cuh` indirectly includes `stdexec/__detail/__variant.hpp`, which includes `<memory>`.  As part of NVC++'s implementation of the parallel algorithms in `<memory>`, some CCCL headers are included, one of which indirectly includes `<cuda/atomic>`.)

dkolsen-pgi:
This breaks every GPU test and example in our stdexec testing.  (At least I think it does.  This failure is hiding behind #1591, which also breaks every GPU test and example.)

ericniebler:
from `cuda_fwd.cuh`:

```c++
#if _LIBCUDACXX_STD_VER > 17 && _LIBCUDACXX_CUDA_API_VERSION < 1'009'000
#  define STDEXEC_LIBCUDACXX_NEEDS_ATOMIC_WORAROUND() 1
#else
#  define STDEXEC_LIBCUDACXX_NEEDS_ATOMIC_WORAROUND() 0
#endif
```

is using a more recent version of CCCL in your tests an options?

dkolsen-pgi:
The `#error` directive that aborts the compilation is at the very top of `cuda_fwd.cuh`, before the check that sets `STDEXEC_LIBCUDACXX_NEEDS_ATOMIC_WORAROUND`.  And we are already using the very latest CCCL main branch.  An older CCCL is not the problem.  And it has been years since we used a CCCL version old enough to cause `STDEXEC_LIBCUDACXX_NEEDS_ATOMIC_WORAROUND()` to be true.

Maybe this code:
```c++
#ifdef _CUDA_ATOMIC
#  error This file must be included before <cuda/atomic> from libcudacxx
#endif
```
[...]

PR #1616, opened 2025-08-21
remove the dependence of `__variant.hpp` on `<memory>`

ericniebler:


ericniebler:
this is a stab at a "fix" for #1609

PR #1620, opened 2025-08-25
asioexec::completion_token: Various Fixes/Updates

RobertLeahy:
See the description for each commit. Please do not squash.

### DIFF
--- a/test/asioexec/test_completion_token.cpp
+++ b/test/asioexec/test_completion_token.cpp
@@ -968,4 +968,22 @@ namespace {
     CHECK(ex);
   }
 
+  TEST_CASE(
+    "I/O objects may be transformed to use senders as their default vocabulary with only minimal "
+    "transformations (i.e. no error adaptation)",
+    "[asioexec][completion_token]") {
+    bool invoked = false;
+    asio_impl::io_context ctx;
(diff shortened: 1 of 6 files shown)


test/asioexec/test_completion_token.cpp
```diff
@@ -329,16 +330,33 @@ namespace {
...
+    //
+    //  /root/stdexec/test/asioexec/test_completion_token.cpp:380: FAILED:
+    //    CHECK( !ex )
+    //  with expansion:
+    //    false
+    //
+    //  /root/stdexec/test/asioexec/test_completion_token.cpp:382: FAILED:
+    //    CHECK( ptr.use_count() == 1U )
+    //  with expansion:
+    //    -1564306560 == 1
```
ericniebler, line 356: is this a bug in ASIO then?

RobertLeahy: I chased this around a bit before restructuring the function and realizing that fixed it. The closest I've come to explaining it is that it seems like either a code gen bug, or a bug in `std::shared_ptr`.

The fact that it only surfaced in a release build, not debug build, helps push me in this direction. But walking through a release with debug info build I saw the use count move unexpectedly to a nonsense value, I could never isolate whereat that actually happened.

PR #1637, opened 2025-09-13
cmake: Install asioexec and asio_pool headers

Perdixky:
### Description

This PR adds the necessary CMake installation rules for the `asioexec` and `asio_pool` components.

### Motivation and Context

I noticed that the Asio integration (`asioexec`) is mature and available for use within the project. However, its headers were not being installed, which prevents downstream projects from using this functionality when they depend on an installed version of `stdexec`.

This change makes the Asio-based schedulers fully usable for consumers of the installed library, consistent with other optional components like `tbbpool`.

### Changes Made

- Added `PUBLIC FILE_SET` definitions for the `stdexec_boost_pool`, `asioexec_boost`, `stdexec_asio_pool`, and `asioexec_asio` interface library targets.
- Ensured that the dynamically generated `asio_config.hpp` header is included in the file sets and correctly installed.
- Added `install(TARGETS ...)` rules for these components.

### How Has This Been Tested?

I have tested this locally by:
1. Configuring the project with `-DSTDEXEC_ENABLE_ASIO=ON`.
2. Building the project.
3. Running the install step: `cmake --install . --prefix ./install_dir`
4. Verifying that the headers from `include/asioexec` and `include/execpools/asio` are correctly placed within the `./install_dir/include` directory.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -457,16 +457,21 @@ if(STDEXEC_ENABLE_ASIO)
       message(FATAL_ERROR "Unknown configuration for ASIO implementation: " ${STDEXEC_ASIO_IMPLEMENTATION})
     endif()
 
-    file(GLOB_RECURSE boost_pool_sources include/execpools/asio/*.hpp)
-    set(STDEXEC_ASIO_CONFIG_FILE_PATH ${CMAKE_CURRENT_SOURCE_DIR}/include/execpools/asio)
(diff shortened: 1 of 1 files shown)


ericniebler (approved):

ericniebler:
@Perdixky this is fantastic. can you look into the CI failures, pls?

Perdixky:
Absolutely! I’ll investigate the CI failures and follow up soon.

PR #1639, opened 2025-09-13
the trampoline needs to be fifo to properly interleave sequences

kirkshoop:
also:
- added a stack size limit
- fixed the limits so that the origin trampoline limits apply to all nested trampolines


ericniebler (approved):
looks great, thanks kirk!

PR #1641, opened 2025-09-14
Use `get_completion_signatures` in `any_sender`.

jehelset:
Ensures that `any_sender` implements `get_completion_signatures`
rather than `completion_signatures`.

`get_completion_signatures` is constrained to only exist if the
passed environment satisfies any non-`never_stop_token` stop-token
receiver-queries. This avoids hard compilation errors when a
non-`never_stop_token` receiver-query is present and the sender
is probed for completion signatures in an empty environment.

Also renames `is_never_stop_token_query` to
`is_never_stop_token_query_v` to be consistent with the naming
used for `__is_not_stop_token_query_v`, where `_v` is a `__mbool`
and non-`_v` is a concept.

Solves issue in https://godbolt.org/z/bq3nGhdzr.

### DIFF
--- a/include/exec/any_sender_of.hpp
+++ b/include/exec/any_sender_of.hpp
@@ -17,6 +17,8 @@
 
 #include "../stdexec/execution.hpp"
 #include "../stdexec/__detail/__any_receiver_ref.hpp"
+#include "../stdexec/__detail/__concepts.hpp"
+#include "../stdexec/__detail/__env.hpp"
 #include "../stdexec/__detail/__transform_completion_signatures.hpp"
 
 #include "sequence_senders.hpp"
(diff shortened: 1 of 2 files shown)


include/exec/any_sender_of.hpp
```diff
@@ -612,16 +614,42 @@ namespace exec {
...
+      -> __mbool<true>;
+
+    template <class _Tag, class _Ret, class... _As>
+    auto __test_never_stop_token(_Tag (*)(_Ret (*)(_As...) noexcept)) -> __mbool<false>;
+
+    template <class _Tag, class _Ret, class... _As>
+    auto __test_never_stop_token(_Tag (*)(_Ret (*)(_As...))) -> __mbool<false>;
+
+    template <class _Query>
+    using __is_never_stop_token_query_v = decltype(__test_never_stop_token(
```
ericniebler: the `_v` suffix is typically used for variable templates. this is an alias template, so `_t` would be more appropriate.

include/exec/any_sender_of.hpp
```diff
@@ -1060,10 +1075,6 @@ namespace exec {
             __storage_.__get_object_pointer(), static_cast<__receiver_ref_t&&>(__receiver));
         }
 
-        explicit operator bool() const noexcept {
-          return __get_object_pointer(__storage_) != nullptr;
-        }
-
```
ericniebler, line 1066: i'm curious why you removed this conversion operator.

ericniebler (changes requested):
a nit and a question. otherwise looks good. thanks!

include/exec/any_sender_of.hpp
jehelset: it's just a drive-by cleanup, can remove it. from corresponding commit-message (i should have put it in the MR-text as well):

Remove unnused operator bool on `any_sender` base.

Not used anywhere and implementation calls a function that does
not exist: `__get_object_pointer(__storage)`. Instead it could
use `__storage_.__get_object_pointer()`, but since this is dead
code, might as well remove it.

include/exec/any_sender_of.hpp
jehelset: yap, fixed.

ericniebler (approved):

PR #1647, opened 2025-09-17
stdexec: fix exception support detection (clang-cl)

### LINKED ISSUE #1636: make exceptions work when compiling for windows with clang-cl
https://github.com/NVIDIA/stdexec/blob/daf12bcc46438f88b3c78212363f7b6531338780/include/stdexec/__detail/__config.hpp#L505-L511

Currently exceptions are "disabled" when compiling for Windows with **clang-cl** (LLVM 21.1.0 Release) and with /EHsc. Meaning, the `STDEXEC_TRY` macro expands to `if constexpr (true)`, so exceptions thrown from coroutines are not caught and eventually lead to a `std::terminate()` call in one of the `unhandled_exception()` definitions. Using a MSVC toolchain (cl.exe) with the same flags works as expected.

For **clang-cl** `__EXCEPTIONS` seems to be undefined so `STDEXEC_STD_NO_EXCEPTIONS` evaluates to `1`. The MSVC defines `_HAS_EXCEPTIONS` and `_CPPUNWIND` however are correctly set when using **clang-cl**.

Locally I "fixed" the issue by adding `|| STDEXEC_CLANG_CL()` in the referenced lines.

```cpp
#if STDEXEC_CUDA_COMPILATION() && defined(__CUDA_ARCH__)
#  define STDEXEC_STD_NO_EXCEPTIONS() 1
#elif STDEXEC_MSVC() || STDEXEC_CLANG_CL()
#  define STDEXEC_STD_NO_EXCEPTIONS() (_HAS_EXCEPTIONS == 0) || (_CPPUNWIND == 0)
#else
#  define STDEXEC_STD_NO_EXCEPTIONS() (__EXCEPTIONS == 0)
#endif
```

armpribb:
closes #1636

### DIFF
--- a/include/stdexec/__detail/__config.hpp
+++ b/include/stdexec/__detail/__config.hpp
@@ -504,7 +504,7 @@ namespace stdexec {
 
 #if STDEXEC_CUDA_COMPILATION() && defined(__CUDA_ARCH__)
 #  define STDEXEC_STD_NO_EXCEPTIONS() 1
-#elif STDEXEC_MSVC()
+#elif STDEXEC_MSVC() || STDEXEC_CLANG_CL()
 #  define STDEXEC_STD_NO_EXCEPTIONS() (_HAS_EXCEPTIONS == 0) || (_CPPUNWIND == 0)
 #else
 #  define STDEXEC_STD_NO_EXCEPTIONS() (__EXCEPTIONS == 0)


ericniebler (approved):

ericniebler:
thanks!

PR #1652, opened 2025-09-22
Optimize `async_scope` Performance, Make `spawn` Lock-Free

lixin-wei:
I'm using `async_scope.spawn()` in a hot function, and find the lock in it cost a lot.

This PR makes the `spawn` code path lock-free .


ericniebler:
this looks good to me. @kirkshoop and @lewissbaker do you agree?

PR #1654, opened 2025-09-27
write_env: No Evaluated Context for Environment Type

RobertLeahy:
The get_env implementation for stdexec::write_env is supplied via a lambda. Prior to this commit that lambda relied on automatic return type deduction. When a function relies on automatic return type deduction the compiler must evaluate the body thereof to determine the type returned thereby. A consequence of the aforegoing formulation is that if any statement in the body of the lambda directly or transitively relied on the completeness of an incomplete type compilation would fail.

The above might seem normal and non-problematic until one considers that it is fairly normal in C++ code to compute (and rely upon) types and other properties of functions long before any bona fide evaluation of that function occurs. This motivates the distinction between evaluated and unevaluated contexts. However as a consequence of the preceding paragraph the previous implementation of get_env meant that when code relied upon the return type thereof in an unevaluated context the context became an evaluated context in the body of get_env (and transitively through all statements therein).

The newly-added unit test highlights this. With the previous implementation of get_env (which uses automatic return type deduction) compilation fails. With the new implementation of get_env (which provides an explicit return type) compilation succeeds.

### DIFF
--- a/include/stdexec/__detail/__write_env.hpp
+++ b/include/stdexec/__detail/__write_env.hpp
@@ -61,10 +61,10 @@ namespace stdexec {
           stdexec::get_env(__child));
       };
 
-      static constexpr auto get_env =
-        [](__ignore, const auto& __state, const auto& __rcvr) noexcept {
-          return __env::__join(__state, stdexec::get_env(__rcvr));
-        };
+      static constexpr auto get_env = [](__ignore, const auto& __state, const auto& __rcvr) noexcept
+        -> decltype(__env::__join(__state, stdexec::get_env(__rcvr))) {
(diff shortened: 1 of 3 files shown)


RobertLeahy:
I noticed on further investigation that this change disagrees with [the wording](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2025/p3284r4.html#proposed-wording).

ericniebler (approved):
thanks! i guess we should file a LWG issue.

RobertLeahy:
> thanks! i guess we should file a LWG issue.

[This is LWG issue 4395.](https://cplusplus.github.io/LWG/issue4395)

PR #1655, opened 2025-09-27
stdexec::env: Queryability Pass Through

RobertLeahy:
stdexec::env is a class template which accepts any number of queryables and presents them as a unified/aggregated queryable. The intention is that stdexec::env is queryable in the same way that each of its children is. In the case where multiple children support a query the tie is broken by using the earliest queryable in the template argument list.

As an implementation detail stdexec::env is not simply a primary class template but rather an incomplete primary class template and three specializations which handle template argument lists which are:

- Nullary
- Nary
- Binary

Prior to this commit queryability of the latter two specializations was implemented through two member helpers:

- __get_1st, a static member function which retrieved the first queryable child, and
- __1st_env_t, a member type alias which computed the type of the first queryable child

With the latter being implemented as simply decltype of the former. Unfortunately prior to this commit __get_1st was not constrained meaning instantiations of stdexec::env with non-empty template argument lists always appeared queryable with a certain query type, with compilation failure occurring in a manner undetectable via SFINAE and requires expressions in the case where no child was actually thusly queryable.

Added reproducing tests and fixed.

### DIFF
--- a/.github/workflows/ci.cpu.yml
+++ b/.github/workflows/ci.cpu.yml
@@ -4,7 +4,6 @@ on:
   push:
     branches:
       - main
-      - "member-function-customization"
       - "pull-request/[0-9]+"
 
 concurrency:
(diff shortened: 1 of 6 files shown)


ericniebler:
in #1656 i am replacing the implementation of `env` with something simpler that shouldn't have this problem. when that lands, we should rebase this PR on that and land your new tests.

PR #1660, opened 2025-10-03
Remove unnecessary `typename` before dependent names

localspook:
Now that the minimum clang version is 16 (#1657), the code can take advantage of implicit typename, right?

### DIFF
--- a/include/stdexec/__detail/__basic_sender.hpp
+++ b/include/stdexec/__detail/__basic_sender.hpp
@@ -391,7 +391,7 @@ namespace stdexec {
     using __desc_t = __decay_t<_Sexpr>::__desc_t;
     using __tag_t = __desc_t::__tag;
     using __data_t = __desc_t::__data;
-    using __state_t = __op_state::__state_t;
+    using __state_t = __op_state::__op_base::__state_t;
     using __inner_ops_t =
       __result_of<__sexpr_apply, _Sexpr, __detail::__connect_fn<_Sexpr, _Receiver>>;
 


ericniebler:
you would think, but i've seen msvc struggle without the `typename`. i'll run the tests to confirm, but i think this pr is premature.

localspook:
Now that's quite bizarre. This is the pattern that's causing the failures:

```cpp
template <typename T>
struct base {
    using type = int;
};

template <typename T>
struct derived : base<T> {
    using type = /*typename*/ derived::type; // Finds base<T>::type
};

template struct derived<int>;
```

Looks like (ironically) MSVC accepts it, Clang accepts it only since 18, and GCC doesn't accept it at all: https://godbolt.org/z/EK33394hP

ericniebler:
> ```cpp
> template <typename T>
> struct derived : base<T> {
>     using type = /*typename*/ derived::type; // Finds base<T>::type
> };
> ```

oh god that's terrible code. where are we doing that? i'd prefer:

```cpp
template <typename T>
struct derived : base<T> {
    using type = /*typename*/ derived::base::type; // Finds base<T>::type
};
```

localspook:
`__state_t` in:
https://github.com/NVIDIA/stdexec/blob/770f7959142c2446319f642c82aed8b07aea3f86/include/stdexec/__detail/__basic_sender.hpp#L389-L394
Finds the `__state_t` in the base class:
https://github.com/NVIDIA/stdexec/blob/770f7959142c2446319f642c82aed8b07aea3f86/include/stdexec/__detail/__basic_sender.hpp#L219-L222

ericniebler:
> ```cpp
>    using __state_t = typename __op_state::__state_t; 
> ```

can you try changing it to:

```cpp
   using __state_t = typename __op_state::__op_base::__state_t; 
```

?

localspook:
Pushed a commit with that change, please retest

PR #1661, opened 2025-10-07
Fix constraint on `__optional::__emplace_from`.

jehelset:
The `_Args` template parameter pack is currently not passed to the `__optional<_Tp>::__emplace_from` constraint that checks that the return-type of invoking `_Fn` with `_Args...` is `_Tp`.


ericniebler (approved):

ericniebler:
nice catch, thanks!

PR #1664, opened 2025-10-13
Add merge each sender adaptor

kirkshoop:
This is a seminal event. `flat_map` (aka *map-merge* or *transform-merge*) 
can be composed from `transform_each` and `merge_each`.  
`flat_map` is the root of all async sequence algorithms. 

In other libraries merge has the problem of unbounded allocations. 
This `merge_each` does zero allocations.

This `merge_each` also supports bulk processing in a more natural form.
each value can be running in parallel on different execution contexts. 
A future *fork* adaptor would take a sequence and fork portions of the 
sequence into separate sequences on separate schedulers. Then a 
`merge_each` would gather all the result values into a new sequence. 
Adding a `continue_each_on` after `merge_each` using a strand scheduler 
would remove all the parallel completions by queueing the result values and
emitting them one at a time.

-----

`merge_each` is a *sequence adaptor* that takes a sequence of nested
sequences and merges all the nested values from all the nested
sequences into a single output sequence.

the first error encountered will trigger a stop request for all
active operations. The error is stored and is emitted only after
all the active operations have completed.

If the error was emitted from an item, a new item is emitted
at the end to deliver the stored error.

any nested sequence or nested value that completes with
`set_stopped` will not cause any other operations to be stopped.
This allows individual nested sequences to be stopped without
breaking the merge of the remaining sequences.

### DIFF
--- a/include/exec/sequence/merge_each.hpp
+++ b/include/exec/sequence/merge_each.hpp
@@ -1,0 +1,1290 @@
+/*
+ * Copyright (c) 2023 Maikel Nadolski
+ * Copyright (c) 2023 NVIDIA Corporation
+ *
+ * Licensed under the Apache License Version 2.0 with LLVM Exceptions
+ * (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ *   https://llvm.org/LICENSE.txt
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#pragma once
+
+#include <exception>
+
+#include "../../stdexec/concepts.hpp"
+#include "../../stdexec/execution.hpp"
+#include "../sequence_senders.hpp"
+
+#include "../__detail/__basic_sequence.hpp"
+#include "../../stdexec/__detail/__concepts.hpp"
+#include "../../stdexec/__detail/__config.hpp"
+#include "../../stdexec/__detail/__diagnostics.hpp"
+#include "../../stdexec/__detail/__execution_fwd.hpp"
+#include "../../stdexec/__detail/__meta.hpp"
+#include "../../stdexec/__detail/__sender_introspection.hpp"
+#include "../../stdexec/__detail/__senders_core.hpp"
+#include "../../stdexec/__detail/__stop_token.hpp"
+#include "../../stdexec/__detail/__transform_completion_signatures.hpp"
+#include "../../stdexec/__detail/__unstoppable.hpp"
+#include "../../stdexec/__detail/__variant.hpp"
+
+#include <atomic>
+
+namespace exec {
+  namespace __merge_each {
+    using namespace stdexec;
+
+    struct __env_with_inplace_stop_token_t {
+      auto operator()(inplace_stop_token& __stop_token) const noexcept {
+        return stdexec::prop{stdexec::get_stop_token, __stop_token};
+      }
+      template <class _Env>
+      auto operator()(stdexec::inplace_stop_token& __stop_token, _Env&& __env) const noexcept {
+        return __env::__join((*this)(__stop_token), static_cast<_Env&&>(__env));
+      }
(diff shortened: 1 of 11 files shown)


kirkshoop:
I think that we need to stop using TSAN on gcc11. 
I think that the warnings from TSAN that are breaking the builds are false positives:
  [Bug 101978 - thread sanitizer false positive when condition variable](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=101978)
Could we switch the TSAN builds to a later gcc?

kirkshoop:
Ready for another go at the ci

kirkshoop:
Ready to try CI again

include/exec/sequence/merge_each.hpp
```diff
@@ -0,0 +1,1290 @@
...
+#include "stdexec/__detail/__config.hpp"
+#include "stdexec/__detail/__diagnostics.hpp"
+#include "stdexec/__detail/__execution_fwd.hpp"
+#include "stdexec/__detail/__meta.hpp"
+#include "stdexec/__detail/__sender_introspection.hpp"
+#include "stdexec/__detail/__senders_core.hpp"
+#include "stdexec/__detail/__stop_token.hpp"
+#include "stdexec/__detail/__transform_completion_signatures.hpp"
+#include "stdexec/__detail/__unstoppable.hpp"
+#include "stdexec/__detail/__variant.hpp"
```
ericniebler:
```suggestion
#include "../../stdexec/__detail/__concepts.hpp"
#include "../../stdexec/__detail/__config.hpp"
#include "../../stdexec/__detail/__diagnostics.hpp"
#include "../../stdexec/__detail/__execution_fwd.hpp"
#include "../../stdexec/__detail/__meta.hpp"
#include "../../stdexec/__detail/__sender_introspection.hpp"
#include "../../stdexec/__detail/__senders_core.hpp"
#include "../../stdexec/__detail/__stop_token.hpp"
#include "../../stdexec/__detail/__transform_completion_signatures.hpp"
#include "../../stdexec/__detail/__unstoppable.hpp"
#include "../../stdexec/__detail/__variant.hpp"
```

include/exec/sequence_senders.hpp
```diff
@@ -17,10 +17,72 @@
 #pragma once
 
 #include "../stdexec/execution.hpp"
-#include "../stdexec/__detail/__concepts.hpp"
-#include "../stdexec/__detail/__meta.hpp"
+#include "stdexec/__detail/__concepts.hpp"
+#include "stdexec/__detail/__meta.hpp"
+#include "stdexec/__detail/__diagnostics.hpp"
```
ericniebler:
```suggestion
#include "../stdexec/__detail/__concepts.hpp"
#include "../stdexec/__detail/__meta.hpp"
#include "../stdexec/__detail/__diagnostics.hpp"
```

kirkshoop:
> /ok to test 29e445b

It looks like this commit passed the checks.
The failures are after a force push. Perhaps we can go back to 29e445b

ericniebler:
> The failures are after a force push. Perhaps we can go back to [29e445b]

yeah, silly me. i tried to rebase and squash all the little fixit commits. somehow i screwed up `sequence_senders.hpp`. i think i have fixed the problem and force-pushed again. maybe you could check that `sequence_senders.hpp` is as it should be now.

include/exec/sequence_senders.hpp
```diff
@@ -449,52 +545,73 @@ namespace exec {
...
-      receiver<_Receiver> && sequence_sender_in<_Sequence, env_of_t<_Receiver>>
-      && sequence_receiver_from<_Receiver, _Sequence>
-      && requires(_Sequence&& __sequence, _Receiver&& __rcvr) {
-           {
-             STDEXEC_REMOVE_REFERENCE(_Sequence)
-             ::subscribe(static_cast<_Sequence&&>(__sequence), static_cast<_Receiver&&>(__rcvr))
-           };
-         };
+    using __subscribe_member_result_t = decltype(__declval<_Sequence>()
+                                                   .subscribe(__declval<_Receiver>()));
```
ericniebler, line 549: @kirkshoop could you pls give this a spot check? i had to reconstruct some of your work after a bad rebase/force push, and i'm not 100% sure i got everything back the way it should be.

kirkshoop: I checked, it is close enough. Ship it! - None of the remaining differences are functional, just spelling and strings..

PR #1666, opened 2025-10-21
stdexec::when_all: P3887

RobertLeahy:
Applies the fixes proposed by P3887 to stdexec::when_all. Particularly:

- Removes early detection and handling of stop requests
- Eliminates stdexec::set_stopped_t() as an advertised completion signatures when no child sender sends that signature


RobertLeahy:
https://isocpp.org/files/papers/P3887R0.pdf

PR #1667, opened 2025-10-24
exec::unless_stop_requested

RobertLeahy:
Adds the functionality removed from `stdexec::when_all` by #1666 (see also [P3887](https://isocpp.org/files/papers/P3887R0.pdf)) back as a separate algorithm.

### DIFF
--- a/.github/workflows/ci.cpu.yml
+++ b/.github/workflows/ci.cpu.yml
@@ -23,8 +23,8 @@ jobs:
           - { name: "CPU (clang 16, Release)",     build: "Release", tag: llvm16-cuda12.9, cxxflags: "-stdlib=libc++" }
           - { name: "CPU (gcc 11, Debug)",         build: "Debug",   tag: gcc11-cuda12.9,  cxxflags: "", }
           - { name: "CPU (gcc 11, Release)",       build: "Release", tag: gcc11-cuda12.9,  cxxflags: "", }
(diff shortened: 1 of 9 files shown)


include/exec/unless_stop_requested.hpp
```diff
@@ -0,0 +1,103 @@
...
+      static constexpr auto start = []<class _State, class _Receiver, class _Operation>(
+                                      _State&,
+                                      _Receiver& __rcvr,
+                                      _Operation& __child_op) noexcept -> void {
+        if constexpr (!__unstoppable_receiver<_Receiver>) {
+          if (get_stop_token(stdexec::get_env(__rcvr)).stop_requested()) {
+            stdexec::set_stopped((_Receiver&&) __rcvr);
+            return;
+          }
+        }
```
ispeters: This code is only ever executed when `!__unstoppable_receiver<_Receiver>` is true, right? If so, the `if constexpr` can be dropped.
```suggestion
        static_assert(!__unstoppable_receiver<_Receiver>);

        if (get_stop_token(stdexec::get_env(__rcvr)).stop_requested()) {
          stdexec::set_stopped((_Receiver&&) __rcvr);
          return;
        }
```

RobertLeahy:
Note this is now [P3892](https://isocpp.org/files/papers/P3892R0.pdf).

PR #1674, opened 2025-11-03
Optimize the performance of `intrusive_heap`

TypeCombinator:
Recently, I refactored the intrusive heap. Conservatively speaking, after the refactoring, the performance of the intrusive heap on small datasets has nearly doubled, while on large datasets, it has improved by at least 20%, depending on the specific application scenario.

When inserting into or removing nodes from a binary heap, the parent pointers of nodes along the sift-up or sift-down path remain unchanged. By leveraging this fact, we can reduce the maintenance of parent pointers in the heap. Moreover, using double pointers greatly simplifies operations and eliminates many branches.

Undoubtedly, there are faster implementations available, such as the quad heap. However, without altering the original design, preserving compatibility, and after careful trade-offs, this is arguably the best implementation you will find.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -220,9 +220,6 @@ endif()
 
 add_library(STDEXEC::stdexec ALIAS stdexec)
 
-# Don't require building everything when installing
-set(CMAKE_SKIP_INSTALL_ALL_DEPENDENCY ON)
-
 # Support target for examples and tests
 add_library(stdexec_executable_flags INTERFACE)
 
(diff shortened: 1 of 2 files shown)


TypeCombinator:
Ping...

ericniebler:
> Ping...

sorry for the delay. last week was a c++ standardization meeting. i was maxed.

TypeCombinator:
> > Ping...
> 
> sorry for the delay. last week was a c++ standardization meeting. i was maxed.

No worries, and thanks for your reply! I was aware of the C++ meeting from my email with Maikel, so I totally understand and was happy to wait.

Issue #1684, opened 2025-11-25
asio completion_token error with async_read and streambuf


SidneyCogdill:
because asioexec completion token implementation doesn't perfect forward the reference, and `asio::streambuf` is non-copyable and non-movable.

https://github.com/NVIDIA/stdexec/blob/394a7c50237fbe3151350027a7786f001de67355/include/asioexec/completion_token.hpp#L352-L366

Code snippet to reproduce:

```cpp
#include <asioexec/use_sender.hpp>
#include <boost/asio.hpp>
#include <stdexec/execution.hpp>

auto f(boost::asio::readable_pipe pipe, boost::asio::streambuf buf)
{
    auto p = boost::asio::async_read(pipe, buf, asioexec::use_sender);
}
```

<details><summary>compiler error messages</summary>

```
[build] ~/.cache/CPM/stdexec/ab26/stdexec/include/asioexec/completion_token.hpp:532:14: error: no matching constructor for initialization of '::asioexec::detail::completion_token::sender< ::asioexec::detail::completion_token::completion_signatures<void (boost::system::error_code, unsigned long)>, std::remove_cvref_t<boost::asio::detail::initiate_async_read_dynbuf_v1<boost::asio::basic_readable_pipe<boost::asio::any_io_executor>>>, boost::asio::basic_streambuf<std::allocator<char>> &, boost::asio::detail::transfer_all_t>' (aka '::asioexec::detail::completion_token::sender< ::stdexec::completion_signatures< ::stdexec::set_value_t (boost::system::error_code, unsigned long), ::stdexec::set_error_t (std::exception_ptr), ::stdexec::set_stopped_t ()>, boost::asio::detail::initiate_async_read_dynbuf_v1<boost::asio::basic_readable_pipe<boost::asio::any_io_executor>>, boost::asio::basic_streambuf<std::allocator<char>> &, boost::asio::detail::transfer_all_t>') [clang-diagnostic-error]
[build]   532 |       return ::asioexec::detail::completion_token::sender<
[build]       |              ^
[build]   533 |         ::asioexec::detail::completion_token::completion_signatures<Signatures...>,
[build]   534 |         std::remove_cvref_t<Initiation>,
[build]   535 |         Args...
[build]   536 |       >(static_cast<Initiation&&>(i), static_cast<Args&&>(args)...);
[...]

PR #1685, opened 2025-11-27
asioexec::completion_token & ::use_sender: Constrain Initiate

RobertLeahy:
Previously async_result<asioexec::completion_token, ...>:: and async_result<asioexec::use_sender, ...>::initiate:

- Used automatic return type deduction, and
- Were not constrained on any properties of the arguments thereto

This meant that initiating functions of the form were problematic:

```c++
  template <typename AsyncReadStream, typename Allocator,
      BOOST_ASIO_COMPLETION_TOKEN_FOR(void (boost::system::error_code,
        std::size_t)) ReadToken = default_completion_token_t<
          typename AsyncReadStream::executor_type>>
  inline auto async_read(AsyncReadStream& s, basic_streambuf<Allocator>& b,
      ReadToken&& token = default_completion_token_t<
        typename AsyncReadStream::executor_type>(),
      constraint_t<
        !is_completion_condition<ReadToken>::value
      > = 0)
    -> decltype(
      async_initiate<ReadToken,
        void (boost::system::error_code, std::size_t)>(
          declval<detail::initiate_async_read_dynbuf_v1<AsyncReadStream>>(),
          token, basic_streambuf_ref<Allocator>(b), transfer_all()));
```

Note that:

- The above is an actual initiating function from Asio, and
- The return type is defined in terms of the type returned by async_initiate

The latter of the two points above means that the compiler must determine the type of an expression involving async_initiate. Since that expression transitively defers to async_result<...>::initiate this meant that the compiler had to determine the return type thereof. Since (as mentioned above) that function uses automatic return type deduction that meant the compiler had to substitute into the body thereof. Since errors in such a context are not SFINAE-friendly this meant that if the variadic pack of arguments thereto did not consist only of decay- copyable types there would be a hard compilation error (rather than the overload simply being removed from the candidate set).

Both newly-added unit tests failed to compile before this change due to the above.

Added a constraint to both previously-mentioned initiate functions to address the above.

### DIFF
--- a/include/asioexec/completion_token.hpp
+++ b/include/asioexec/completion_token.hpp
@@ -527,6 +527,7 @@ namespace ASIOEXEC_ASIO_NAMESPACE {
   template <typename... Signatures>
   struct async_result<::asioexec::completion_token_t, Signatures...> {
     template <typename Initiation, typename... Args>
+      requires (std::is_constructible_v<std::decay_t<Args>, Args> && ...)
     static constexpr auto
       initiate(Initiation&& i, const ::asioexec::completion_token_t&, Args&&... args) {
       return ::asioexec::detail::completion_token::sender<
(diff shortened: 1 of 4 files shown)


RobertLeahy:
Addresses #1684.

ericniebler:
thanks, robert!

PR #1697, opened 2025-12-02
Remove `sccache-dist` network envvars, add NVHPC v25.9 devcontainers and CI jobs

trxcllnt:


trxcllnt:
I believe the clang16 CPU tests should be fixed by https://github.com/rapidsai/devcontainers/pull/624

ericniebler:
thanks @trxcllnt !

PR #1713, opened 2025-12-14
Implement P3149R11 and P3815R1

ispeters:
This PR will implement P3149R11 patched with P3815R1.

- [exec.scope.concepts]
    - [x] `concept scope_association`
    - [x] `concept scope_token`
- [exec.associate]
    - [ ] `associate` algorithm
- [exec.stop.when]
    - [x] _`stop-when`_ algorithm
- [exec.spawn]
    - [x] `spawn` algorithm
- [exec.spawn.future]
    - [ ] `spawn_future` algorithm
- [exec.counting.scopes]
    - [x] `simple_counting_scope`
    - [x] `counting_scope`

### DIFF
--- a/include/stdexec/__detail/__spawn_future.hpp
+++ b/include/stdexec/__detail/__spawn_future.hpp
@@ -558,7 +558,7 @@ namespace STDEXEC {
         //     responsibility.
 
         const auto __rcvr = __sentinel;
-        auto __callback = __self->__callback_;
+        auto __cb = __self->__callback_;
 
         // We have already synchronized with the consumer by performing a load-acquire above (and
         // we know it was the *consumer* we synchronized with because __sentinel contains the address
@@ -572,7 +572,7 @@ namespace STDEXEC {
         // __registered_receiver_; when the producer completes, it will load-acquire that value and
         // clean up.
         //
-        // In the second case, the producere will complete the registered receiver, which will have
+        // In the second case, the producer will complete the registered receiver, which will have
         // the side effect of destroying the stop callback that we're running inside, which is a
         // synchronization point with us.
         //
@@ -582,12 +582,12 @@ namespace STDEXEC {
         if (__sentinel == __rcvr) {
           // __registered_receiver_ still contained the value we observed during the CAS, which means
           // the producer still hadn't updated it to contain (__self). This means we succeeded in
-          // marking the operation as abandonded and the producer will destroy it when it completes
+          // marking the operation as abandoned and the producer will destroy it when it completes
           // (which could happen at any moment); we need to complete the consumer with set_stopped. We
-          // invoke __callback (the copy of __callback_ that we put on the stack) with a null
-          // self-pointer to signal that it ought to invoke set_stopped(std::move(*__rcvr)) without
-          // touching the operation.
-          __callback(nullptr, __rcvr);
+          // invoke __cb (the copy of __callback_ that we put on the stack) with a null self-pointer
+          // to signal that it ought to invoke set_stopped(std::move(*__rcvr)) without touching the
+          // operation.
+          __cb(nullptr, __rcvr);
         } else {
           STDEXEC_ASSERT(__sentinel == __self);
           // The producer beat us to the punch; it's busy trying to complete and is about to


include/stdexec/__detail/__scope_concepts.hpp
```diff
@@ -0,0 +1,65 @@
+/*
+ * Copyright (c) 2025 Ian Petersen
```
ericniebler, line 2: pls also include the NVIDIA copyright

ispeters:
The latest push includes an implementation of `stdexec::associate` with some tests, but the tests are incomplete so I haven't marked `associate` as done in the summary.

I have also tried to address the build failures on the previous iteration, but I don't have a local test environment for the failing build types so I haven't tested the fixes locally, other than to confirm that they still build with my local Clang (i.e. `Homebrew clang version 21.1.7`).

ispeters:
Rebased on #1717.

ispeters:
I managed to get GCC 12 installed on my Mac and it repro'd the ICE and the build failure in `test_stop_when.cpp`; I _think_ I've addressed the build failures in GCC, nvcc, and MSVC.

ispeters:
`simple_counting_scope` still needs more tests, but I did discover that [[exec.simple.counting.mem], paragraph 9](https://eel.is/c++draft/exec#simple.counting.mem-9) has a bug: we need to move the scope to the _`joined`_ state and return `true` if _`count`_ is zero, regardless of the value of _`state`_.

ispeters:
I _think_ I've fixed the build breaks in old gcc, nvc++, and MSVC. Getting a repro of the gcc 11 builds was quite a pain—I can't figure out how to get gcc 11 to run properly on macOS Tahoe so I had to use `container`, which I learned about [here](https://www.chicagovps.net/blog/running-kali-linux-natively-on-macos-a-guide-to-using-apple-containers). Fixing MSVC is a matter of guessing. I can't run nvc++ on my Mac, but it turns out it's available on Godbolt, and [this](https://godbolt.org/z/Yj7voaxne) was very helpful.

I've updated `__counting_scopes.hpp` rather extensively; I renamed some of the variables in `simple_counting_scope` to make the naming more consistent, and I've thoroughly documented both the paragraphs in the standard being implemented and my justifications for weaker-than-seq_cst memory orderings.

ispeters:
Pull the guts of `simple_counting_scope` into a `__base_scope` and implement both `simple_counting_scope` and `counting_scope` in terms of it.

ispeters:
Do some template hoisting and renaming, and replace a single `virtual` function with a function pointer to cut down on RTTI.

ispeters:
I can't seem to make Godbolt make stdexec available to MSVC, but [this](https://godbolt.org/z/ahxMeh6en) suggests the Clangs, gccs, and nvccs of the world are happy with the current iteration.

ispeters:
Reflect the file rename in `CMakeLists.txt`.

ispeters:
Add `spawn_future` and rebase on #1682.

ispeters:
I think I've fixed both the gcc build errors and the memory leak. The GPUs seems to be on holiday, though—I haven't seen any CI results from the GPU jobs in a while.

ispeters:
* Rebase on a bunch of recent commits.
* Incorporate @ericniebler's "add missing include" change into [401944f](https://github.com/NVIDIA/stdexec/pull/1713/commits/401944fda1727fd94d52f0c4a240f7c5e7ad3093) and update authorship attribution
* Simplify the implementation of forwarding stop requests by only invoking `__abandon` if the operation is never started
* Still haven't addressed the ASAN error, though

ispeters:
I just force-pushed after rebasing on `main`.

ispeters:
I've re-jiggered the implementation of stop request handling in `spawn_future`; I need to hand over the computer to my kid now, so I haven't had a chance to re-review my own code like I usually do, but I _think_ this cleans up the implementation some, at the cost of another virtual function on `__future_state_base`.

include/stdexec/__detail/__associate.hpp
```diff
@@ -0,0 +1,252 @@
...
+      //         using __sender_ref = std::unique_ptr<
+      //             __wrap_sender_t,
+      //             // this decltype(<lamnda>) breaks things
+      //             decltype([](auto* p) noexcept { std::destroy_at(p); })
+      //         >;
+      //
+      //       but the above code ICEs gcc 11 and 12 (and maybe MSVC)
+      //       so we declare a named callable
+      struct __deleter {
+        constexpr void operator()(__wrap_sender_t* p) const noexcept {
```
ericniebler: all identifier names need to be uglified.
```suggestion
        constexpr void operator()(__wrap_sender_t* __p) const noexcept {
```

include/stdexec/__detail/__associate.hpp
```diff
@@ -0,0 +1,252 @@
...
+        __nothrow_copy_constructible<__wrap_sender_t> && noexcept(__other.__assoc_.try_associate()))
+        requires __std::copy_constructible<__wrap_sender_t>
+        : __assoc_(__other.__assoc_.try_associate()) {
+        if (__assoc_) {
+          std::construct_at(&__sndr_, __other.__sndr_);
+        }
+      }
+
+      __associate_data(__associate_data&& __other)
+        noexcept(__nothrow_move_constructible<__wrap_sender_t>)
```
ericniebler, line 98: move operations should all be `noexcept` unconditionally. you can then `static_assert` that `__nothrow_move_constructible<__wrap_sender_t>` is `true`.

include/stdexec/__detail/__associate.hpp
```diff
@@ -0,0 +1,252 @@
...
+      //
+      //       but the above code ICEs gcc 11 and 12 (and maybe MSVC)
+      //       so we declare a named callable
+      struct __deleter {
+        constexpr void operator()(__wrap_sender_t* p) const noexcept {
+          std::destroy_at(p);
+        }
+      };
+
+      using __sender_ref = std::unique_ptr<__wrap_sender_t, __deleter>;
```
ericniebler, line 66: using a `unique_ptr` for this seems like overkill. my preference would be to use the `__scope_guard` in `__detail/__scope.hpp` instead. something like:
```c++
__scope_guard __g{std::destroy_at<_Sender>, &__sndr};
```

include/stdexec/__detail/__associate.hpp
```diff
@@ -0,0 +1,252 @@
...
+      }
+
+      ~__associate_data() {
+        if (__assoc_) {
+          std::destroy_at(&__sndr_);
+        }
+      }
+
+      std::pair<__assoc_t, __sender_ref> release() && noexcept {
+        __sender_ref u(__assoc_ ? std::addressof(__sndr_) : nullptr);
```
ericniebler: i don't think you can pass `nullptr` to `std::destroy_at` the way you can with `delete`.

include/stdexec/__detail/__associate.hpp
```diff
@@ -0,0 +1,252 @@
...
+        return {std::move(__assoc_), std::move(u)};
+      }
+
+     private:
+      __associate_data(std::pair<__assoc_t, __sender_ref> __parts)
```
ericniebler:
```suggestion
      explicit __associate_data(std::pair<__assoc_t, __sender_ref> __parts)
```

include/stdexec/__detail/__counting_scopes.hpp
```diff
@@ -0,0 +1,764 @@
...
+      template <class _Env>
+      using __scheduler_of_t = __call_result_t<get_scheduler_t, const _Env&>;
+
+      template <class _Env>
+      using __sched_sender_of_t = __call_result_t<schedule_t, __scheduler_of_t<_Env>>;
+
+      template <class _Sender, class _Env>
+      static consteval auto get_completion_signatures() //
+        -> __completion_signatures_of_t<__sched_sender_of_t<_Env>, _Env> {
+        return {};
```
ericniebler: in c++26, `get_completion_signatures` propagates errors with constexpr exceptions. that means `STDEXEC::get_completion_signatures` actually needs to be called. (there are a lot of senders in stdexec that need updating for this reason.)

```suggestion
      static consteval auto get_completion_signatures() {
        return STDEXEC::get_completion_signatures<__sched_sender_of_t<_Env>, _Env>();
```

include/stdexec/__detail/__counting_scopes.hpp
```diff
@@ -0,0 +1,764 @@
...
+      };
+
+      template <class _Rcvr>
+      struct __join_state final : __join_state_base {
```
ericniebler:
```suggestion
      struct __join_state : __join_state_base {
```

include/stdexec/__detail/__counting_scopes.hpp
```diff
@@ -0,0 +1,764 @@
...
+        __bits_.fetch_or(__closed, __std::memory_order_release);
+      }
+
+      bool __try_associate() noexcept {
+        constexpr auto makeNewBits = [](std::size_t bits) noexcept {
+          // [exec.simple.counting.mem] paragraph 5 and 5.3 say there is no
+          // effect if state is closed or if count is equal to max_associations
+          // so we should not be calculating a new (count, state) pair if either
+          // of those conditions hold.
+          assert(!__is_closed(bits) && __count(bits) < max_associations);
```
ericniebler:
```suggestion
          STDEXEC_ASSERT(!__is_closed(bits) && __count(bits) < max_associations);
```

include/stdexec/__detail/__counting_scopes.hpp
```diff
@@ -0,0 +1,764 @@
...
+          //
+          // NOTE: Paragraph 5.3 applies when state is closed, closed-and-joining,
+          //       or joined, all of which can be detected by checking the closed bit
+          if (__is_closed(oldBits) || __count(oldBits) == max_associations) {
+            // Paragraph 6 (the Returns clause) says we return assoc-t() "otherwise",
+            // which applies when count is not incremented, i.e. right here.
+            return false;
+          }
+
+          newBits = makeNewBits(oldBits);
```
ericniebler: no camel case, pls

include/stdexec/__detail/__counting_scopes.hpp
```diff
@@ -0,0 +1,764 @@
...
+
+        return (__newCount << 3) | __state(__newState);
+      }
+
+      bool __register(__join_state_base& __joinOp) noexcept {
+        // we need acquire semantics in case the join operation being started is about to
+        // observe that the last decrement has already happened; in that case, we need to
+        // establish that all of the now-completed associated operations happen-before the
+        // completion of this join operation
+        auto* head = __registeredJoinOps_.load(__std::memory_order_acquire);
```
ericniebler:
```suggestion
        auto* __head = __registeredJoinOps_.load(__std::memory_order_acquire);
```

include/stdexec/__detail/__spawn.hpp
```diff
@@ -0,0 +1,141 @@
...
+namespace STDEXEC {
+  /////////////////////////////////////////////////////////////////////////////
+  // [exec.spawn]
+  namespace __spawn {
+    struct __spawn_state_base {
+      __spawn_state_base() = default;
+
+      __spawn_state_base(__spawn_state_base&&) = delete;
+
+      virtual void __complete() noexcept = 0;
```
ericniebler: elsewhere you avoid virtuals because of RTTI bloat. should this also be a function pointer?

include/stdexec/__detail/__spawn.hpp
```diff
@@ -0,0 +1,141 @@
...
+
+        auto* op = traits::allocate(alloc, 1);
+
+        try {
+          traits::construct(
+            alloc, op, alloc, std::move(senderWithEnv), static_cast<_Token&&>(__tkn));
+        } catch (...) {
+          traits::deallocate(alloc, op, 1);
+          throw;
+        }
```
ericniebler: pls use a scope guard instead. `catch(...) { throw; }` makes debugging harder because it obfuscates the source of the exception.

include/stdexec/__detail/__spawn_common.hpp
```diff
@@ -0,0 +1,102 @@
...
+
+    // [exec.spawn] paragraph 9
+    // [exec.spawn.future] paragraph 15
+    //   (9/15.3) -- otherwise, alloc is allocator<void>()
+    template <class _Env, class _Attrs>
+      requires(!__callable<get_allocator_t, const _Env&>)
+           && (!__callable<get_allocator_t, const _Attrs&>)
+    std::allocator<void> operator()(const _Env&, const _Attrs&) const {
+      return {};
+    }
```
ericniebler: one overload with cascading `if constexpr` would be a lot more readable.

include/stdexec/__detail/__spawn_future.hpp
```diff
@@ -0,0 +1,756 @@
...
+
+    template <class _Sig>
+    struct __future_sig_fns;
+
+    template <class _Tag, class... _Args>
+    struct __future_sig_fns<_Tag(_Args...)> {
+      using __tuple = __decayed_tuple<_Tag, _Args...>;
+
+      static constexpr bool __is_nothrow_storable =
+        (__nothrow_constructible_from<__decay_t<_Args>, _Args> && ...);
```
ericniebler:
```suggestion
        __nothrow_decay_copyable<_Args...>;
```

include/stdexec/__detail/__spawn_future.hpp
```diff
@@ -0,0 +1,756 @@
...
+
+      using __decayed_sig = _Tag(__decay_t<_Args>...);
+    };
+
+    // [exec.spawn.future] paragraph 4
+    template <class _Sig>
+    using __as_tuple = __future_sig_fns<_Sig>::__tuple;
```
ericniebler: just fyi, this could be `__mapply_q<__decayed_tuple, _Sig>`.

include/stdexec/__detail/__spawn_future.hpp
```diff
@@ -0,0 +1,756 @@
...
+        __munique<__qq<completion_signatures>>,
+        set_stopped_t(),
+        set_error_t(std::exception_ptr),
+        __decayed_sig<_Sigs>...
+      >;
+    };
+
+    template <class _Env>
+    inline constexpr bool __stop_callback_is_nothrow_constructible = __nothrow_constructible_from<
+      typename stop_token_of_t<_Env>::template callback_type<__future_stop_callback>,
```
ericniebler:
```suggestion
      stop_callback_for_t<stop_token_of_t<_Env>, __future_stop_callback>,
```

include/stdexec/__detail/__spawn_future.hpp
```diff
@@ -0,0 +1,756 @@
...
+      : __spawn_future_state_base<
+          // the spec says completion_signatures_of_t<__future_spawned_sender<_Sender, _Env>> but
+          // that breaks with an inscrutable error for _Sender = starts_on(sched, just() | then(...))
+          //
+          // I managed to fix the break by adding an extra _Env to the query, like so:
+          //
+          //     completion_signatures_of_t<__future_spawned_sender<_Sender, _Env>, _Env>
+          //
+          // but that's hard to justify--the future-spawned-sender will be connected to a receiver
+          // with an empty environment after all. This code works and seems sensible (neither
```
ericniebler: then it should be `completion_signatures_of_t<__future_spawned_sender<_Sender, _Env>, env<>>`, no?

include/stdexec/__detail/__spawn_future.hpp
```diff
@@ -0,0 +1,756 @@
...
+        }
+      }
+
+      void __try_cancel() noexcept override {
+        // Consider: there's a sense in which we only need to invoke request_stop if we've arrived
+        //           here before the producer has invoked __complete so I wonder whether it's possible
+        //           to avoid invoking request_stop when it's unnecessary. It might be.
+        //
+        //           We *must* invoke request_stop before returning from the function if our CAS
+        //           succeeds (because we arrived before either __complete or __coneume), or if the
```
ericniebler:
```suggestion
        //           succeeds (because we arrived before either __complete or __consume), or if the
```

include/stdexec/__detail/__spawn_future.hpp
```diff
@@ -0,0 +1,756 @@
...
+          traits::destroy(alloc, this);
+          traits::deallocate(alloc, this, 1);
+        }
+      }
+
+      // NOTE: __rcvr's type is unconstrained because the thing we pass doesn't satisfy receiver
+      void __do_consume(auto& __rcvr) noexcept {
+        using __variant_t = decltype(this->__result_);
+
+        __variant_t::visit(
```
ericniebler:
```suggestion
        __visit(
```

include/stdexec/__detail/__spawn_future.hpp
```diff
@@ -0,0 +1,756 @@
...
+            alloc,
+            op,
+            alloc,
+            static_cast<_Sender&&>(__sndr),
+            static_cast<_Token&&>(__tkn),
+            __spawn_common::__choose_senv(__env, get_env(__sndr)));
+        } catch (...) {
+          traits::deallocate(alloc, op, 1);
+          throw;
+        }
```
ericniebler: god i hate having to deal with allocators. scope guard here instead of try/catch.

include/stdexec/__detail/__spawn_future.hpp
```diff
@@ -0,0 +1,756 @@
...
+        return __make_sexpr<spawn_future_t>(std::unique_ptr<spawn_future_state_t, abandoner>(op));
+      }
+    };
+
+    template <class _Sender>
+    struct __future_operation_base {
+      // __data_of<_Sender> is a unique_ptr specialization
+      using __unique_ptr_t = __data_of<std::remove_cvref_t<_Sender>>;
+
+      __future_operation_base(__unique_ptr_t&& __future) noexcept
```
ericniebler: explicit

include/stdexec/__detail/__stop_when.hpp
```diff
@@ -0,0 +1,178 @@
...
+      };
+
+      static constexpr auto get_state =
+        []<class _Self, class _Receiver>(_Self&& __self, _Receiver& __rcvr) noexcept {
+          return __apply(
+            [&__rcvr](__ignore, auto&& token, __ignore) noexcept {
+              return __make_token_fn{}(
+                static_cast<decltype(token)>(token), get_stop_token(STDEXEC::get_env(__rcvr)));
+            },
+            static_cast<_Self&&>(__self));
```
ericniebler: i think this can now be (untested):
```c++
auto &[__tag, __token, __child] = __self;
return __make_token_fn{}(__forward_like<_Self>(__token),
                         get_stop_token(STDEXEC::get_env(__rcvr)));
```

include/stdexec/__detail/__associate.hpp
ispeters: This has been written into the spec; I think it's required unless you propose we change the wording.

include/stdexec/__detail/__associate.hpp
ispeters: What if I write `just(std::list<int>{})`? I believe that, at least in some STLs, `std::list` doesn't have a no-throw move constructor….

include/stdexec/__detail/__associate.hpp
ispeters: Tomasz tells me that `unique_ptr` doesn't invoke its deleter on `nullptr` and [the docs at cppreference](https://en.cppreference.com/w/cpp/memory/unique_ptr/~unique_ptr.html) agree.

include/stdexec/__detail/__spawn.hpp
ispeters: Yeah, I didn't get that far, yet.

include/stdexec/__detail/__counting_scopes.hpp
ispeters: The proposed change breaks a test that requires that the join-sender have a scheduler in its receiver's environment. Keeping the deduced return type (which I expect to SFINAE away this function when appropriate) and adding the call to `get_completion_signatures` brings the test back to passing. Is that expected?

ispeters:
Ok, I think this is correct-ish on top of a very recent `main`. There's still work to do, but I need to share the laptop. More tomorrow.

include/stdexec/__detail/__counting_scopes.hpp
ericniebler: i have updated my pr to show how to report errors from `get_completion_signatures`. https://github.com/ispeters/stdexec/pull/1/changes

include/stdexec/__detail/__associate.hpp
```diff
@@ -0,0 +1,253 @@
...
+
+      template <class _Sender, class... _Env>
+      static consteval auto get_completion_signatures() //
+        -> transform_completion_signatures<
+          __completion_signatures_of_t<__wrap_sender_of_t<_Sender>, _Env...>,
+          completion_signatures<set_stopped_t()>
+        > {
+        static_assert(sender_expr_for<_Sender, associate_t>);
+        return {};
+      };
```
ispeters, line 222: @ericniebler, does this need updating?

include/stdexec/__detail/__spawn_future.hpp
```diff
@@ -0,0 +1,757 @@
...
+    template <class _Completions>
+    struct __spawn_future_state_base;
+
+    template <class... _Sigs>
+    struct __spawn_future_state_base<completion_signatures<_Sigs...>> : __try_cancelable {
+      using __variant_t = __future_variant_t<_Sigs...>;
+      template <class _Env>
+      using __completions_t = __future_completions_t<_Env, _Sigs...>;
+
+      __variant_t __result_{__no_init};
```
ispeters, line 160: Given that `__variant_t` is initialized to a "no-init" state, the `__monostate` entry may not be pulling its weight; perhaps I should remove it and use the no-init state to mean "not completed, yet".

include/stdexec/__detail/__spawn_future.hpp
```diff
@@ -0,0 +1,757 @@
...
+      using __spawn_future_state_t = __unique_ptr_t<_Sender>::element_type;
+
+      template <class _Sender, class _Env>
+      using __completions_t = __spawn_future_state_t<_Sender>::template __completions_t<_Env>;
+
+      template <class _Sender, class _Env>
+      static consteval auto get_completion_signatures() //
+        -> __completions_t<_Sender, _Env> {
+        return {};
+      };
```
ispeters, line 743: @ericniebler, does this need updating?

include/stdexec/__detail/__stop_when.hpp
```diff
@@ -0,0 +1,185 @@
...
+      }
+    };
+
+    struct __stop_when_impl : __sexpr_defaults {
+      template <class _Sender, class... _Env>
+      static consteval auto get_completion_signatures() //
+        -> __completion_signatures_of_t<__child_of<_Sender>, _Env...> {
+        static_assert(sender_expr_for<_Sender, __stop_when_t>);
+        return {};
+      };
```
ispeters, line 68: And this one?

include/stdexec/__detail/__spawn_future.hpp
ericniebler: sure

include/stdexec/__detail/__associate.hpp
ericniebler: yes, but that can be a separate pr. same answer for the others, btw.

include/stdexec/__detail/__spawn_future.hpp
ispeters: That builds, but why is `completion_signatures_of_t<Foo>` different from `completion_signatures_of_t<Foo, env<>>`?

ispeters:
There are still a few things I'd like to improve, but this is probably landable.

PR #1715, opened 2025-12-16
let_value, _error, & _stopped: Destroy Child Operation State After Completion

RobertLeahy:
Previously the operation states for let_value, let_error, and let_stopped directly contained the operation states of one or two child operations. When first connected this number was one, and after the completion of that first operation and the connection of the sender returned by the wrapped invocable the number was two. Notably when the second operation state is connected the first operation has completed.

[P3373](https://isocpp.org/files/papers/P3373R2.pdf) proposes changing the above behavior such that after the tuple of values sent by the first operation is populated the operation state for the first operation is destroyed. Thereafter that storage can be reused for the operation state of the second operation thereby:

- Release resources held by the first operation state sooner, and
- Reducing the size of the operation states of let_value, let_error, and let_stopped

This commit implements the above-described change.

### DIFF
--- a/include/stdexec/__detail/__let.hpp
+++ b/include/stdexec/__detail/__let.hpp
@@ -33,6 +33,7 @@
 #include "__variant.hpp"
 
 #include <exception>
+#include <type_traits>
 
 namespace stdexec {
   //////////////////////////////////////////////////////////////////////////////
@@ -40,6 +41,11 @@ namespace stdexec {
   namespace __let {
     template <class _SetTag>
     struct __let_t;
+
+    template <class _SetTag>
+    struct __let_tag {
+      using __t = _SetTag;
+    };
 
     template <class _SetTag>
     inline constexpr __mstring __in_which_let_msg{"In stdexec::let_value(Sender, Function)..."};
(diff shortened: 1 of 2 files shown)


include/stdexec/__detail/__let.hpp
```diff
@@ -282,32 +284,90 @@ namespace stdexec {
...
+            if constexpr (__nothrow) {
+              __impl();
+            } else {
+              STDEXEC_TRY {
+                __impl();
+              }
+              STDEXEC_CATCH_ALL {
+                ::stdexec::set_error(static_cast<_Receiver&&>(__rcvr), std::current_exception());
+              }
+            }
```
ericniebler: i learned from lewis that it is simpler to use `try` unconditionally, and only guard the `set_error` call in the catch block. the optimized codegen is identical.

```suggestion
            STDEXEC_TRY {
              auto& __tuple =
                __state.__args_.emplace_from(__tup::__mktuple, static_cast<_Args&&>(__args)...);
              auto&& __sender = __tuple.apply(static_cast<_Fun&&>(__state.__fun_), __tuple);
              __state.__storage_.template emplace<__monostate>();
              __second_rcvr_t __r{__rcvr, static_cast<__env2_t&&>(__state.__env2_)};
              auto& __op = __state.__storage_.template emplace<__submit_t>(
                static_cast<__sender_t&&>(__sender), static_cast<__second_rcvr_t&&>(__r));
              __op.submit(static_cast<__sender_t&&>(__sender), static_cast<__second_rcvr_t&&>(__r));
            }
            STDEXEC_CATCH_ALL {
              if constexpr (!__nothrow) {
                ::stdexec::set_error(static_cast<_Receiver&&>(__rcvr), std::current_exception());
              }
            }
```

include/stdexec/__detail/__let.hpp
```diff
@@ -504,6 +564,17 @@ namespace stdexec {
...
+    template <class _Sender, class _Fun>
+    struct __data_t {
+      _Sender __sndr;
+      _Fun __fun;
+    };
+
+    template <typename _Sender>
+    using __sender_of = decltype((std::declval<__data_of<_Sender>>().__sndr));
+    template <typename _Sender>
+    using __fun_of = decltype((std::declval<__data_of<_Sender>>().__fun));
```
ericniebler: microbenchmarks showed that this somewhat bizarre implementation of `declval` compiles about 2x faster than the standard one:

```c++
  template <class _Tp, bool _Noexcept = true>
  using __declfn_t = auto (*)() noexcept(_Noexcept) -> _Tp;

  template <class _Tp>
  extern __declfn_t<_Tp &&> __declval;
```


```suggestion
    template <typename _Sender>
    using __sender_of = decltype((__declval<__data_of<_Sender>>().__sndr));
    template <typename _Sender>
    using __fun_of = decltype((__declval<__data_of<_Sender>>().__fun));
```

include/stdexec/__detail/__let.hpp
```diff
@@ -512,7 +583,7 @@ namespace stdexec {
       template <sender _Sender, __movable_value _Fun>
       auto operator()(_Sender&& __sndr, _Fun __fun) const -> __well_formed_sender auto {
         return __make_sexpr<__let_t<_SetTag>>(
-          static_cast<_Fun&&>(__fun), static_cast<_Sender&&>(__sndr));
+          __data_t{static_cast<_Sender&&>(__sndr), static_cast<_Fun&&>(__fun)});
```
ericniebler: this is a problem. the structured binding of `let_value` is part of its interface. this should compile:

```c++
auto [tag, fn_copy, sndr_copy] = stdexec::let_value(sndr, fn);
```

you can use `transform_sender` to swivel the `fn` and the `sndr` into a `__data_t` object. `exec::repeat_n` does this, so you can crib from that.

include/stdexec/__detail/__let.hpp
```diff
@@ -524,114 +595,50 @@ namespace stdexec {
...
           _Child,
           env_of_t<_Receiver>,
           __q<__decayed_tuple>,
-          __mk_let_state
-        >;
-
-        return __sndr.apply(
-          static_cast<_Sender&&>(__sndr),
-          [&]<class _Fn, class _Child>(__ignore, _Fn&& __fn, _Child&& __child) {
-            // TODO(ericniebler): this needs a fallback
```
ericniebler, line 568: please don't loose this comment. i need to come back to this.

ericniebler (dismissed):

include/stdexec/__detail/__let.hpp
```diff
@@ -282,32 +284,90 @@ namespace stdexec {
...
     //! The core of the operation state for `let_*`.
     //! This gets bundled up into a larger operation state (`__detail::__op_state<...>`).
-    template <class _Receiver, class _Fun, class _SetTag, class _Env2, class... _Tuples>
+    template <class _SetTag, class _Sender, class _Fun, class _Receiver, class... _Tuples>
     struct __let_state {
-      using __fun_t = _Fun;
-      using __env2_t = _Env2;
-      using __env_t = __join_env_t<_Env2, env_of_t<_Receiver>>;
-      using __rcvr_t = __receiver_with_env_t<_Receiver, _Env2>;
+      using __env2_t = __env2_t<_SetTag, env_of_t<const _Sender&>, env_of_t<const _Receiver&>>;
```
ericniebler:
```suggestion
      using __env2_t = __let::__env2_t<_SetTag, env_of_t<const _Sender&>, env_of_t<const _Receiver&>>;
```

include/stdexec/__detail/__let.hpp
```diff
@@ -639,5 +657,19 @@ namespace stdexec {
...
+    static constexpr auto get_attrs = []<class _Child>(__ignore, const _Child& __child) noexcept {
+      // BUGBUG:
+      return stdexec::get_env(__child);
+      //return __attrs<__let_t<_SetTag>, _Child, _Fun>{};
+    };
+
+    static constexpr auto get_completion_signatures =
+      []<class _Sender, class... _Env>(_Sender&&, const _Env&...) noexcept
+      -> __completion_signatures_of_t<transform_sender_result_t<_Sender, _Env...>, _Env...> {
+    };
```
ericniebler: why are these needed?

RobertLeahy: `get_attrs` is not, I put it there to preserve the comments since you didn't want what I thought was a similar comment lost elsewhere.

The bottom one may actually be unnecessary, I cribbed it from `repeat_n`, I'll check.

RobertLeahy: It's definitely necessary to have the `get_completion_signatures`, this is the default implementation thereof:

```c++
static constexpr auto get_completion_signatures =
  []<class _Sender>(_Sender&&, auto&&...) noexcept {
    static_assert(
      __mnever<tag_of_t<_Sender>>,
      "No customization of get_completion_signatures for this sender tag type.");
  };
  ```

I can yank `get_attrs` if you're happy to lose the comment.

ericniebler: Yes please. I didn't see that the comment had been moved, not removed.

PR #1718, opened 2025-12-17
Fix: when_all cancelled from environment

sithhell:
In the case when when_all is being stopped from a stop source in the environment, the stop_source was triggered multiple times leading to a use after free bug.

This patch fixes this by performing the following steps:

 - Inside the stop callback, increment the reference count.
 - Try setting the state to stopped, only if stopped, trigger the internal stop_source which cancels the children.
 - Arrive to decrement the count again and handle any concurrent completions if needed.

### DIFF
--- a/include/stdexec/__detail/__when_all.hpp
+++ b/include/stdexec/__detail/__when_all.hpp
@@ -188,7 +188,7 @@ namespace stdexec {
       void operator()() const noexcept {
         // Temporarily increment the count to avoid concurrent/recursive arrivals to
         // pull the rug under our feet. Relaxed memory order is fine here.
-        __state_->__count_.fetch_add(1, std::memory_order_relaxed);
+        __state_->__count_.fetch_add(1, __std::memory_order_relaxed);
 
         __state_t __expected = __started;
         // Transition to the "stopped" state if and only if we're in the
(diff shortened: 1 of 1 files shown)


include/stdexec/__detail/__when_all.hpp
```diff
@@ -183,18 +183,42 @@ namespace stdexec {
 
     struct _INVALID_ARGUMENTS_TO_WHEN_ALL_ { };
 
-    template <class _ErrorsVariant, class _ValuesTuple, class _StopToken, bool _SendsStopped>
+    template <class _State, class _Receiver>
+    struct __forward_stop_request {
+      void operator()() const noexcept {
+        // Temporarily increment the count to avoid concurrent/recursive arrivals to
+        // pull the rug under our feet. Relaxed memory order is fine here.
+        __state_->__count_.fetch_add(1, std::memory_order_relaxed);
```
ericniebler:
```suggestion
        __state_->__count_.fetch_add(1, __std::memory_order_relaxed);
```

include/stdexec/__detail/__when_all.hpp
```diff
@@ -183,18 +183,42 @@ namespace stdexec {
...
-      using __stop_callback_t = stop_callback_for_t<_StopToken, __forward_stop_request>;
+      using __stop_callback_t = stop_callback_for_t<
+        stop_token_of_t<env_of_t<_Receiver>>,
+        __forward_stop_request<__when_all_state, _Receiver>
+      >;
 
-      template <class _Receiver>
       void __arrive(_Receiver& __rcvr) noexcept {
-        if (1 == __count_.fetch_sub(1)) {
+        if (1 == __count_.fetch_sub(1, std::memory_order_acq_rel)) {
```
ericniebler:
```suggestion
        if (1 == __count_.fetch_sub(1, __std::memory_order_acq_rel)) {
```

PR #1730, opened 2026-01-05
minor: align the target name in conanfile with the document

kexianda:


conanfile.py
```diff
@@ -57,4 +57,5 @@ def package(self):
   def package_info(self):
     self.cpp_info.set_property("cmake_file_name", "P2300")
     self.cpp_info.set_property("cmake_target_name", "P2300::P2300")
+    self.cpp_info.set_property("cmake_target_aliases", ["STDEXEC::stdexec"])
```
ericniebler, line 60: i'm no conan expert myself, but what do the two lines above yours do? there is no "P2300" file or "P2300::P2300" target.

ericniebler: attn @kexianda

kexianda: @ericniebler  Sorry for late replay. 

It is a minor issue.  Let me explain:

1,  `self.cpp_info.set_property("cmake_file_name", "P2300")`
     `self.cpp_info.set_property("cmake_target_name", "P2300::P2300")`
   
The cmake target name of `execution`  is `STDEXEC::stdexec`.  It was declared as `P2300::P2300` in the conanfile years ago.  Conan will generate a Find-P2300.cmake, wiht the target name `P2300::P2300`.

2,  `self.cpp_info.set_property("cmake_target_aliases", ["STDEXEC::stdexec"])`
Adding an alias(`STDEXEC::stdexec`) of cmake target for `execution`

PR #1737, opened 2026-01-07
Add CPU (gcc 13, Debug) configuration to CI

romintomasetti:
I'm seeing issues with later GCC compilers. So let's increase the coverage.


ericniebler:
thanks!

PR #1752, opened 2026-01-11
fix bwos queue bug

viktorika:
1.Running TSAN on my M3 computer reports a data race. I believe there are some memory ordering issues with thief_block_, tail_, and steal_tail_.
2.There is an extremely small chance that elements within a certain block might be skipped and never executed. The reason is that the steal function uses steal_tail to determine whether it is complete or empty, but steal_tail == block_size might indicate that the current block has already been taken over.

`std::uint64_t spos = steal_head_.load(stdexec::__std::memory_order_acquire);`

Using steal_head for checking can avoid this issue.

### DIFF
--- a/include/exec/__detail/__bwos_lifo_queue.hpp
+++ b/include/exec/__detail/__bwos_lifo_queue.hpp
@@ -178,7 +178,7 @@ namespace exec::bwos {
   auto lifo_queue<Tp, Allocator>::steal_front() noexcept -> Tp {
     std::size_t thief = 0;
     do {
-      thief = thief_block_.load(stdexec::__std::memory_order_relaxed);
+      thief = thief_block_.load(stdexec::__std::memory_order_acquire);
       std::size_t thief_index = thief & mask_;
       block_type &block = blocks_[thief_index];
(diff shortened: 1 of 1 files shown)


ericniebler:
@maikel could you take a look please?

viktorika:
Please check this issue. I have described the problem in more detail.https://github.com/NVIDIA/stdexec/issues/1755

maikel:
@viktorika Does #1967 fix your issue?

viktorika:
> @viktorika Does #1967 fix your issue? #1967 解决你的问题了吗？

@maikel 
I believe this implementation is correct, but does it create any efficiency issues if the thief always needs to start from start_block_ to find a block to steal?

maikel:
That difference would need to be measured. I don't think that its a big issue, but intuition is often wrong here. 

More imprtantly is, that we have correct version first. Dropping tasks was a big nogo

ericniebler:
what is the status of this pr? is a fix still needed here? attn: @viktorika @maikel

maikel:
I believe my other changes a month ago have fixed this issue here

PR #1753, opened 2026-01-11
stdexec::when_all: Nothrow Connect

RobertLeahy:
get_state was missing noexcept and therefore stdexec::when_all senders were never nothrow connectable.

This defect is present in std::execution and addressing it is LWG4502.

### DIFF
--- a/include/stdexec/__detail/__when_all.hpp
+++ b/include/stdexec/__detail/__when_all.hpp
@@ -310,7 +310,7 @@ namespace stdexec {
     template <class _Receiver>
     static auto __mk_state_fn(const _Receiver&) noexcept {
       using __env_of_t = env_of_t<_Receiver>;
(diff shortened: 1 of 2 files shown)


include/stdexec/__detail/__when_all.hpp
```diff
@@ -368,7 +368,7 @@ namespace stdexec {
       };
 
       static constexpr auto get_state =
-        []<class _Self, class _Receiver>(_Self&& __self, _Receiver& __rcvr)
+        []<class _Self, class _Receiver>(_Self&& __self, _Receiver& __rcvr) noexcept
```
ericniebler, line 371: could you also pls add `noexcept` to the lambda returned from `__when_all::__mk_state_fn`. thanks.

ericniebler:
thanks @RobertLeahy !

PR #1765, opened 2026-01-14
Fix data race in async_scope

lixin-wei:
This PR fixes a subtle data race in `async_scope` I introduced in https://github.com/NVIDIA/stdexec/commit/a61207c28dd8bab294e8b144e05253adc337699b.

The race occurs between `__nest_rcvr::__complete()` and `__when_empty_op::start()` (the `on_empty()` operation).

### The Problematic Code Flow

```
Thread T2 (__complete)           Thread T1 (on_empty)
─────────────────────           ───────────────────
fetch_sub: 1→0
                                lock mutex
                                load __active (sees 0!)
                                unlock
                                start continuation → destroy scope
lock mutex → USE-AFTER-FREE
```

# Solution

added another `__pending_notifiers_` counter to guard this case.

I've tested this in my heavy workflow with TSAN.

### DIFF
--- a/include/exec/async_scope.hpp
+++ b/include/exec/async_scope.hpp
@@ -51,6 +51,7 @@ namespace exec {
       inplace_stop_source __stop_source_{};
       mutable std::mutex __lock_{};
       mutable __std::atomic_ptrdiff_t __active_ = 0;
+      mutable __std::atomic_ptrdiff_t __pending_notifiers_ = 0;  // Track in-flight __complete() calls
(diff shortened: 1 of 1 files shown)


lixin-wei:
@ericniebler  I noticed you already had a fix in https://github.com/NVIDIA/stdexec/pull/1734, but this solution has better performance, not need to lock everytime in __complete, only the last one will.

Sorry for the trouble!

PR #1767, opened 2026-01-16
Fix compile error for tuple size > 8

lixin-wei:
Met a compile error when passing a lot of elements to `when_all`.

This PR fixes it.

### DIFF
--- a/include/stdexec/__detail/__tuple.hpp
+++ b/include/stdexec/__detail/__tuple.hpp
@@ -345,7 +345,11 @@ namespace STDEXEC {
     constexpr auto __size = STDEXEC_REMOVE_REFERENCE(_Tuple)::__size;
     static_assert(_Index < __size, "Index out of bounds in __get");
 
-    if constexpr (_Index == 0) {
+    // Only use __valN accessors for tuples with <= 8 elements (which have specialized storage)
+    // Larger tuples use __box base class and need __tup::__get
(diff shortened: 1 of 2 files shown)


include/stdexec/__detail/__tuple.hpp
```diff
@@ -345,7 +345,11 @@ namespace STDEXEC {
     constexpr auto __size = STDEXEC_REMOVE_REFERENCE(_Tuple)::__size;
     static_assert(_Index < __size, "Index out of bounds in __get");
 
-    if constexpr (_Index == 0) {
+    // Only use __valN accessors for tuples with <= 8 elements (which have specialized storage)
+    // Larger tuples use __box base class and need __tup::__get
+    if constexpr (__size > 8) {
+      return __tup::__get<_Index>(static_cast<_Tuple&&>(__tupl));
```
ericniebler, line 351: i'm curious why you moved this branch to the top. i expect tuples larger than 8 elements to be quite rare, which is why i put the fallback branch last.

lixin-wei: because tuples larger than 8 elements don't have `__val` fields.

if we put this to the end, the compile will fail.

ericniebler: oh right, this is `__get`, not `__apply`. :facepalm:

PR #1771, opened 2026-01-20
stdexec::just, ::just_error, & ::just_stopped: constexpr

RobertLeahy:
Includes other changes necessary to support the above-listed algorithms being usable in a constexpr context.


ericniebler:
i've rebased #1772 on your commit. i'll land it there.

PR #1777, opened 2026-01-22
revert "Fix data race in async_scope"

lixin-wei:
Sorry I'm here again🤦, this PR reverts https://github.com/NVIDIA/stdexec/commit/58a2a918f2738127abc64e00ba9e2bae7ec4332b, because there is still a data race when subtracting `__pending_notifiers_` in the `else` branch of `__complete()`, where the scope might already be destroyed.

also, according to the standard [exec.scope#simple.counting.general](https://eel.is/c++draft/exec.scope#simple.counting.general-1):

> For purposes of determining the existence of a data race, get_token, close, join, try-associate, disassociate, and start-join-sender behave as **atomic operations**. These operations on a single object of type simple_counting_scope appear to occur in a single total order.

IIUC it's impossible to impl a lock-free `counting_scope`. Beacuse at least start-join-sender can't be lock-free, then in order to make others atomic, all operations should share the same lock.

-----

If it can't be lock-free, I wonder in C++26, what's the recommended way to async start a sender in a hot code path, without suffering from the lock and memory allocatioin?

The best way I can think up is: make a dummy scope which ignores everything + custom allocator.

```cpp
ex::spawn(sender, dummy_scope.get_token(), my_allocator);
```

I wonder whether it's a good practice, is there any better way?

### DIFF
--- a/include/exec/async_scope.hpp
+++ b/include/exec/async_scope.hpp
@@ -51,8 +51,6 @@ namespace exec {
       inplace_stop_source __stop_source_{};
       mutable std::mutex __lock_{};
       mutable __std::atomic_ptrdiff_t __active_ = 0;
-      mutable __std::atomic_ptrdiff_t __pending_notifiers_ =
-        0; // Track in-flight __complete() calls
       mutable __intrusive_queue<&__task::__next_> __waiters_{};
 
       ~__impl() {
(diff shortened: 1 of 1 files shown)


lixin-wei:
Sorry a typo caused compile failure, fixed.

ccotter:
> Sorry I'm here again🤦, this PR reverts [58a2a91](https://github.com/NVIDIA/stdexec/commit/58a2a918f2738127abc64e00ba9e2bae7ec4332b), because there is still a data race when subtracting `__pending_notifiers_` in the `else` branch of `__complete()`, where the scope might already be destroyed.

btw, @lixin-wei - do you have a test that reproduces this? If you are able to, it'd be great if you can provide either a small test case that hits the race, and/or provide the test case as a Relacy (https://github.com/NVIDIA/stdexec/tree/main/test/rrd#relacy-tests) test. I've found Relacy to be great at finding rare thread interleavings.

PR #1780, opened 2026-01-23
Enable compile-time evaluation of just senders

romintomasetti:
### Motivation

C++23's expanded `constexpr` support enables compile-time computations with complex types, but `stdexec` currently lacks the annotations needed to use sender/receiver patterns in constant expressions.

### Changes

This PR adds `constexpr` annotations to make `just` constructible and connectable at compile time.

See the added test for usage.

> I've tried to make `then` work too, so that at least some interesting sender can be constructed that really does some work, but there were too many issues I couldn't fix myself.


romintomasetti:
@ericniebler It might be a first step towards yours (#1772).

RobertLeahy:
Rationale for closing #1771 seems to apply here as well.

ericniebler:
i have managed to land #1772, which subsumes this pr

PR #1797, opened 2026-01-27
`repeat_*` noexcept fixes

justend29:
My original goal was to conditionally add `set_error_t(std::exception_ptr)` to the completion signatures of the `repeat_*` senders, only to realize that exceptions from `connect()` was one of the locations `set_error()` could be called, where the nothrow-ness of `connect()` is only discernible once a receiver is known (after advertising `completion_signatures`)

Nevertheless, on my travels, I learned a great deal about the meta-programming in this library and made some fixes:

1. repeat state base classes are abstract but were missing virtual destructors -> made them protected
2. repeat state classes were missing virtual destructors -> made the classes final
3. add `noexcept` specifiers on `__connect()` and propagate to the operation states

### DIFF
--- a/include/exec/completion_signatures.hpp
+++ b/include/exec/completion_signatures.hpp
@@ -61,7 +61,7 @@ namespace exec {
   //! @tparam DeducedSigs Completion signature types to be deduced from the function arguments.
   //! @param unnamed Pointer arguments (unused) used for type deduction of DeducedSigs.
   //!
-  //! @return An instance of `detail::make_completion_signatures_t` containing the combined
+  //! @return An instance of `STDEXEC::completion_signatures` containing the combined
   //!         signatures.
   //!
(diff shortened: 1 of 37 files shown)


RobertLeahy:
> where the nothrow-ness of `connect()` is only discernible once a receiver is known (after advertising `completion_signatures`)

Note that this is the topic of [P3373](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2025/p3388r3.pdf) which is part of C++26.

You can use `::STDEXEC::__receiver_archetype<Env>` to check whether or not connect will throw without a receiver.

justend29:
You are literally the mystical wizard with all the answers. I'll review the provided info and take another stab at the original goal. Thanks, @RobertLeahy!

justend29:
Original goal lives on https://github.com/NVIDIA/stdexec/pull/1799

PR #1812, opened 2026-02-04
Fix race in MPSC algo

ccotter:
Relacy helped identify a race in the existing MPSC algo. I am having a hard time exactly explaining what's going on, but in the newly added unit test (the non Relacy one), I am able to observe three different odd behaviors

 - a consumer consuming the same elemment in an finite loop, apparently due to the internal next pointers pointing in some sort of cycle
 - consumer returning &__nil_!
 - consumer never able to consume a produced value (node is lost)

With the non-relacy unit test, in the existing algo, if I insert a random sleep of 0-10 microseconds in push_back after __back_ is exchanged, I can observe one of the above behaviors nearly every single time. The most common was the first behavior.

The existing algo claims it came from Dmitry Vyukov's implementation, though one key difference is that the existing one uses an atomic pointer to a Node for the "nil" object, whereas Dmitry's stores an actual Node object embedded in the queue.

I re-implemented the version in stdexec exactly as it appears on Dmitry's website (which I had to dig up on archive.org), and it passes newly added Relacy (exploring many thread interleavings) and non-Relacy unit tests.

I originally tracked down a bug in timed_thread_scheduler.cpp, where sometimes `STDEXEC_ASSERT(op->command_ == command_type::command_type::stop);` failed.

### DIFF
--- a/include/stdexec/__detail/__stop_token.hpp
+++ b/include/stdexec/__detail/__stop_token.hpp
@@ -90,7 +90,7 @@ STDEXEC_P2300_NAMESPACE_BEGIN()
    private:
     struct __callback_type
     {
-      constexpr explicit __callback_type(never_stop_token, STDEXEC::__ignore) noexcept { }
+      constexpr explicit __callback_type(never_stop_token, STDEXEC::__ignore) noexcept {}
     };
    public:
     template <class>
--- a/test/rrd/intrusive_mpsc_queue.cpp
+++ b/test/rrd/intrusive_mpsc_queue.cpp
@@ -18,31 +18,41 @@
 
 #include <stdexec/__detail/__intrusive_mpsc_queue.hpp>
 
-struct test_node {
+struct test_node
+{
   std::atomic<test_node*> next_{nullptr};
-  int value_{0};
+  int                     value_{0};
 
   test_node() = default;
-  explicit test_node(int val) : value_(val) {}
+  explicit test_node(int val)
+    : value_(val)
+  {}
 };
 
 using test_queue = STDEXEC::__intrusive_mpsc_queue<&test_node::next_>;
 
-struct mpsc_single_producer_consumer : rl::test_suite<mpsc_single_producer_consumer, 2> {
+struct mpsc_single_producer_consumer : rl::test_suite<mpsc_single_producer_consumer, 2>
+{
   test_queue queue;
-  test_node node1{42};
-  test_node node2{100};
-  int consumed_count{0};
-  int values_sum{0};
+  test_node  node1{42};
+  test_node  node2{100};
+  int        consumed_count{0};
+  int        values_sum{0};
 
-  void thread(unsigned thread_id) {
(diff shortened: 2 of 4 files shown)


ccotter:
cc @maikel - can you check this over, since I think you implemented the original version of the MPSC queue?

ccotter:
Dmitry's website appears to have stopped working a few months ago, though the original link is on archive.org: https://web.archive.org/web/20221127081044/https://www.1024cores.net/home/lock-free-algorithms/queues/intrusive-mpsc-node-based-queue (Dmitry is also the author of Relacy).

ccotter:
Also of note, I originally noticed the race/bug in timed_thread_scheduler while running the exec unit tests through with a version of TSAN I am trying to upstream which fuzzes the scheduler with random sleeps, and this is another success of that approach I think: https://github.com/llvm/llvm-project/pull/178836

ccotter:
I caught up with @maikel. I'm converting this to a draft for now, as @maikel will take a closer look to understand if the existing algo can be fixed. We also suspect my version may be overly specified in the atomics (it appears to be correct, but could be more relaxed).

ccotter:
@maikel as an experiment to help narrow down where the race leading to the queue state becoming corrupt, I added a mutex in two places: https://gist.github.com/ccotter/37b0d3cb3ad88f8c4fea4b2adbb5dce7 ... Relacy says this queue is correct...

maikel:
Yes after revisiting the old code I agree that the management of nil is the culprit here. We should simplify the implementation towards dmitrys original one. The only thing I would check is whether your usage of memory orderings and atomics can be improved

ccotter:
My changes to build and run the relacy tests are failing because TBB ends up needing std::shared_mutex. My local build does not run with TBB, so I didn't encounter this. I will soon have a PR to upstream relacy to add support for std::shared_mutex, which should address this. If I can't get that in / merge in the next few days, I'll disable the relacy tests in this PR for the time being.

ccotter:
Ok, I believe this should be good to go. Do you mind squashing this as the individual commits are more noisy than valuable in the history.

Of note, I found two compiler crashes, one in gcc (not sure if the bugzilla reprot is in yet) and one in clang (https://github.com/llvm/llvm-project/issues/181088) while working on this!

ccotter:
@ericniebler bump? This should be ready for review/testing.

Also ... the custom changes to TSAN I've been using to try to shake out other bugs has just landed in upstream clang. If anyone has easy access to trunk builds, https://github.com/llvm/llvm-project/blob/main/clang/docs/ThreadSanitizer.rst#adaptive-delay describes the feature and various flags.

ericniebler:
@ccotter is it ok to squash this to a single commit?

ccotter:
Yes squash definitely. CI is passing now? I thought I had one more thing to fix for the relacy tests. We can disable the build for now if they are still not passing (I might need one more upstream change to relacy)

ericniebler:
i fixed the build problem.

PR #1810, opened 2026-02-02
exec(fork_join): make it customizable

romintomasetti:
This PR makes `exec::fork_join` customizable.

Related to:
- #1802

PR #1813, opened 2026-02-04
exec(fork_join): can be nested

romintomasetti:
This PR tests that `exec::fork_join` can be nested.


romintomasetti:
@ericniebler Could you please have a look ? I would rather merge this one before #1810. Thanks !

ericniebler (approved):

PR #1831, opened 2026-02-11
Move associate's opstate into the get_state function

ispeters:
This diff better aligns the implementation of `associate` with how it's worded in the spec; it moves the definition of the associate-sender's operation state type into the `get_state` function, making it a function-local class.

### DIFF
--- a/include/stdexec/__detail/__associate.hpp
+++ b/include/stdexec/__detail/__associate.hpp
@@ -147,59 +147,6 @@ namespace STDEXEC {
       }
     };
 
-    // NOTE: the spec declares this class template inside the get_state function
-    //       but I couldn't get that to build with Clang 21 so I moved it out to
-    //       this namespace-scoped template and __uglified all the symbols
-    template <class _Sender, class _Receiver>
-    struct __op_state {
-      using __associate_data_t = std::remove_cvref_t<__data_of<_Sender>>;
-      using __assoc_t = __associate_data_t::__assoc_t;
-      using __sender_ref_t = __associate_data_t::__sender_ref;
-
-      using __op_t = connect_result_t<typename __sender_ref_t::element_type, _Receiver>;
-
-      __assoc_t __assoc_;
-      union {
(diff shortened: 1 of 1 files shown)


include/stdexec/__detail/__associate.hpp
```diff
@@ -231,8 +178,56 @@ namespace STDEXEC {
...
+            union {
+              _Receiver* __rcvr_;
+              __op_t __op_;
+            };
+
+            explicit __op_state(std::pair<__assoc_t, __sender_ref_t> __parts, _Receiver& __rcvr)
+              : __assoc_(std::move(__parts.first)) {
+              if (__assoc_) {
+                ::new ((void*) std::addressof(__op_))
+                  __op_t(STDEXEC::connect(std::move(*__parts.second), std::move(__rcvr)));
```
ispeters, line 198: [The spec](https://eel.is/c++draft/exec#associate-11) doesn't fully-qualify this call to `connect`; it's required in stdexec to avoid an ambiguity with (I think) a member `connect` that's in scope here. I'm wondering if _`basic-sender`_ has the same ambiguity as `__sexpr_defaults`.

include/stdexec/__detail/__associate.hpp
```diff
@@ -231,8 +178,56 @@ namespace STDEXEC {
...
+
+            ~__op_state() {
+              if (__assoc_) {
+                std::destroy_at(std::addressof(__op_));
+              }
+            }
+
+            void __run() noexcept {
+              if (__assoc_) {
+                STDEXEC::start(__op_);
```
ispeters, line 223: This call to `start` has the same problem as the above call to `connect`.

ericniebler:
tbh i don't really want local classes for anything, ever. their mangled names are horrible and make diagnostics harder to read. i only used them in P2300 to make the spec more terse. my hope is that library implementers would liberally apply the AS-IF rule and use different implementation strategies.

Issue #1836, opened 2026-02-13
asioexec senders no longer conform to sender_of concepts


jessmorecroft:
It looks like some recent changes may have broken asioexec's adherence to sender concepts.

Specifically this code,

```
inline stdexec::sender_of<stdexec::set_value_t(std::size_t)> auto
Write(boost::asio::ip::tcp::socket& socket, std::string_view buffer)
{
    return boost::asio::async_write(socket, boost::asio::buffer(buffer),
                                    asioexec::use_sender);
}
```

No longer compiles. I get this:

```
Deduced type 'decltype(async_initiate<const asioexec::use_sender_t &, void (boost::system::error_code, std::size_t)>(declval<detail::initiate_async_write<boost::asio::basic_stream_socket<boost::asio::ip::tcp, boost::asio::any_io_executor>>>(), token, buffers, transfer_all()))' (aka 'sender<asioexec::detail::completion_token::sender<stdexec::completion_signatures<stdexec::set_value_t (boost::system::error_code, unsigned long), stdexec::set_error_t (std::exception_ptr), stdexec::set_stopped_t ()>, boost::asio::detail::initiate_async_write<boost::asio::basic_stream_socket<boost::asio::ip::tcp, boost::asio::any_io_executor>>, const boost::asio::const_buffer &, boost::asio::detail::transfer_all_t>>') does not satisfy 'sender_of<stdexec::set_value_t (std::size_t)>'clang(placeholder_constraints_not_satisfied)
__senders.hpp(43, 23): Because 'asioexec::detail::use_sender::sender<asioexec::detail::completion_token::sender<stdexec::completion_signatures<stdexec::set_value_t (boost::system::error_code, unsigned long), stdexec::set_error_t (std::exception_ptr), stdexec::set_stopped_t ()>, boost::asio::detail::initiate_async_write<boost::asio::basic_stream_socket<boost::asio::ip::tcp>>, const boost::asio::const_buffer &, boost::asio::detail::transfer_all_t>>' does not satisfy 'sender_in'
```

I'll take a closer look tomorrow to see if I can figure it out.

RobertLeahy:
Note that `sender_of` is exposition only in the current working draft.

Recently (f797a6555) `stdexec::sender_of` was changed from having a third defaulted template parameter:

```c++
template <class _Sender, class _SetSig, class _Env = env<>>
concept sender_of = sender_in<_Sender, _Env>
```

To having a trailing variadic pack:

```c++
[...]

PR #1838, opened 2026-02-13
asioexec::completion_token & ::use_sender: Non-Dependent Senders

RobertLeahy:
asioexec::completion_token and ::use_sender are completion tokens which induce an initiating function to return a sender. Previously these senders were dependent, but this was not necessary: The completion signatures thereof are not changed by the environment. Updated the implementation so that the senders yielded by initiating functions parameterized by asioexec::completion_token or ::use_sender are no longer dependent.

Addresses #1836.

### DIFF
--- a/include/asioexec/use_sender.hpp
+++ b/include/asioexec/use_sender.hpp
@@ -158,15 +158,10 @@ namespace asioexec {
 
       using sender_concept = ::STDEXEC::sender_t;
 
-      template <typename Env>
-      constexpr completion_signatures<::STDEXEC::completion_signatures_of_t<Sender, Env>>
-        get_completion_signatures(const Env&) && noexcept {
-        return {};
-      }
-
-      template <typename Env>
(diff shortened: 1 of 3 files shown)


include/asioexec/use_sender.hpp
```diff
@@ -158,15 +158,11 @@ namespace asioexec {
...
-        get_completion_signatures(const Env&) && noexcept {
-        return {};
-      }
-
-      template <typename Env>
-      constexpr completion_signatures<::STDEXEC::completion_signatures_of_t<const Sender&, Env>>
-        get_completion_signatures(const Env&) const & noexcept {
+      template <typename Self, typename... Env>
+      static consteval completion_signatures<::STDEXEC::completion_signatures_of_t<
+        decltype(::STDEXEC::__forward_like<Self>(std::declval<Sender&>())),
```
ericniebler: this could be
```suggestion
        ::STDEXEC::__copy_cvref_t<Self, Sender>,
```

PR #1841, opened 2026-02-13
move `parallel_scheduler` into namespace `STDEXEC`

ericniebler:


test/exec/CMakeLists.txt
```diff
@@ -64,7 +64,6 @@ set(exec_test_sources
     $<$<BOOL:${STDEXEC_ENABLE_TBB}>:../execpools/test_tbb_thread_pool.cpp>
     $<$<BOOL:${STDEXEC_ENABLE_TASKFLOW}>:../execpools/test_taskflow_thread_pool.cpp>
     $<$<BOOL:${STDEXEC_ENABLE_ASIO}>:../execpools/test_asio_thread_pool.cpp>
-    test_system_context.cpp
```
romintomasetti, line 67: @ericniebler It seems that `test_system_context.cpp` has not been deleted, but I can't find any `CMake` target that uses it.

Issue #1849, opened 2026-02-15
`when_all` hangs in finally when triggered by stop


jessmorecroft:
This is a strange one - when_all doesn't seem to complete properly when used in finally triggered by a stop, as shown here:

```c++
    execpools::asio_thread_pool pool(1);

    boost::asio::steady_timer timer1(pool.get_executor());
    boost::asio::steady_timer timer2(pool.get_executor());
    timer1.expires_after(std::chrono::seconds(2));
    timer2.expires_after(std::chrono::seconds(1));

    auto result = sync_wait(starts_on(
        pool.get_scheduler(),
        when_any(
            timer1.async_wait(asioexec::use_sender) | then([] { return 41; }) |
                finally(sequence(
                    just() | then([] { fmt::println("before when_all"); }),
                    when_all(
                        just() | then([] { fmt::println("in finally 1"); }),
                        just() | then([] { fmt::println("in finally 2"); })),
                    just() | then([] { fmt::println("after when_all"); }))),
            timer2.async_wait(asioexec::use_sender) |
                then([] { return 42; }))));

    fmt::println("result: {}", result);
```
This outputs:
```
before when_all
in finally 1
in finally 2
(hangs)
```

If I change this to swap the timer durations, so our finally will be invoked as a consequence of a non-stopped completion, it completes as expected:
```
.....
    timer1.expires_after(std::chrono::seconds(1));
    timer2.expires_after(std::chrono::seconds(2));
.....
```
Output:
```
before when_all
in finally 1
in finally 2
after when_all
result: optional((41))
```
I have been trawling finally and when_all but can't figure it out. Is my usage wrong in some way perhaps..?

ericniebler:
interesting. i can confirm the hang with the asio thread pool. i ported the code to use the `io_uring_context` and found that the problem went away.

```c++
    exec::io_uring_context pool;
    auto sch = pool.get_scheduler();

    auto result = ex::sync_wait(
      ex::starts_on(
        pool.get_scheduler(),
        exec::when_any(
          exec::schedule_after(sch, std::chrono::seconds(2)) //
            | ex::then([] { return 41; })
            | exec::finally(
              exec::sequence(
                ex::just() | ex::then([] { std::puts("before when_all"); }),
                ex::when_all(
                  ex::just() | ex::then([] { std::puts("in finally 1"); }),
                  ex::just() | ex::then([] { std::puts("in finally 2"); })),
[...]

PR #1851, opened 2026-02-16
stdexec::when_all: Hang After Stop Request

RobertLeahy:
std::execution::when_all forwards stop requests through to its children via the exposition-only class on-stop-request, defined as follows (see [exec.snd.general]):

```
struct on-stop-request {
  inplace_stop_source& stop-src;       // exposition only
  void operator()() noexcept { stop-src.request_stop(); }
};
```

stdexec::when_all used to be conforming in this regard, however this changed in 36a92fd776c835abd4dc5e62d43cf040c20a9add. That commit not only complicates the stop callback used by stdexec::when_all but also causes the __stopped state to be reachable even when no child is capable of sending a stopped completion signal. The change to std::execution:: when_all introduced by P3887 (see 712953b245412bf8ebdfdf369136d637beb43aec) depends on the __stopped state being impossible unless a child sends a stopped completion signal. This tension meant that 36a92fd776c835abd4dc5e62d43cf040c20a9add caused stdexec::when_all to not complete (thereby leading to hangs) when:

- No child potentially sent a stopped completion signal, and
- Stop was requested

712953b245412bf8ebdfdf369136d637beb43aec intended to include a unit test preventing the above-described regression but due to a typo (corrected in this commit) that unit test failed in this regard.

Note that 36a92fd776c835abd4dc5e62d43cf040c20a9add is a misdiagnosis and should eventually be reverted (along with this commit save the above- mentioned unit test correction). The issue that
36a92fd776c835abd4dc5e62d43cf040c20a9add aims to address has nothing to do with stdexec::when_all and is instead caused by the fact that stdexec::inplace_stop_source is not well-behaved when the lifetime of the stop source ends during the dispatch of the final stop callback. This issue should be fixed separately and this commit is only intended to workaround the issue in the meantime.

Addresses #1849, fixes issue introduced by #1718.

### DIFF
--- a/include/stdexec/__detail/__when_all.hpp
+++ b/include/stdexec/__detail/__when_all.hpp
@@ -285,6 +285,15 @@ namespace STDEXEC {
         __on_stop_.reset();
         // All child operations have completed and arrived at the barrier.
         switch (__state_.load(__std::memory_order_relaxed)) {
+        case __stopped:
+          if constexpr (_SendsStopped) {
+            STDEXEC::set_stopped(static_cast<_Receiver&&>(__rcvr_));
+            break;
+          }
+          // This is reachable because the stop callback sets __stopped whether
+          // or not any child can send set_stopped, therefore we handle this the
+          // same as __started
+          [[fallthrough]];
         case __started:
           if constexpr (!__std::same_as<_ValuesTuple, __ignore>) {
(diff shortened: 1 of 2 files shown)


ericniebler (approved):

ericniebler:
thanks!

PR #1858, opened 2026-02-17
exec(fork_join): dependent sender support

romintomasetti:
This PR fixes `exec::fork_join` so that it supports dependent senders.


romintomasetti:
@ericniebler Could you please have a look ?

ericniebler (approved):
looks good, thanks!

PR #1873, opened 2026-02-20
tests(stdexec): fix and test demangling utility

romintomasetti:


include/stdexec/__detail/__basic_sender.hpp
```diff
@@ -394,11 +394,15 @@ namespace STDEXEC {
...
-    using __basic_sender_t = __basic_sender<_Tag, _Data, __demangle_t<_Child>...>::type;
+    using __basic_sender_t = __basic_sender<_Tag, _Data, __demangle_t<_Child>...>;
 
     template <auto _Descriptor>
     extern __declfn_t<__minvoke<__result_of<_Descriptor>, __q<__basic_sender_t>>>
       __demangle_v<__sexpr<_Descriptor>>;
+
+    template <class _Sexpr, class _Receiver>
+    extern __declfn_t<::STDEXEC::__opstate<__demangle_t<_Sexpr>, _Receiver>>
+      __demangle_v<::STDEXEC::__opstate<_Sexpr, _Receiver>>;
```
ericniebler: demangling is only for sender types

include/stdexec/__detail/__basic_sender.hpp
```diff
@@ -394,11 +394,15 @@ namespace STDEXEC {
   // senders in compiler diagnostics.
   namespace __detail {
     template <class _Tag, class _Data, class... _Child>
-    using __basic_sender_t = __basic_sender<_Tag, _Data, __demangle_t<_Child>...>::type;
+    using __basic_sender_t = __basic_sender<_Tag, _Data, __demangle_t<_Child>...>;
```
ericniebler: i added `::type` recently as a necessary bit of ADL isolation. what problem is it causing?

include/stdexec/__detail/__basic_sender.hpp
romintomasetti: OK, reverted.

include/stdexec/__detail/__basic_sender.hpp
romintomasetti: I just found it weird that the reported demangled type looks like `STDEXEC::__basic_sender<...>::type`. But I've reverted the change.

romintomasetti:
@ericniebler I've addressed your comments, this PR now only adds a test, does not touch anything else.

PR #1874, opened 2026-02-20
ci: clang-format

romintomasetti:
When working on a feature or fix, it's easy to inadvertently touch unrelated lines due to pre-existing formatting inconsistencies. This adds noise to diffs and makes it harder to understand what a commit is actually doing.

This PR introduces a `style.yml` CI job as a dedicated place for style-related checks, starting with `clang-format`. Having this enforced in CI keeps formatting concerns out of feature PRs going forward.

Note that I expect the job to fail for now because there are quite a few pre-existing formatting issues in the codebase. 

@ericniebler Would you be open to landing a formatting cleanup commit first so we can get this green from the start?

@ericniebler I've also deleted `.github/workflows/test-clang-format.sh` because I don't think it's used. But there seems to be other scripts in there that can be deleted.


ericniebler:
i have been wanting to make some changes to stdexec formatting to improve readability. i can do that now, then we can use this CI job to keep things tidy.

ericniebler:
> ```
> /__w/_temp/24e4b664-d13b-4329-b4f5-a2dd768f00ab.sh: 1: git: not found
> ```

looks like the job will need to apt install git first.

romintomasetti:
Indeed, and done :heavy_check_mark:

PR #1878, opened 2026-02-21
exec::sequence: Exception Correctness

RobertLeahy:
Corrected two exception-related problems with exec::sequence:

- Previously exec::sequence senders unconditionally reported that they could throw when connected
- Previously exec::sequence senders unconditionally reported that they could send set_error_t(std::exception_ptr), updated so this is now reported only if it's possible (i.e. if a child sender has that completion, or if any child sender other than the first possibly throws when connected)

### DIFF
--- a/include/exec/sequence.hpp
+++ b/include/exec/sequence.hpp
@@ -236,6 +236,11 @@ namespace experimental::execution
     {
       using sender_concept = STDEXEC::sender_t;
 
+      // Even without an Env, we can sometimes still determine the completion signatures
+      // of the sequence sender. If any of the child senders has a
+      // set_error(exception_ptr) completion, then the sequence sender has a
+      // set_error(exception_ptr) completion. We don't have to ask if any connect call
+      // throws.
       template <class Self, class... Env>
         requires(sizeof...(Env) > 0)
              || __has_eptr_completion<STDEXEC::__copy_cvref_t<Self, Sender0>>
@@ -272,10 +277,10 @@ namespace experimental::execution
       template <STDEXEC::__decay_copyable Self, class Rcvr>
       STDEXEC_ATTRIBUTE(host, device)
       constexpr STDEXEC_EXPLICIT_THIS_BEGIN(auto connect)(this Self&& self, Rcvr rcvr)
-        noexcept(std::is_nothrow_constructible_v<
+        noexcept(STDEXEC::__nothrow_constructible_from<
                  _opstate<Rcvr, STDEXEC::__copy_cvref_t<Self, Sender0>, Senders...>,
                  Rcvr,
-                 decltype(((Self&&) self)._sndrs)>)
+                 decltype((static_cast<Self&&>(self)._sndrs))>)
       {
         return _opstate<Rcvr, STDEXEC::__copy_cvref_t<Self, Sender0>, Senders...>{
           static_cast<Rcvr&&>(rcvr),


include/exec/sequence.hpp
```diff
@@ -156,6 +154,12 @@ namespace experimental::execution
       template <class CvSndrs>
       STDEXEC_ATTRIBUTE(host, device)
       constexpr explicit _opstate(Rcvr&& rcvr, CvSndrs&& sndrs)
+        noexcept(::STDEXEC::__nothrow_connectable<
+                   decltype(::STDEXEC::__get<0>(::STDEXEC::__declval<CvSndrs>())),
```
ericniebler:
```suggestion
                   ::STDEXEC::__tuple_element_t<0, CvSndrs>,
```

include/exec/sequence.hpp
```diff
@@ -170,19 +174,39 @@ namespace experimental::execution
...
-                                                                 std::move(sndr),
-                                                                 _rcvr_t<Remaining == 1>{self});
-        if constexpr (Remaining > 1)
+        constexpr auto __nth   = sizeof...(Senders) - Remaining;
+        auto*          self    = static_cast<_opstate*>(_self);
+        auto&          sndr    = STDEXEC::__get<__nth + 1>(self->_sndrs);
+        constexpr auto nothrow = noexcept(
+          self->_ops.template __emplace_from<__nth + 1>(STDEXEC::connect,
+                                                        std::move(sndr),
+                                                        _rcvr_t < Remaining == 1 > {self}));
```
ericniebler: i think this can just be:
```c++
constexpr bool nothrow =
  STDEXEC::__nothrow_connectable<STDEXEC::__m_at_c<__nth + 1, Senders...>,
                                 _rcvr_t<Remaining == 1>>;
```

include/exec/sequence.hpp
```diff
@@ -202,36 +226,48 @@ namespace experimental::execution
...
+      template <bool First, bool AddExceptionPtr, class Head>
+      struct _fold_left<First, AddExceptionPtr, Head>
       {
         using __t = STDEXEC::__mtry_q<STDEXEC::__concat_completion_signatures_t>::__f<
-          STDEXEC::completion_signatures<STDEXEC::set_error_t(std::exception_ptr)>,
-          STDEXEC::__completion_signatures_of_t<Head, Env...>>;
+          std::conditional_t<
+            _add_exception_ptr<First, AddExceptionPtr, Head>,
+            STDEXEC::completion_signatures<STDEXEC::set_error_t(std::exception_ptr)>,
+            STDEXEC::completion_signatures<>>,
```
ericniebler:
```suggestion
          STDEXEC::__eptr_completion_unless<_add_exception_ptr<First, AddExceptionPtr, Head>>,
```

include/exec/sequence.hpp
```diff
@@ -248,7 +284,11 @@ namespace experimental::execution
       STDEXEC_ATTRIBUTE(host, device)
       static consteval auto get_completion_signatures()
       {
-        if constexpr (STDEXEC::__decay_copyable<Self>)
+        if constexpr (sizeof...(Env) == 0)
+        {
+          return STDEXEC::__dependent_sender<Self>();
```
ericniebler: a better way to do this is to require an `Env` parameter:
```c++
      template <class Self, class Env>
      STDEXEC_ATTRIBUTE(host, device)
      static consteval auto get_completion_signatures()
      {
```
then you don't need the `if constexpr (sizeof...(Env) == 0)` branch.

ericniebler (approved):

PR #1879, opened 2026-02-21
exec::repeat_until et al.: Dependent Child Sender Support

RobertLeahy:
The TODO eliminated by this commit not only meant that exec::repeat_until et al. previously didn't the usual constant evaluation techniques to report errors, it also meant that a hard compile error (rather than a return type indicating that the sender is dependent) resulted if:

- The child sender was dependent, and
- An attempt was made to determine the completion signatures of the parent without an environment

Updated so exec::repeat_until et al. now properly report that they are dependent if their child is dependent.

### DIFF
--- a/include/exec/repeat_until.hpp
+++ b/include/exec/repeat_until.hpp
@@ -28,6 +28,9 @@
 #include <exception>
 #include <type_traits>
 
+STDEXEC_PRAGMA_PUSH()
+STDEXEC_PRAGMA_IGNORE_EDG(not_used_in_template_function_params)
+
 namespace experimental::execution
 {
   struct _EXPECTING_A_SENDER_OF_ONE_VALUE_THAT_IS_CONVERTIBLE_TO_BOOL_;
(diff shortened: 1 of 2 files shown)


include/exec/repeat_until.hpp
```diff
@@ -255,13 +255,28 @@ namespace experimental::execution
...
+
     struct __repeat_until_impl : __sexpr_defaults
     {
       template <class _Sender, class... _Env>
       static consteval auto __get_completion_signatures()
       {
-        // TODO: port this to use constant evaluation
-        return __completions_t<__child_of<_Sender>, _Env...>{};
+        using __child_t = __child_of<_Sender>;
+        if constexpr (::STDEXEC::__minvocable_q<__completions_t, __child_t, _Env...>)
```
ericniebler: imo, better would be to change `__completions_t` to always be well-formed and return descriptive errors. best would be to rewrite this function to use consteval instead of TMP.

PR #1900, opened 2026-02-26
Add missing include for std::launder

mathisloge:
std::launder is defined in <new> and is missing here. 

I haven't had this issue with gcc14/15 on linux but with llvm 21 on macos.

Ref:
https://en.cppreference.com/w/cpp/utility/launder.html


ericniebler:
thanks!

PR #1903, opened 2026-02-27
replace `std::variant` with `__variant` in `as_awaitable`

ericniebler:


ericniebler:
possible msvc codegen bug

PR #1930, opened 2026-03-13
fix: add STDEXEC_BUILD_SYSTEM_CONTEXT option and fix Conan recipe for header-only vs compiled package

### LINKED ISSUE #1540: To Be Header-Only or Not to Be?
I'm exploring the `stdexec` project and have noticed an inconsistency between its description as a header-only library and the handling of the `system_context` binary in the Conan package.

The README states: "This library is header-only, so all the source code can be found in the include/ directory." However, the `system_context` library, defined in [CMakeLists.txt](https://github.com/NVIDIA/stdexec/blob/b9294978f99d2c4bb0567815f5b3f37558ce1f44/CMakeLists.txt#L513-L524) and built from `src/system_context/system_context.cpp`, is a compiled binary, not header-only. This is confirmed by [Issue #1477](https://github.com/NVIDIA/stdexec/issues/1477), where a user resolved a linking error by explicitly linking to `system_context` when using `exec::get_system_scheduler()`.

The Conan recipe (`conanfile.py`) includes `system_context` in the package and specifies it as a required library (`self.cpp_info.libs = ["system_context"]`). However, it also clears the package ID (`self.info.clear()`), treating the package as header-only. This creates a potential issue: since `system_context` is a platform-specific binary, using the same package ID across different platforms and compilers could lead to compatibility problems (e.g., using a Windows-built `system_context` binary on Linux or macOS).

Could the maintainers clarify:
1. Is `stdexec` intended to be fully header-only, or should the documentation reflect that `system_context` is a required binary component?
2. Should the Conan recipe remove `self.info.clear()` to ensure platform-specific package IDs for the `system_context` binary, avoiding potential cross-platform compatibility issues?

Thank you for maintaining this project and for any clarification you can provide!

solarispika:
Fixes #1540

## Problem

The Conan recipe had two related issues:

1. **CMake always built `system_context`** — there was no option to skip it, even though the README says the library is header-only.
2. **`package_id()` called `self.info.clear()` unconditionally** — this treated the package as always header-only, even when `system_context` (a compiled binary) was included. The `cpp_info.libs = ["system_context"]` entry was also generating an incorrect/missing CMake target.

### CMake (`CMakeLists.txt`, `test/CMakeLists.txt`)

- Added `STDEXEC_BUILD_SYSTEM_CONTEXT` option (default `OFF`) to match the README's header-only claim
- Wrapped `system_context` library target creation and installation in `if(STDEXEC_BUILD_SYSTEM_CONTEXT)`
- Wrapped `test.system_context_replaceability` test target in the same guard

### Conan (`conanfile.py`)

- Added `system_context` option (default `False`)
- Added `configure()` to set `package_type` dynamically (`"header-library"` or `"static-library"`)
- Fixed `package_id()` to only clear settings when `system_context=False` (i.e. header-only)
- Fixed `package_info()` to declare `system_context` as a **component** so CMakeDeps correctly generates the `STDEXEC::system_context` CMake target

### Test package (`test_package/CMakeLists.txt`, `test_package/test.cpp`)

- Use `$<TARGET_NAME_IF_EXISTS:STDEXEC::system_context>` to link conditionally
- Use `if(TARGET STDEXEC::system_context)` to set `STDEXEC_TEST_SYSTEM_CONTEXT` compile definition
- `test.cpp` selects between `exec::get_parallel_scheduler()` and `exec::static_thread_pool` at compile time via `#ifdef`

## Testing

```sh
# Header-only (default) — package_id is empty hash (da39a3ee...)
conan create . --build=missing -tf test_package

# With system_context — package_id reflects platform/compiler settings
conan create . -o system_context=True --build=missing -tf test_package
```


ericniebler:
this is fantastic, thanks! could you fix the formatting? then i'll gladly merge this.

PR #1931, opened 2026-03-13
Add STDEXEC_MAIN_PROJECT and STDEXEC_INSTALL to CMake

mathisloge:
It is common practice to be able to disable install targets because they can conflict with in-tree builds (for example when `stdexec` is added via CPM).

Most projects use a top-level project check to enable installs by default when `stdexec` is the main project and otherwise disable them. I've applied the same pattern here and also replaced the remaining `BUILD_TESTING` usages with `STDEXEC_BUILD_TESTS`.

- Disable install/export targets by default; enable only when `stdexec` is the top-level project.
- Add `STDEXEC_MAIN_PROJECT` and `STDEXEC_INSTALL`.
- Replace remaining `BUILD_TESTING` with `STDEXEC_BUILD_TESTS`.
- Conditionally install `asioexec`, `tbbexec`, etc., to avoid conflicts when used as a dependency.

At least resolves my pain with the install targets when pulling stdexec via CPM/FetchContent or add_subdirectory.


CMakeLists.txt
```diff
@@ -82,21 +82,32 @@ rapids_cmake_build_type(Release)
...
+  if (NOT DEFINED PROJECT_NAME)
+    set(STDEXEC_MAIN_PROJECT ON)
+  endif ()
+endif ()
+
 # Don't build tests if configuring stdexec as a submodule of another
 # CMake project, unless they explicitly set STDEXEC_BUILD_TESTS=TRUE,
 # or they enabled CTest's BUILD_TESTING option
-if ((CMAKE_PROJECT_NAME STREQUAL PROJECT_NAME) OR BUILD_TESTING)
+if (STDEXEC_MAIN_PROJECT OR BUILD_TESTING)
```
ericniebler: did you say you were replacing uses of `BUILD_TESTING`?

mathisloge: Oh yeah, that was one question I wanted to ask:
I wanted to add `STDEXEC_MAIN_PROJECT AND BUILD_TESTING` (changed OR -> AND) to have the following behavior: 

if STDEXEC_MAIN_PROJECT then use BUILD_TESTING as a default value of STDEXEC_BUILD_TESTS
if not STDEXEC_MAIN_PROJECT default to FALSE as a default value of STDEXEC_BUILD_TESTS. 

Therefore if using stdexec as a top level project, you could either use BUILD_TESTING or STDEXEC_BUILD_TESTS and for cases where stdexec is used as a subproject one have to use STDEXEC_BUILD_TESTS to enable tests. 

Would that be an accaptable default behavior?

ericniebler: sounds good to me.

mathisloge: Applied the above mentioned changes

ericniebler:
thanks!

PR #1943, opened 2026-03-20
ASIO can call associated_allocator<CT, Alloc>::get with only one parameter

ericniebler:
```c++
  STDEXEC::task<void> test(asio::steady_timer& timer)
  {
    co_await timer.async_wait(asio::as_tuple(exec::asio::use_sender));
  }
```

fails to compile with:

```
[build] /usr/include/boost/asio/impl/as_tuple.hpp:301:65: fatal error: too few arguments to function call, expected 2, have 1
[build]   301 |     return Associator<Handler, DefaultCandidate>::get(h.handler_);
[build]       |            ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~           ^
```

here, `Associator` is `asio::associated_allocator` specialized for `::exec::asio::detail::completion_token::completion_handler`. ASIO clearly expects the `get` static member function to be callable with one argument. `get` doesn't use its second argument, so making it optional fixes the issue.

attn @RobertLeahy

### DIFF
--- a/test/exec/asio/test_use_sender.cpp
+++ b/test/exec/asio/test_use_sender.cpp
@@ -270,4 +270,21 @@ namespace
     CHECK(ctx.stopped());
   }
 
+#if !STDEXEC_NO_STDCPP_COROUTINES()
+  template <typename Timer>
+  ::STDEXEC::task<void> test_awaitable_in_stdexec_task(Timer& timer)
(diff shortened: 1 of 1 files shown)


RobertLeahy:
LGTM.

PR #1959, opened 2026-03-24
make `STDEXEC::__umin` actually work

ericniebler:
well this is embarrassing


Nable80:
I wish someone would add at least ``-Wall -Wextra`` to compilation options.

Issue #1963, opened 2026-03-26
Should stdexec::spawn accept an lvalue scope token?


gamecentric:
Hi, 
it seems that spawn accepts only rvalue scope tokens:
```cpp
counting_scope scope;
spawn(just(), scope.get_token()); // ok

auto token = scope.get_token();
spawn(just(), token); // error: counting_scope::token& does not satisfy the scope_token concept
spawn(just(), std::move(token)); // ok
```
Scope tokens are required to copiable and all token types in the C++ standard are usually expected to be cheaply copiable. In the specific case of counting_scope::token it's just a pointer to the scope. Thus the restriction seems quite unnatural.

ericniebler:
fixed by #1966

PR #1966, opened 2026-03-27
Make spawn accept lvalue scope tokens

gamecentric:
Spawn only accepts rvalues scope tokens, which is unexpected, since tokens are typically value-oriented, cheaply copiable, objects:
```cpp
counting_scope scope;
spawn(just(), scope.get_token()); // ok

auto token = scope.get_token();
spawn(just(), token); // error: counting_scope::token& does not satisfy the scope_token concept
spawn(just(), std::move(token)); // ok
```
See issue https://github.com/NVIDIA/stdexec/issues/1963


ericniebler:
thanks!

PR #1967, opened 2026-03-28
Fix BWOS queue bugs: stolen items lost across block boundaries

maikel:
This fixes an issue where tasks can be lost without being completed

This PR rewrites the block-level indexing in the BWOS work-stealing queue to use round-tagged counters, preventing items from being silently lost when blocks are reused across wraparounds. This prevents ABA problems


ericniebler:
thanks!

PR #1970, opened 2026-03-30
cicd: enable `-Wall -Wextra -Werror` for clang

romintomasetti:
This PR enables for `Clang` compiler ID:
1. `-Wall -Wextra -Werror`
2. `-pedantic-errors`

### DIFF
--- a/include/stdexec/__detail/__read_env.hpp
+++ b/include/stdexec/__detail/__read_env.hpp
@@ -57,7 +57,7 @@ namespace STDEXEC
       }
 
       _Receiver       __rcvr_;
-      __optional<_Ty> __result_;
+      __optional<_Ty> __result_ = __nullopt;
     };
 
     template <class _Receiver, class _Query, class _Ty>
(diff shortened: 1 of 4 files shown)


CMakeLists.txt
```diff
@@ -262,6 +262,7 @@ add_library(stdexec_executable_flags INTERFACE)
 
 # Enable warnings
 target_compile_options(stdexec_executable_flags INTERFACE
+                       $<$<CXX_COMPILER_ID:Clang>:-pedantic-errors;-Wall;-Wextra;-Werror>
```
ericniebler: I could be wrong but I think by adding these flags to  stdexec_executable_flags, you are adding them to every target that links with `STDEXEC::stdexec`. Consumers of stdexec may not want that.

Perhaps we should add them to the test targets instead. Thoughts?

romintomasetti: I agree :+1:

romintomasetti: I've introduced the `add_compile_diagnostics` function that applies a series of compiler options to a target.

I haven't applied it to all targets, so that both compile diagnostic flags and targets they apply to can be updated in the future.

ericniebler:
thanks!

PR #1974, opened 2026-03-30
Shrink the inline storage in connect_awaitable

ispeters:
This diff adapts an idea originally due to @lewissbaker, originally shared at https://godbolt.org/z/zGG9fsPrz,
and combines it with the synthetic coroutine frame idea from @sgerbino and @vinniefalco in Capy.

Changes to Lewis's original include:
 * never invoke `coro.destroy()` since it's a no-op anyway
 * support `unhandled_stopped()` and `get_env()`
 * no need for a real coroutine since we're synthesizing a frame from a single function pointer

Changes to the synthetic coroutine frame:
 * don't store a `void* data` and use `reinterpret_cast` and known offsets instead

While we're in here optimizing for size, also make storage of the optional awaitable objects conditional on
their existence (i.e. don't store the result of `as_awaitable(auto& promise)` or `operator co_await` if we're
falling back to the defaults).

### DIFF
--- a/include/stdexec/__detail/__connect_awaitable.hpp
+++ b/include/stdexec/__detail/__connect_awaitable.hpp
@@ -74,11 +74,14 @@ namespace STDEXEC
     struct __synthetic_coro_frame
     {
       void (*__resume_)(void*) noexcept;
-      // a coroutine frame is usually headed by two function pointers, resume and destroy
-      // in our use, below, we never invoke destroy so there's no need to actually store
-      // it; instead, leave some padding at the end of the struct so we can use its last
-      // byte to store a bool indicating whether or not we ever started the operation
-      std::byte __padding_[sizeof(void*) - 1]{};
+      // we never invoke __destroy_ so a no-op implementation is fine; we've chosen
+      // the address of a no-op function rather than nullptr in case some rogue awaitable
+      // *does* invoke destroy on the synthesized handle that it receives in its
+      // await_suspend function
+      void (*__destroy_)(void*) noexcept = &__noop_destroy;
+
+     private:
+      static void __noop_destroy(void*) noexcept {}
     };
 
     static constexpr std::ptrdiff_t __promise_offset = sizeof(__synthetic_coro_frame);
@@ -532,11 +535,11 @@ namespace STDEXEC
 
       __synthetic_coro_frame __synthetic_frame_{&__promise_t::__resume};
       [[no_unique_address]]
-      bool __started_{false};
-      [[no_unique_address]]
       _Receiver __rcvr_;
       [[neo_unique_addres]]
       __awaitable_state<_Awaitable, __promise_t> __awaiter_;
+      [[no_unique_address]]
+      bool __started_{false};
     };
   }  // namespace __connect_await
 


include/stdexec/__detail/__connect_awaitable.hpp
```diff
@@ -76,32 +74,33 @@ namespace STDEXEC
...
     {
       static constexpr auto await_ready() noexcept -> bool
       {
         return false;
       }
 
+      template <class _Promise>
+      static constexpr void await_suspend(__std::coroutine_handle<_Promise> __h) noexcept
+      {
+        try
```
ericniebler:
```suggestion
        STDEXEC_TRY
```

include/stdexec/__detail/__connect_awaitable.hpp
```diff
@@ -76,32 +74,33 @@ namespace STDEXEC
...
+        {
+          __h.promise().__opstate_.__on_resume();
+        }
+        catch (...)
```
ericniebler:
```suggestion
        STDEXEC_CATCH_ALL
```

include/stdexec/__detail/__connect_awaitable.hpp
```diff
@@ -110,71 +109,47 @@ namespace STDEXEC
...
+        __std::unreachable();
+      }
+
       static constexpr auto
       operator new([[maybe_unused]] std::size_t __bytes, __opstate_t& __opstate) noexcept -> void*
       {
-        STDEXEC_ASSERT(__bytes <= sizeof(__opstate.__storage_));
+        // the first implementation of storing the coroutine frame inline in __opstate using the
+        // technique in this file is due to Lewis Baker <lewissbaker@gmail.com>, and was first
+        // shared at https://godbolt.org/z/zGG9fsPrz
```
ispeters: @lewissbaker, do have any feedback on this? Placement, content, whatever?

include/stdexec/__detail/__connect_awaitable.hpp
```diff
@@ -216,67 +192,134 @@ namespace STDEXEC
...
+  namespace __connect_await
+  {
     template <class _Awaitable, class _Receiver>
     struct __opstate
     {
       constexpr explicit __opstate(_Awaitable&& __awaitable, _Receiver&& __rcvr)
         noexcept(__is_nothrow)
         : __rcvr_(static_cast<_Receiver&&>(__rcvr))
-        , __task_(__co_impl(*this))
+        , __coro(__co_impl(*this))
```
ericniebler: data members should have a trailing underscore
```suggestion
        , __coro_(__co_impl(*this))
```

include/stdexec/__detail/__connect_awaitable.hpp
```diff
@@ -38,9 +38,7 @@ namespace STDEXEC
   // __connect_await
   namespace __connect_await
   {
-    STDEXEC_PRAGMA_OPTIMIZE_BEGIN()
-
-    static constexpr std::size_t __storage_size  = 8 * sizeof(void*);
+    static constexpr std::size_t __storage_size  = 5 * sizeof(void*);
```
ispeters: I don't know why this needs to be `5 * sizeof(void*)`; I'd expect `4 * sizeof(void*)`.

include/stdexec/__detail/__connect_awaitable.hpp
```diff
@@ -184,26 +159,27 @@ namespace STDEXEC
...
+        __std::unreachable();
       }
 
       constexpr auto unhandled_stopped() noexcept -> __std::coroutine_handle<>
       {
-        STDEXEC::set_stopped(static_cast<_Receiver&&>(__opstate_.__rcvr_));
+        __opstate_.__on_stopped();
         // Returning noop_coroutine here causes the __connect_awaitable
         // coroutine to never resume past the point where it co_await's
         // the awaitable.
```
ispeters: I should clean this up…

ericniebler (commented):
looks great!

include/stdexec/__detail/__connect_awaitable.hpp
```diff
@@ -216,67 +192,134 @@ namespace STDEXEC
...
+              else
+              {
+                // returning false means immediately resume
+              }
+            }
+            else
+            {
+              static_assert(__std::convertible_to<__suspend_result_t, __std::coroutine_handle<>>);
+              auto __resume_target = __awaiter_.await_suspend(__coro);
+              __resume_target.resume();
```
ispeters: Technically, this is allowed to throw, and it's never marked `noexcept` so we can't determine statically if it will or not. The code as-is will swallow any such exception and never complete the operation represented by `__opstate`, which is UB. We should do something about that—choices seem to include catching the exception and completing with `set_error`, or making the UB explicit by invoking `unreachable()` here.

include/stdexec/__detail/__connect_awaitable.hpp
ispeters: Staring at some disassembly, it looks like we need `5 * sizeof(void*)` because the coroutine frame has _three_ function pointers in it: `… (.resume)`, `…(.cleanup)`, and `…(.destroy)`. Lewis's original Godbolt didn't have `…(.cleanup)`; Gemini suggests `…(.cleanup)` gets emitted when the compiler applies the HALO optimization—it differs from `…(.destroy)` in that it doesn't deallocate memory. It seems we need to deceive the compiler into not doing HALO to be able to shrink the inline storage further.

lewissbaker: Usually the compiler stores in the destroy-slot of the coroutine frame either .cleanup or .destroy depending on whether the allocation was optimised out with HALO. i.e. if it did elide the allocation then it sets the destroy function pointer to point at the impl that does not call operator delete to free the memory, otherwise if it didn't elide the allocation then it sets the destroy function-pointer to point to the impl that does call operator delete to free the memory.

The existence of both .destroy and .cleanup methods probably just means that in some places this function is being HALO'd and in other places it is not.

include/stdexec/__detail/__connect_awaitable.hpp
ericniebler: i like the idea of using `unreachable()` for this

ericniebler:
so, it looks like CI has been vacuously green for a couple of weeks, and in that time i managed to break coroutine support for msvc. i can't fix the CI issue until i get the coro tests passing on msvc again. i'll try to get to that soon.

sorry for the delay.

include/stdexec/__detail/__connect_awaitable.hpp
ispeters: I _think_ the reason for the compiler's uncertainty was that we were storing the coroutine handle and awaitable machinery in the opstate, which has a clear lifetime bound. Moving the invocation of the coroutine from the opstate constructor to the body of `start()` seems to have convinced the compiler never to HALO.

include/stdexec/__detail/__connect_awaitable.hpp
```diff
@@ -38,9 +38,10 @@ namespace STDEXEC
   // __connect_await
   namespace __connect_await
   {
-    STDEXEC_PRAGMA_OPTIMIZE_BEGIN()
-
-    static constexpr std::size_t __storage_size  = 8 * sizeof(void*);
+    // four pointers' worth of space when compiling with Clang or MSVC; five with GCC
+    static constexpr std::size_t __num_pointers  = 4 * (STDEXEC_CLANG() + STDEXEC_MSVC())
+                                                 + 5 * STDEXEC_GCC();
```
ispeters: @ericniebler, I wonder what you think of this strategy. I don't know what kind of ODR issues it might raise, or cross-compiler incompatibilities. I also haven't figured out why GCC needs another pointer's worth of space than Clang—it's not an extra function pointer because `__promise_offset` is still `2 * sizeof(void*)`  and the assert that checks that constant against the runtime offset still passes (plus I did some `cout` debugging). GCC seems to generate a larger coroutine frame than Clang does but I haven't been able to grok any disassembly to figure out why, yet.

ericniebler: i could be wrong, but i think when compiling with clang in msvc-compatibility mode (CLANG_CL), both `STDEXEC_CLANG()` and `STDEXEC_MSVC()` are 1, but you should double-check me on that.

ispeters: I'm less concerned about the "clever" way I'm using the compiler-detection macros and more concerned about varying the storage size with the preprocessor. If using the preprocessor for this is acceptable but the current clever approach is broken then I can be less clever.

I guess you're not worried about cross-compiler miscompilations, assuming I get the compiler detection logic actually correct?

ispeters:
Oh blarg, I upgraded Clang and I guess I got a new `clang-format`, which has ruined the formatting. Uhh… that's annoying. I'll see what I can do to fix my local tools.

ericniebler:
> Oh blarg, I upgraded Clang and I guess I got a new clang-format, which has ruined the formatting. Uhh… that's annoying. I'll see what I can do to fix my local tools.

the whole formatting thing has been a bit of a pain. CI uses clang-format-21 version 21.1.5. i also have clang-format-21, but i have 21.0.0, which formats _slightly_ differently. it's crazy-making.

ispeters:
Whoops, I missed the template-template bug; I'll have to iterate. How in the heck did any of this work locally?

ericniebler:
this looks interesting:
```
30: test.stdexec: /home/coder/stdexec/include/stdexec/__detail/__connect_await
able.hpp:419: static constexpr void* std::execution::__connect_await::__promis
e<_Awaitable, _Receiver>::operator new(std::size_t, __opstate_t&) [with _Await
able = {anonymous}::ready_awaitable<int>; _Receiver = {anonymous}::expect_valu
e_receiver<std::execution::env<>, int>; std::size_t = long unsigned int; __ops
tate_t = std::execution::__connect_await::__promise<{anonymous}::ready_awaitab
le<int>, {anonymous}::expect_value_receiver<std::execution::env<>, int> >::__
opstate_t]: Assertion `__bytes == __storage_size + 1' failed.
```

ispeters:
Kinda looks like a bunch of the failures are compiler-specific size mismatches. I can't run GCC with ASAN on Apple silicon, nor can I run MSVC, so I plan to add a commit that prints the values of the various asserted-upon variables to stdout so the failure log is more illuminating.

As for the GCC 11 failure, it's an ICE that repros in debug mode but not release mode, so that's a bit wild.

include/stdexec/__detail/__connect_awaitable.hpp
```diff
@@ -26,23 +26,59 @@
...
   {
-    STDEXEC_PRAGMA_OPTIMIZE_BEGIN()
+#  if STDEXEC_CLANG()
+// clang normally needs four pointer's worth of storage for the promise type
+// but the requirement is different when it's a CUDA compilation
+#    if !STDEXEC_CUDA_COMPILATION()
+#      define STDEXEC_CONNECT_AWAITABLE_NUM_POINTERS 4
+#    elif CUDART_VERSION == 12'00'0
+// Clang for CUDA 12.0 only needs three pointers
+#      define STDEXEC_CONNECT_AWAITBALE_NUM_POINTERS 3
```
ispeters:
```suggestion
#      define STDEXEC_CONNECT_AWAITABLE_NUM_POINTERS 3
```

include/stdexec/__detail/__connect_awaitable.hpp
```diff
@@ -80,215 +71,473 @@ namespace STDEXEC
       __with_await_transform() = default;
     };
 
-    struct __awaiter_base
+    struct __synthetic_coro_frame
     {
-      static constexpr auto await_ready() noexcept -> bool
+      void (*__resume_)(void*) noexcept;
+      void (*__destroy_)(void*) noexcept = nullptr;
```
ericniebler: why is one field initialized and the other isn't?

ispeters: This changes in the next commit, but the reason is that we never invoke `destroy` so only the `resume` member needs a value. I figured a null function pointer was a better "don't touch this" value than an indeterminate pointer value in case some rogue awaitable does this
```c++
void await_suspend(coroutine_handle<> coro) {
  coro.destroy();
}
```

include/stdexec/__detail/__connect_awaitable.hpp
```diff
@@ -80,215 +71,477 @@ namespace STDEXEC
...
+      [[no_unique_address]]
+      _Awaitable __source_awaitable_;
+      union
+      {
+        [[no_unique_address]]
+        __state __awaiter_;
+      };
+
+      template <class _A>
+        requires(!std::same_as<std::remove_cvref_t<_A>, __awaitable_state>)
```
ericniebler:
```suggestion
      template <__not_decays_to<__awaitable_state> _A>
```
also, i have learned to avoid identifiers like `_A` (leading underscore followed by a single capital letter). too often system headers make these macros.

include/stdexec/__detail/__connect_awaitable.hpp
ispeters: This whole question is moot.

PR #2020, opened 2026-04-12
workaround: task use-after-free on MSVC due to symmetric transfer codegen bug

lixin-wei:
## The Problem

I met a segment fault on MSVC (_MSC_VER 1944)

The root cause seems to be a bug of MSVC's code generation in symmetric transfer. It generates something like:

```cpp
// MSVC's compiler-generated pseudocode:
handle = awaiter.await_suspend(current_coro);
current_coro.frame->__temp = handle;   // store into suspended frame
handle.resume();                        // regular call, not tail-jump
```

That intermediate write back into the suspended coroutine's frame is the problem. In stdexec's task<T>, await_suspend calls __completed(), which destroys the coroutine frame (via __sink(task)) as part of its cleanup. When await_suspend returns, the frame is already freed — but MSVC still writes the returned handle into it, causing a write-after-free crash.

I've searched around but this bug seems not fixed yet.

## The Workaround

This PR makes a workaround by disabling symmetric transfer in MSVC.
But this will cause the `test_task_awaits_inline_sndr_without_stack_overflow` to fail, which tests symmetric transfer. So I also disabled this test in MSVC.

### DIFF
--- a/include/stdexec/__detail/__task.hpp
+++ b/include/stdexec/__detail/__task.hpp
@@ -784,11 +784,33 @@ namespace STDEXEC
         return false;
       }
 
+#if defined(_MSC_VER)
+      // MSVC does not implement symmetric transfer as a tail call.  When
+      // await_suspend returns coroutine_handle<>, MSVC writes the returned
+      // handle back into the suspended coroutine's frame before transferring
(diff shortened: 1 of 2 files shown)


lixin-wei:
it's fixed in 1950, closing

PR #2025, opened 2026-04-14
No when_all_fail test with NVHPC

dkolsen-pgi:
When building with NVHPC, CMake is not configured for CUDA.  The fact that newly added test when_all_fail was set up with "LANGUAGE CUDA" was causing CMake failures when the compiler was NVHPC.

To fix this, put the definition of test when_all_fail inside an
```
	if(NOT (CMAKE_CXX_COMPILER_ID STREQUAL "NVHPC"))
```
block.

### DIFF
--- a/test/nvexec/CMakeLists.txt
+++ b/test/nvexec/CMakeLists.txt
@@ -63,16 +63,18 @@ target_link_libraries(test.nvexec
 
 catch_discover_tests(test.nvexec PROPERTIES TIMEOUT 30)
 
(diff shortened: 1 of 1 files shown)


ericniebler:
but why? nvc++ is passing this test in CI. i took special care to make it so.

dkolsen-pgi:
After discussing this with Eric, we decided that this isn't necessary.  I have reconfigured the NVHPC tests so that a CUDA language target isn't a problem.

PR #2030, opened 2026-04-17
fix: add inline to constexpr vars and extern C to _mm_pause for C++ module compat

guptapratykshh:
Add `inline` to namespace-scope `constexpr` variable definitions in`__any.hpp`, `__config.hpp`, `__query.hpp`, `__utility.hpp`, `__let.hpp`,and `__meta.hpp` to prevent duplicate symbol errors when multiple C++ module
units include the same stdexec headers. Also add `extern "C"` to the`_mm_pause` declaration in `stop_token.hpp` to match the C linkage expected by MSVC's intrinsic headers in module builds


ericniebler (approved):

ericniebler:
thanks!

PR #2035, opened 2026-04-19
fix: remove digit separators from integer literals for clang-scan-deps C++ module compat

guptapratykshh:
clang-scan-deps rejects apostrophe-separated digit literals (e.g. 1'000) when scanning dependencies for C++ modules. Remove all digit separators from integer literals in exec and stdexec headers so that dependency scanning succeeds without source patching on the consumer side.

### DIFF
--- a/include/stdexec/__detail/__config.hpp
+++ b/include/stdexec/__detail/__config.hpp
@@ -599,7 +599,7 @@ namespace STDEXEC
 #endif
 
 #if __has_include(<memory_resource>) && \
-  (defined(__cpp_lib_memory_resource) && __cpp_lib_memory_resource >= 2016'03L)
+  (defined(__cpp_lib_memory_resource) && __cpp_lib_memory_resource >= 201603L)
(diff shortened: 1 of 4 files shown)


include/stdexec/__detail/__config.hpp
```diff
@@ -599,7 +599,7 @@ namespace STDEXEC
 #endif
 
 #if __has_include(<memory_resource>) && \
-  (defined(__cpp_lib_memory_resource) && __cpp_lib_memory_resource >= 2016'03L)
```
ericniebler, line 602: all of the changes except this one should be unnecessary. clang-scan-deps only looks for #if/#else, #include, and import.

ericniebler (changes requested):

include/stdexec/__detail/__config.hpp
guptapratykshh: @ericniebler, please have a look, I have reverted the other files

ericniebler (approved):

ericniebler:
thanks!

PR #2038, opened 2026-04-20
fix deadlock when awaiting a sender completes inline with `set_stopped`

### LINKED ISSUE #2036: Deadlock in stdexec::task destructor when affine sender completes synchronously with set_stopped
The test below deadlocks with current main. 

The underlying issue seems to be that in `__as_awaitable.hpp` `__async_receiver::set_value()` and `set_error()` both call `__done()`, which CAS-sets `__ready_` to true to signal that the inline (same-thread) completion has occurred. However, `set_stopped()` does not call `__done()`. It calls `promise.unhandled_stopped()` directly, leaving `__ready_` as false.

The `unhandled_stopped()` chain eventually destroys the coroutine frame, which destroys the `__sender_awaitable`. The destructor calls `__ready_.wait(false)`, blocking until `__ready_` becomes non-false. Since nobody ever set it to true, this is a permanent deadlock on the same thread that called `start()`.

I'm not sure what the proper fix is.

```cpp
  struct inline_affine_stopped_sender
  {
    using sender_concept        = ex::sender_tag;
    using completion_signatures = ex::completion_signatures<ex::set_stopped_t()>;

    template <class Receiver>
    struct operation
    {
      Receiver rcvr_;

      void start() & noexcept
      {
        ex::set_stopped(std::move(rcvr_));
      }
    };

    template <class Receiver>
    auto connect(Receiver rcvr) && -> operation<Receiver>
    {
      return {std::move(rcvr)};
    }

    struct attrs
    {
      [[nodiscard]]
      static constexpr auto query(ex::__get_completion_behavior_t<ex::set_stopped_t>) noexcept
      {
        return ex::__completion_behavior::__inline_completion
             | ex::__completion_behavior::__asynchronous_affine;
      }
    };

    [[nodiscard]]
    auto get_env() const noexcept -> attrs
    {
      return {};
    }
  };

  TEST_CASE("task co_awaiting inline|async_affine stopped sender does not deadlock",
            "[types][task]")
  {
    auto res = ex::sync_wait([]() -> ex::task<int> {
        co_await inline_affine_stopped_sender{};
        FAIL("Expected co_awaiting inline_affine_stopped_sender to stop the task");
        co_return 42;
    }());
    CHECK(!res.has_value());
  }
```

### LINKED ISSUE #2041: Crash with stdexec::task (probably premature coro destruction)
We are seeing this crash relatively reliably on MSVC 17.14.13 (cl 19.44.35215).

I suspect that the task coro handle is destroyed before final_suspend or something along those lines.

I had hoped that the changes in https://github.com/NVIDIA/stdexec/pull/2038 relating to coro destruction would have fixed this, but unfortunately, it still happens even with those changes.

```cpp
TEST_CASE("stdexec.task.task_crash_repro") {
    auto task = [] -> stdexec::task<void> { co_return; };
    auto pool = exec::static_thread_pool(1);
    auto scope = stdexec::counting_scope();
    for (int i = 0; i < 1000; ++i) {
        stdexec::spawn(stdexec::starts_on(pool.get_scheduler(), task())
                           | stdexec::upon_error([](auto) noexcept {}),
                       scope.get_token());
    }
    stdexec::sync_wait(scope.join());
}
```

ericniebler:
fixes #2036

### DIFF
--- a/include/exec/at_coroutine_exit.hpp
+++ b/include/exec/at_coroutine_exit.hpp
@@ -18,11 +18,12 @@
 
 // The original idea is taken from libunifex and adapted to stdexec.
 
-#include <exception>
-
 #include "../stdexec/execution.hpp"
 
 #include "any_sender_of.hpp"
+
+#include <exception>
+#include <tuple>
 
 namespace experimental::execution
 {
--- a/include/exec/at_coroutine_exit.hpp
+++ b/include/exec/at_coroutine_exit.hpp
@@ -79,8 +79,8 @@ namespace experimental::execution
 
         template <receiver _Receiver>
           requires sender_to<_Sender, __receiver<_Receiver>>
-        auto
-        connect(_Receiver __rcvr) && noexcept -> connect_result_t<_Sender, __receiver<_Receiver>>
+        auto connect(_Receiver __rcvr) && noexcept  //
+          -> connect_result_t<_Sender, __receiver<_Receiver>>
         {
           return STDEXEC::connect(static_cast<_Sender&&>(__sender_),
                                   __receiver<_Receiver>{static_cast<_Receiver&&>(__rcvr)});
@@ -137,6 +137,12 @@ namespace experimental::execution
         : __coro_(std::exchange(__that.__coro_, {}))
       {}
 
+      ~__task()
+      {
+        if (__coro_)
+          __coro_.destroy();
+      }
+
       [[nodiscard]]
       static constexpr auto await_ready() noexcept -> bool
       {
@@ -148,7 +154,7 @@ namespace experimental::execution
       //! coroutine exit; i.e., the coroutine that is co_await-ing the result of calling
       //! at_coroutine_exit.
       template <__has_continuation _Promise>
-      auto await_suspend(__std::coroutine_handle<_Promise> __parent) noexcept -> bool
+      auto await_suspend(__std::coroutine_handle<_Promise> __parent) -> bool
       {
         // Set the cleanup task's scheduler to the parent coroutine's scheduler.
         __coro_.promise().__scheduler_ = get_start_scheduler(get_env(__parent.promise()));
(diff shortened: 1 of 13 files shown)


mika-fischer:
@ericniebler FYI, the current changes in this PR (i.e. b8cd4b6bef6d53f60578a22ed5261919cfbb1647) do fix both #2036 and #2041 here!

ericniebler:
Hooray! Thanks for letting me know.

mika-fischer:
I want to add that the changes in this PR broke `sender_to` for `stdexec::task` when the environment has no `start_scheduler`, at least in MSVC. This works in main, but now gives a hard compile error:

```cpp
struct void_rcvr {
    using receiver_concept = stdexec::receiver_tag;
    auto get_env() const noexcept { return stdexec::env<>{}; }
    void set_value() && noexcept {}
};
static_assert(!stdexec::sender_to<stdexec::task<void>, void_rcvr>);
```

~~Incidentally, `sender_in` was already broken in the main branch before these changes. This gives a hard compile error both in main and this PR:~~

EDIT: Ignore the last part, I misunderstood `sender_in` semantics.

ericniebler:
> I want to add that the changes in this PR broke `sender_to` for `stdexec::task` when the environment has no `start_scheduler`

i think the fact that it compiled before this pr is a bug. you'll get an error if you try to actually connect that receiver with the `task`. just like the `on` algorithm, `task` _needs_ a start scheduler.

UPDATE: see for example https://godbolt.org/z/E5j9beGq1

mika-fischer:
>> `static_assert(!stdexec::sender_to<stdexec::task<void>, void_rcvr>);`
>
> i think the fact that it compiled before this pr is a bug. you'll get an error if you try to actually connect that receiver with the `task`. just like the `on` algorithm, `task` _needs_ a start scheduler.

I think you might have missed the `!` in the `static_assert` above.

I don't say that `sender_to<task<void>, void_rcvr>` should be `true`, I'm saying it should be `false`, which it was in main, but now in this PR it just doesn't compile at all anymore, which seems wrong.

EDIT: See https://godbolt.org/z/M5M459jzo

ericniebler:
Ah thanks for clarifying. I'll look into it.

PR #2040, opened 2026-04-21
Introduce exec::function<...>

ispeters:
This PR proposes a new type-erased sender named `exec::function`. There's an in-code comment giving a bunch of examples, but a simple example is:
```c++
exec::function<int(int)> fn(42, [](int i) { return ex::just(i); });

auto [result] = ex::sync_wait(std::move(fn)).value();

assert(result == 42);
```

There are a bunch of TODOs left, including lots of tests that are missing, but the API is ready to collect early feedback. If this looks like a promising direction, I intend to write a paper for Brno proposing this type for inclusion in C++29.

### DIFF
--- a/include/exec/function.hpp
+++ b/include/exec/function.hpp
@@ -18,6 +18,7 @@
 #include "../stdexec/__detail/__completion_signatures.hpp"
 #include "../stdexec/__detail/__concepts.hpp"
 #include "../stdexec/__detail/__env.hpp"
+#include "../stdexec/__detail/__read_env.hpp"
 #include "../stdexec/__detail/__receivers.hpp"
 #include "../stdexec/__detail/__scope.hpp"
 #include "../stdexec/__detail/__sender_concepts.hpp"
@@ -48,6 +49,31 @@
 // the frame allocator from the environment without relying on TLS.
 namespace experimental::execution
 {
+  struct get_frame_allocator_t : STDEXEC::__query<get_frame_allocator_t>
+  {
+    using STDEXEC::__query<get_frame_allocator_t>::operator();
+
+    constexpr auto operator()() const noexcept
+    {
+      return STDEXEC::read_env(get_frame_allocator_t{});
+    }
+
+    template <class Env>
+    static constexpr void __validate() noexcept
+    {
+      static_assert(STDEXEC::__nothrow_callable<get_frame_allocator_t, Env const &>);
+      using __alloc_t = STDEXEC::__call_result_t<get_frame_allocator_t, Env const &>;
+      static_assert(STDEXEC::__simple_allocator<STDEXEC::__decay_t<__alloc_t>>);
+    }
+
+    static consteval auto query(STDEXEC::forwarding_query_t) noexcept -> bool
+    {
+      return true;
+    }
+  };
+
+  inline constexpr get_frame_allocator_t get_frame_allocator{};
+
   namespace _func
   {
     using namespace STDEXEC;
@@ -197,7 +223,6 @@ namespace experimental::execution
                                     _func_op<Receiver, completion_signatures<Sigs...>>,
                                     Sigs...>
     {
-      // TODO: use get_frame_allocator(get_env(rcvr_)) to allocate and destroy this
       std::unique_ptr<_base_op> op_;
       [[no_unique_address]]
       Receiver rcvr_;
@@ -222,6 +247,23 @@ namespace experimental::execution
         op_->start();
       }
     };
+
+    template <class Env>
+    constexpr auto choose_frame_allocator(Env const & env) noexcept
+    {
+      if constexpr (requires { get_frame_allocator(env); })
+      {
+        return get_frame_allocator(env);
+      }
+      else if constexpr (requires { get_allocator(env); })
+      {
+        return get_allocator(env);
+      }
+      else
+      {
+        return std::allocator<std::byte>();
+      }
+    }
 
     template <class Args, class Sigs, class Env>
     class _func_impl;
@@ -253,12 +295,12 @@ namespace experimental::execution
 
         factory_ = [](receiver_t rcvr, Args&&... args) -> std::unique_ptr<_base_op>
         {
-          using traits = std::allocator_traits<std::allocator<op_t>>;
+          using traits = std::allocator_traits<decltype(choose_frame_allocator(
+            get_env(rcvr)))>::template rebind_traits<op_t>;
 
           Factory factory;
 
-          // TODO: query rcvr for a frame allocator and use it
-          typename traits::allocator_type alloc;
+          typename traits::allocator_type alloc(choose_frame_allocator(get_env(rcvr)));
 
           auto* op = traits::allocate(alloc, 1);
 
--- a/include/exec/function.hpp
+++ b/include/exec/function.hpp
@@ -19,10 +19,12 @@
 #include "../stdexec/__detail/__concepts.hpp"
 #include "../stdexec/__detail/__env.hpp"
 #include "../stdexec/__detail/__receivers.hpp"
+#include "../stdexec/__detail/__scope.hpp"
 #include "../stdexec/__detail/__sender_concepts.hpp"
 
 #include <exception>
 #include <memory>
+#include <new>
 #include <tuple>
 #include <type_traits>
 #include <utility>
@@ -135,12 +137,13 @@ namespace experimental::execution
       virtual void start() & noexcept = 0;
     };
 
-    template <class Sender, class Receiver>
+    template <class Sender, class Receiver, class Allocator>
     struct _derived_op : _base_op
     {
-      explicit _derived_op(Sender&& sndr, Receiver rcvr)
+      explicit _derived_op(Sender&& sndr, Receiver rcvr, Allocator const & alloc)
         noexcept(std::is_nothrow_invocable_v<connect_t, Sender, Receiver>)
         : op_(connect(std::forward<Sender>(sndr), std::move(rcvr)))
+        , alloc_(alloc)
       {}
 
       _derived_op(_derived_op&&) = delete;
@@ -152,8 +155,19 @@ namespace experimental::execution
         ::STDEXEC::start(op_);
       }
 
+      static constexpr void operator delete(_derived_op* p, std::destroying_delete_t)
+      {
+        using traits = std::allocator_traits<Allocator>::template rebind_traits<_derived_op>;
+
+        typename traits::allocator_type alloc = std::move(p->alloc_);
+        traits::destroy(alloc, p);
+        traits::deallocate(alloc, p, 1);
+      }
+
      private:
       connect_result_t<Sender, Receiver> op_;
+      [[no_unique_address]]
+      Allocator alloc_;
     };
 
     template <class Base, class Derived, class... Sigs>
@@ -215,7 +229,7 @@ namespace experimental::execution
     template <class SndrCncpt, class... Args, class... Sigs, class Env>
     class _func_impl<SndrCncpt(Args...), completion_signatures<Sigs...>, Env>
     {
-      _base_op* (*factory_)(_func_rcvr<completion_signatures<Sigs...>>, Args&&...);
+      std::unique_ptr<_base_op> (*factory_)(_func_rcvr<completion_signatures<Sigs...>>, Args&&...);
       [[no_unique_address]]
       std::tuple<Args...> args_;
 
@@ -235,13 +249,30 @@ namespace experimental::execution
         using sender_t   = std::invoke_result_t<Factory, Args...>;
         using receiver_t = _func_rcvr<completion_signatures<Sigs...>>;
 
-        using op_t = _derived_op<sender_t, receiver_t>;
+        using op_t = _derived_op<sender_t, receiver_t, std::allocator<std::byte>>;
 
-        factory_ = [](receiver_t rcvr, Args&&... args) -> _base_op*
+        factory_ = [](receiver_t rcvr, Args&&... args) -> std::unique_ptr<_base_op>
         {
+          using traits = std::allocator_traits<std::allocator<op_t>>;
+
           Factory factory;
+
           // TODO: query rcvr for a frame allocator and use it
-          return new op_t(factory(std::forward<Args>(args)...), std::move(rcvr));
+          typename traits::allocator_type alloc;
+
+          auto* op = traits::allocate(alloc, 1);
+
+          __scope_guard guard{[&]() noexcept { traits::deallocate(alloc, op, 1); }};
+
+          traits::construct(alloc,
+                            op,
+                            factory(std::forward<Args>(args)...),
+                            std::move(rcvr),
+                            alloc);
+
+          guard.__dismiss();
+
+          return std::unique_ptr<_base_op>(op);
         };
       }
 
--- a/include/exec/function.hpp
+++ b/include/exec/function.hpp
@@ -53,40 +53,14 @@ namespace experimental::execution
     template <class Sig>
     struct _virt_completion;
 
-    template <class Error>
-    struct _virt_completion<set_error_t(Error)>
+    template <class CPO, class... Args>
+    struct _virt_completion<CPO(Args...)>
     {
       _virt_completion() = default;
 
       _virt_completion(_virt_completion&&) = delete;
 
-      virtual void set_error(Error&& err) noexcept = 0;
-
-     protected:
-      ~_virt_completion() = default;
-    };
-
-    template <>
-    struct _virt_completion<set_stopped_t()>
-    {
-      _virt_completion() = default;
-
-      _virt_completion(_virt_completion&&) = delete;
-
-      virtual void set_stopped() noexcept = 0;
-
-     protected:
-      ~_virt_completion() = default;
-    };
-
-    template <class... Values>
-    struct _virt_completion<set_value_t(Values...)>
-    {
-      _virt_completion() = default;
-
-      _virt_completion(_virt_completion&&) = delete;
-
-      virtual void set_value(Values&&... values) noexcept = 0;
+      virtual void complete(CPO, Args&&...) noexcept = 0;
 
      protected:
       ~_virt_completion() = default;
@@ -96,44 +70,16 @@ namespace experimental::execution
     struct _virt_completions;
 
     template <class... Sigs>
-    struct _virt_completions<completion_signatures<Sigs...>> : virtual _virt_completion<Sigs>...
+    struct _virt_completions<completion_signatures<Sigs...>> : _virt_completion<Sigs>...
     {
       _virt_completions() = default;
 
       _virt_completions(_virt_completions&&) = delete;
 
+      using _virt_completion<Sigs>::complete...;
+
      protected:
       ~_virt_completions() = default;
-    };
-
-    template <class Sig, class Derived>
-    struct _func_rcvr_base;
-
-    template <class Error, class Derived>
-    struct _func_rcvr_base<set_error_t(Error), Derived>
-    {
-      void set_error(Error&& err) && noexcept
-      {
-        static_cast<Derived*>(this)->completer_->set_error(std::forward<Error>(err));
-      }
-    };
-
-    template <class Derived>
-    struct _func_rcvr_base<set_stopped_t(), Derived>
-    {
-      void set_stopped() && noexcept
-      {
-        static_cast<Derived*>(this)->completer_->set_stopped();
-      }
-    };
-
-    template <class... Value, class Derived>
-    struct _func_rcvr_base<set_value_t(Value...), Derived>
-    {
-      void set_value(Value&&... value) && noexcept
-      {
-        static_cast<Derived*>(this)->completer_->set_value(std::forward<Value>(value)...);
-      }
     };
 
     template <class Sigs>
@@ -141,10 +87,7 @@ namespace experimental::execution
 
     template <class... Sigs>
     class _func_rcvr<completion_signatures<Sigs...>>
-      : public _func_rcvr_base<Sigs, _func_rcvr<completion_signatures<Sigs...>>>...
     {
-      friend _func_rcvr_base<Sigs, _func_rcvr>...;
-
       using completer_t = _virt_completions<completion_signatures<Sigs...>>;
 
       completer_t* completer_;
@@ -155,6 +98,28 @@ namespace experimental::execution
       explicit _func_rcvr(completer_t& completer) noexcept
         : completer_(std::addressof(completer))
       {}
+
+      template <class Error>
+      void set_error(Error&& err) && noexcept
+        requires requires { this->completer_->complete(set_error_t{}, std::forward<Error>(err)); }
+      {
+        this->completer_->complete(set_error_t{}, std::forward<Error>(err));
+      }
+
+      void set_stopped() && noexcept
+        requires requires { this->completer_->complete(set_stopped_t{}); }
+      {
+        this->completer_->complete(set_stopped_t{});
+      }
+
+      template <class... Values>
+      void set_value(Values&&... values) && noexcept
+        requires requires {
+          this->completer_->complete(set_value_t{}, std::forward<Values>(values)...);
+        }
+      {
+        this->completer_->complete(set_value_t{}, std::forward<Values>(values)...);
+      }
 
       // TODO: get_env
     };
@@ -180,9 +145,9 @@ namespace experimental::execution
 
       _derived_op(_derived_op&&) = delete;
 
-      ~_derived_op() override = default;
+      ~_derived_op() final = default;
 
-      void start() & noexcept override
+      void start() & noexcept final
       {
         ::STDEXEC::start(op_);
       }
@@ -191,35 +156,20 @@ namespace experimental::execution
       connect_result_t<Sender, Receiver> op_;
     };
 
-    template <class Sig, class Derived>
+    template <class Base, class Derived, class... Sigs>
     struct _func_op_completion;
 
-    template <class Error, class Derived>
-    struct _func_op_completion<set_error_t(Error), Derived>
-      : virtual _virt_completion<set_error_t(Error)>
-    {
-      void set_error(Error&& err) noexcept final
-      {
-        static_cast<Derived*>(this)->complete(set_error_t{}, std::forward<Error>(err));
-      }
-    };
+    template <class Base, class Derived>
+    struct _func_op_completion<Base, Derived> : Base
+    {};
 
-    template <class Derived>
-    struct _func_op_completion<set_stopped_t(), Derived> : virtual _virt_completion<set_stopped_t()>
+    template <class Base, class Derived, class CPO, class... Args, class... Sigs>
+    struct _func_op_completion<Base, Derived, CPO(Args...), Sigs...>
+      : _func_op_completion<Base, Derived, Sigs...>
     {
-      void set_stopped() noexcept final
+      void complete(CPO, Args&&... args) noexcept final
       {
-        static_cast<Derived*>(this)->complete(set_stopped_t{});
-      }
-    };
-
-    template <class... Value, class Derived>
-    struct _func_op_completion<set_value_t(Value...), Derived>
-      : virtual _virt_completion<set_value_t(Value...)>
-    {
-      void set_value(Value&&... value) noexcept final
-      {
-        static_cast<Derived*>(this)->complete(set_value_t{}, std::forward<Value>(value)...);
+        static_cast<Derived*>(this)->complete(CPO{}, std::forward<Args>(args)...);
       }
     };
 
@@ -228,15 +178,17 @@ namespace experimental::execution
 
     template <class Receiver, class... Sigs>
     class _func_op<Receiver, completion_signatures<Sigs...>>
-      : private _virt_completions<completion_signatures<Sigs...>>
-      , private _func_op_completion<Sigs, _func_op<Receiver, completion_signatures<Sigs...>>>...
+      : private _func_op_completion<_virt_completions<completion_signatures<Sigs...>>,
+                                    _func_op<Receiver, completion_signatures<Sigs...>>,
+                                    Sigs...>
     {
       // TODO: use get_frame_allocator(get_env(rcvr_)) to allocate and destroy this
       std::unique_ptr<_base_op> op_;
       [[no_unique_address]]
       Receiver rcvr_;
 
-      friend _func_op_completion<Sigs, _func_op>...;
+      template <class B, class D, class... S>
+      friend struct _func_op_completion;
 
       template <class CPO, class... Arg>
       void complete(CPO cpo, Arg&&... arg) noexcept
@@ -279,8 +231,8 @@ namespace experimental::execution
         requires STDEXEC::__not_decays_to<Factory, _func_impl>  //
               && std::constructible_from<Factory>               //
               && STDEXEC::__callable<Factory, Args...>
-      //&& STDEXEC::sender_to<STDEXEC::__invoke_result_t<Factory, Args...>,
-      //_func_rcvr<completion_signatures<Sigs...>>>
+              && STDEXEC::sender_to<STDEXEC::__invoke_result_t<Factory, Args...>,
+                                    _func_rcvr<completion_signatures<Sigs...>>>
       explicit(sizeof...(Args) == 0) _func_impl(Args&&... args, Factory&& factory)
         noexcept((std::is_nothrow_constructible_v<Args, Args> && ...))
         : args_(std::forward<Args>(args)...)
--- a/include/exec/function.hpp
+++ b/include/exec/function.hpp
@@ -95,11 +95,11 @@ namespace experimental::execution
       constexpr ~_virt_completion() = default;
     };
 
-    template <class Sigs>
+    template <class Sigs, class Env>
     struct _virt_completions;
 
-    template <class... Sigs>
-    struct _virt_completions<completion_signatures<Sigs...>> : _virt_completion<Sigs>...
+    template <class... Sigs, class Env>
+    struct _virt_completions<completion_signatures<Sigs...>, Env> : _virt_completion<Sigs>...
     {
       constexpr _virt_completions() = default;
 
@@ -107,17 +107,19 @@ namespace experimental::execution
 
       using _virt_completion<Sigs>::complete...;
 
+      virtual Env get_env() const noexcept = 0;
+
      protected:
       constexpr ~_virt_completions() = default;
     };
 
-    template <class Sigs>
+    template <class Sigs, class Env>
     class _func_rcvr;
 
-    template <class... Sigs>
-    class _func_rcvr<completion_signatures<Sigs...>>
+    template <class... Sigs, class Env>
+    class _func_rcvr<completion_signatures<Sigs...>, Env>
     {
-      using completer_t = _virt_completions<completion_signatures<Sigs...>>;
+      using completer_t = _virt_completions<completion_signatures<Sigs...>, Env>;
 
       completer_t* completer_;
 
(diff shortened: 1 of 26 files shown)


include/exec/function.hpp
```diff
@@ -0,0 +1,693 @@
...
+
+  inline constexpr get_frame_allocator_t get_frame_allocator{};
+
+  namespace _qry_detail
+  {
+    template <class Sig>
+    inline constexpr bool is_query_function_v = false;
+
+    template <class Return, class Query, class... Args, bool NoThrow>
+    inline constexpr bool is_query_function_v<Return(Query, Args...) noexcept(NoThrow)> = true;
```
ericniebler: i don't think `NoThrow` can be deduced this way in standard C++, despite the fact that some compilers accept it. you need two specializations, one with `noexcept` and the other without.

include/exec/function.hpp
```diff
@@ -0,0 +1,693 @@
...
+  // function types. For example:
+  //
+  //   queries<
+  //     std::execution::inline_stop_token(std::execution::get_stop_token_t) noexcept,
+  //     std::pmr::polymorphic_allocator<std::byte>(std::execution::get_allocator_t)
+  //   >
+  template <class... Queries>
+    requires(_qry_detail::is_query_function_v<Queries> && ...)
+  struct queries
+  {};
```
ericniebler: there already is an `exec::queries` type list thing in `exec/any_sender_of.hpp`. we should share it.

include/exec/function.hpp
```diff
@@ -0,0 +1,693 @@
...
+    //
+    // the base template is an empty class, representing the empty set of queries.
+    template <class... Queries>
+    struct _env_of_queries
+    {};
+
+    // a special case in the recursion: when there is only one query in the pack, there's
+    // no base implementation of query to put in the using statement
+    template <class Return, class Query, class... Args, bool NoThrow>
+    struct _env_of_queries<Return(Query, Args...) noexcept(NoThrow)>
```
ericniebler: same thing here and elsewhere about deducing `NoThrow`

include/exec/function.hpp
```diff
@@ -0,0 +1,693 @@
...
+
+      _env_of_queries(_env_of_queries &&) = delete;
+
+      using _env_of_queries<Queries...>::query;
+
+      virtual Return query(Query query, Args &&...args) const noexcept(NoThrow) = 0;
+
+     protected:
+      ~_env_of_queries() = default;
+    };
```
ericniebler: i feel like we can reuse the `_iquery_memfn` interface from `exec/any_sender_of.hpp` and `__any` from `stdexec/__detail/__any.hpp`.

include/exec/function.hpp
```diff
@@ -0,0 +1,693 @@
...
+        requires requires { completer_->complete(set_value_t{}, std::forward<Values>(values)...); }
+      {
+        completer_->complete(set_value_t{}, std::forward<Values>(values)...);
+      }
+
+      constexpr auto get_env() const noexcept -> _delegate_env<Queries...>
+      {
+        return STDEXEC::get_env(*completer_);
+      }
+    };
```
ericniebler: this duplicates a ton of `any_sender_of.hpp`.

include/exec/function.hpp
```diff
@@ -0,0 +1,693 @@
...
+        using traits = std::allocator_traits<Allocator>::template rebind_traits<_derived_op>;
+
+        typename traits::allocator_type alloc = std::move(p->alloc_);
+        traits::destroy(alloc, p);
+        traits::deallocate(alloc, p, 1);
+      }
+
+     private:
+      connect_result_t<Sender, Receiver> op_;
+      [[no_unique_address]]
```
ericniebler:
```suggestion
      STDEXEC_ATTRIBUTE(no_unique_address)
```

there are too many compilers that generate bad code with this attribute

include/exec/function.hpp
```diff
@@ -0,0 +1,693 @@
...
+      : _func_op_completion<Base, Derived, Sigs...>
+    {
+      void complete(CPO, Args &&...args) noexcept final
+      {
+        // This seems like it ought to be true, but it fails...
+        //
+        // Some testing shows it's being evaluated when Derived is incomplete
+        // during constraint satisfaction testing.
+        //
+        // static_assert(std::derived_from<_func_op_completion, Derived>);
```
ericniebler: you could turn this into a runtime assert instead, i think. could be wrong about that.

include/exec/function.hpp
```diff
@@ -0,0 +1,693 @@
...
+      std::tuple<Args...> args_;
+
+     public:
+      using sender_concept = SndrCncpt;
+
+      // TODO: I only know this works for empty lambdas; figure out whether function pointers
+      //       and/or pointer-to-member functions can be made to work
+      template <STDEXEC::__callable<Args...> Factory>
+        requires STDEXEC::__not_decays_to<Factory, _func_impl>  //
+              && std::constructible_from<Factory>               //
```
ericniebler:
```suggestion
              && STDEXEC::__std::constructible_from<Factory>               //
```

include/exec/function.hpp
```diff
@@ -0,0 +1,693 @@
...
+              && STDEXEC::__callable<Factory, Args...>
+              && STDEXEC::sender_to<STDEXEC::__invoke_result_t<Factory, Args...>,
+                                    _func_rcvr<completion_signatures<Sigs...>, Queries...>>
+      constexpr explicit(sizeof...(Args) == 0) _func_impl(Args &&...args, Factory &&factory)
```
ericniebler:
```suggestion
      constexpr explicit _func_impl(Args &&...args, Factory &&factory)
```

include/exec/function.hpp
```diff
@@ -0,0 +1,693 @@
...
+        noexcept((std::is_nothrow_constructible_v<Args, Args> && ...))
```
ericniebler:
```suggestion
        noexcept(STDEXEC::__nothrow_move_constructible<Args...>)
```

ericniebler (commented):
why capture a function and arguments instead of just a sender that aggregates the arguments? what does lazy construction of the sender offer here?

could this be implemented in terms of:

```c++
template <class Result,
          class ReceiverQueries = queries<>,
          class Completions = completion_signatures<set_error_t(exception_ptr),
                                                    set_stopped_t()>,
          class SenderQueries = queries<>>
auto function(auto sndr) 
{
  using _completions_t =
    __minvoke<__mpush_back<__q<completion_signatures>>, Completions, set_value_t(Result)>;

  using _sender_t =
    any_sender<any_receiver<_completions_t, ReceiverQueries>, SenderQueries>;

   return _sender_t(let_value(read_env(get_frame_allocator),
                              [=](auto const& alloc)
                              {
                                return __uses_frame_allocator(sndr, alloc);
                              }));
}
```

EDIT: also, the `exec::function` interface suggests to me that it would be used like:

```c++
exec::function<int(int)> fn([](int i) { return ex::just(i); });

auto [result] = ex::sync_wait(fn(42)).value();
```

the lazy construction of the sender would then makes sense.

include/exec/function.hpp
ispeters: [This one](https://github.com/NVIDIA/stdexec/blob/main/include/exec/any_sender_of.hpp#L31-L32)? I can't see the definition of the primary template anywhere; is a declaration enough?

ispeters: I completely agree that the two type-erasing tools should share a `queries` type list thing, but we've arrived at quite different type lists, and I'm not sure I understand the trade offs between the two approaches. Can you weigh in and illuminate your design choices?

To require a `noexcept` query `get_stop_token` that returns an `inplace_stop_token`, your `queries` requires a non-type template parameter of this type:
```c++
get_stop_token_t(*)(inplace_stop_token(*)()) noexcept
```
I communicate the same information with a template type parameter of this type:
```c++
inplace_stop_token(get_stop_token_t) noexcept
```

For an arbitrary query of type `Q`, taking arguments `Args...`, and returning `Return`, the two approaches look like this:
```c++
// cast nullptr to the appropriate function pointer type
(Q(*)(Return(*)(Args...)) noexcept)nullptr

// specify a function type
Return(Q, Args...) noexcept
```

Perhaps unsurprisingly, my idea fits more easily in my head; the way you specify a query is to provide the signature of the `query` member function that must be implemented on the environment type:
```c++
// given a required query like Return(Q, Args...), the env must have:
Return query(Q, Args...)
```

Your approach separates the query's "logical signature" (`Return(Args...)`) from the query's tag type in a way that I see as mapping a query tag type to its function signature, which I can appreciate, and it traffics in compile-time values instead of types.

To me, it seems like it's more work to create the values your approach requires; my approach, in contrast, just requires a type declaration so there's no value declaration required. What value do we get out of having a value at compile time instead of the type? The only part of the value we actually use is its type, right?

ericniebler: > ```c++
> template <class... Queries>
> struct queries;
> ```

yes, that's the one. it doesn't need a definition. it is purely a compile-time thing.

> ```c++
> get_stop_token_t(*)(inplace_stop_token(*)()) noexcept
> ```

that's the old legacy interface. i agree it's bad.

> ```c++
> inplace_stop_token(get_stop_token_t) noexcept
> ```

that's the syntax i currently use. maybe you got confused by some old tests, which i kept while the old syntax is still supported.

include/exec/function.hpp
ispeters: This is redundant now that I've deleted most of the vtable shenanigans and replaced them with the pre-existing `_any_receiver_ref`.

include/exec/function.hpp
ispeters: I think I've cleaned up the duplication.

include/exec/function.hpp
ispeters: It was enough to just use `_any_receiver_ref`.

include/exec/function.hpp
```diff
@@ -0,0 +1,402 @@
...
+        //else
+        //{
+        // TODO: make this error accurate
+        //return __throw_compile_time_error(__unrecognized_sender_error_t<Sender, RcvrEnv>());
+        //}
+      }
+
+      // TODO: this assumes rvalue connection; lvalue connection requires thought and tests
+      template <class Receiver>
+      constexpr _func_op_t<Receiver> connect(Receiver rcvr)
```
ericniebler:
```suggestion
      constexpr _func_op_t<Receiver> connect(Receiver rcvr) &&
```

include/exec/function.hpp
```diff
@@ -0,0 +1,402 @@
...
+      {
+        return _func_op_t<Receiver>{
+          std::move(rcvr),
+          [this](auto rcvr)
+          {
+            return std::apply([&rcvr, this](Args &&...args)
+                              { return factory_(std::move(rcvr), std::forward<Args>(args)...); },
+                              std::move(args_));
```
ericniebler: if you switched from `std::tuple` to `STDEXEC::__tuple`, this could be simply:

```suggestion
            return STDEXEC::__apply(factory_, std::move(args_), std::move(rcvr));
```
`STDEXEC::__apply(fn, tupl, extra_args...)` is equivalent to `fn(extra_args..., __get<I>(tupl)...)`

include/exec/function.hpp
```diff
@@ -0,0 +1,402 @@
...
+    };
+
+    template <class... Args>
+    struct _sigs_from<void(Args...) noexcept>
+    {
+      using type = STDEXEC::completion_signatures<STDEXEC::set_stopped_t(), STDEXEC::set_value_t()>;
+    };
+
+    template <class Sig>
+    using _sigs_from_t = _sigs_from<Sig>::type;
```
ericniebler: i think you could replace all of the above with:
```c++
template <class Return, bool NoExcept>
using _sigs_from_t =
  STDEXEC::__concat_completion_signatures_t<
    STDEXEC::completion_signatures<STDEXEC::__single_value_sig_t<Return>,
                                   STDEXEC::set_stopped_t()>,
    STDEXEC::__eptr_completion_unless_t<NoExcept>>;
```

include/exec/function.hpp
```diff
@@ -0,0 +1,402 @@
...
+      using sender_concept = SndrCncpt;
+
+      // TODO: I only know this works for empty lambdas; figure out whether function pointers
+      //       and/or pointer-to-member functions can be made to work
+      template <STDEXEC::__callable<Args...> Factory>
+        requires STDEXEC::__not_decays_to<Factory, _func_impl>  //
+              && STDEXEC::__std::constructible_from<Factory>    //
+              && STDEXEC::__callable<Factory, Args...>
+              && STDEXEC::sender_to<STDEXEC::__invoke_result_t<Factory, Args...>, _receiver_t>
+      constexpr explicit _func_impl(Args &&...args, Factory &&factory)
```
ericniebler:
```suggestion
      constexpr explicit _func_impl(Args &&...args, Factory)
```
you don't want to deduce `Factory` to be a reference type here. references are not default constructible.

ericniebler (commented):
i still don't understand why this utility captures args and a function that returns a sender instead of just capturing the resulting sender. what benefit do we gain from lazily constructing the sender at `connect` time?

ispeters:
> i still don't understand why this utility captures args and a function that returns a sender instead of just capturing the resulting sender. what benefit do we gain from lazily constructing the sender at `connect` time?

Yeah, sorry, I was responding to the easier feedback before getting to this philosophical question and it's not surprising you have this question—somewhere between having the idea for this PR and actually publishing it, I got distracted by the details and didn't include a very good motivation. Let me try to fix that.

The back story is that I'm looking into the benchmark that Vinnie and Steve have published [here](https://github.com/sgerbino/capy/tree/pr%2Fbeman-bench/bench%2Fbeman). I'm still working on developing a deep enough understanding of the benchmark to have intelligent opinions about it but my initial impression is that the coroutine code is very well written (unsurprising given the author's expertise), but the sender code is unidiomatic, suggesting to me that the author is less experienced with senders than with coroutines. I'd like to improve the sender code to make the comparison between approaches as fair as possible.

One of the apparent weaknesses of senders that is exposed in the benchmark is that it's more difficult to optimize the allocation patterns of an `any_sender_of` than that of a coroutine task type (`capy::task` in the benchmark, but I think it generalizes). There are probably several ways to tackle this problem, but capturing arguments plus a sender factory is an idea I had while ruminating on the issue that I think could prove very useful.

The idea behind this `exec::function` is that it's sort of a "better coroutine". By capturing an arg tuple + sender factory, the initial sender never has to allocate, although, because we're still doing type erasure, we obviously have to support allocating the operation state modulo any small buffer optimization. Ensuring the opstate allocation happens through the newly-introduced "frame allocator", lets this type benefit from the same allocation optimizations that Vinnie and co. have made in Capy. If the frame allocator is what Vinnie calls a "recycling allocator" then the stack-like nature of nested operations can be reflected in a stack-like allocation pattern that leads to amortized zero allocations per operation.

The usage I expect would look like this:
```c++
struct interface {
  virtual exec::function<int(interface*) noexcept> get_int() noexcept = 0;
};

struct impl : interface {
  exec::function<int(interface*) noexcept> get_int() noexcept override {
    return exec::function<int(interface*) noexcept>(this, [](interface* base) noexcept {
      auto* self = static_cast<impl*>(base);
      // presumably some more interesting composition of senders in practice
      return ex::just(self->i_);
    });
  }

 private:
  int i_;
};
```
Letting myself think grandiose thoughts, I could imagine extending this with a language feature that lets you use coroutine syntax to let the compiler write the actual sender for you, like this:
```c++
exec::function<std::string(int)> async_to_string(int i) {
  co_return std::format("{}", i);
}
```
For cases where `async_to_string` is declared in a header and defined in a `.cpp` file, the type erasure, with the concomitant allocation, can't really be avoided, but for cases where the compiler can see the body of `async_to_string` at the call site, I could imagine the compiler generating a parent operation that has space in its opstate for the child opstate, eliding any allocations. Imagining even further, it'd be nice to have an "auto"-like syntax for inline async functions where the compiler not only generates the sender implementation, but also computes the full set of completion signatures to make the declaration more terse. Maybe something like
```c++
exec::function<ex::sender_tag> async_to_string(int i) {
  // compiler computes set_value_t(std::string) and set_error_t(std::exception_ptr)
  // but not set_stopped_t() because this coroutine doesn't complete with stopped
  co_return std::format("{}", i);
}
```

ericniebler:
> By capturing an arg tuple + sender factory, the initial sender never has to allocate

so the allocation this saves is the one that can potentially happen when type-erasing a sender? could you achieve the same thing with a (non-type-erasing) sender adaptor that returns a type-erased opstate from `connect`?

ispeters:
> > By capturing an arg tuple + sender factory, the initial sender never has to allocate
> 
> so the allocation this saves is the one that can potentially happen when type-erasing a sender?

Yes.

> could you achieve the same thing with a (non-type-erasing) sender adaptor that returns a type-erased opstate from `connect`?

I don't think so, but maybe you can see a way? Suppose we had `sender auto exec::type_erase(sender auto&&)`; I think you can't get around the fact that the input sender is of arbitrary size so either the return type has to fully specify the type of the input plus whatever wrapping it does, or you have to type-erase the input and risk an eager allocation.

Also, fwiw, I'm open to feedback on the name of this type. If you look at the commit history, you can see I originally named this thing `capy_sender` because the immediate application is the Capy-sourced benchmark. Then, I thought `io_sender` would be a better name since that's what Vinnie claims this approach is best for. I finally landed on `function` because:
 * Robert's description of how senders work describes how you first curry your arguments into a sender, then you allocate a frame in the form of an operation state, and then you actually execute the async function by invoking `start` so this type implements a kind of platonic sender—curry, allocate, execute—which suggests that it's a type-erased async function; and
 * Robert has also said that he believes everything in the `std::execution` namespace is presumptively async so `std::execution::async_function<void()>` is redundant—we can just have `std::execution::function<void()>`.

That said, the confusion you expressed earlier about the interface suggests to me that the name might be a problem. Perhaps something like `curried<void()>`, `sender_factory<void()>`.

ericniebler:
> Suppose we had sender auto exec::type_erase(sender auto&&); I think you can't get around the fact that the input sender is of arbitrary size so either the return type has to fully specify the type of the input plus whatever wrapping it does

yes, that is what i was asking about. what is the problem with doing an ordinary sender adaptor?

ispeters:
> > Suppose we had sender auto exec::type_erase(sender auto&&); I think you can't get around the fact that the input sender is of arbitrary size so either the return type has to fully specify the type of the input plus whatever wrapping it does
> 
> yes, that is what i was asking about. what is the problem with doing an ordinary sender adaptor?

I'm not certain whether you're asking, "why is `function` a type instead of an algorithm?", or "why does this thing—whether type or algorithm—have to type-erase a factory instead of adapting a concrete sender?"

Regarding "why a type?", I'm not really fussed either way, except that the motivating use case requires that users be able to spell a type so they can return it from functions declared-but-not-defined in headers. So if you're looking for an algorithm to be the "constructor" instead of using a regular type with a constructor, that's fine by me, but I think there needs to be a type regardless.

Regarding "why does this type-erase the factory?", I can't see another way to achieve my goal, but I'm open to counterexamples. I want to be able to write this sort of thing:
```c++
// iface.hpp
#include <memory>

struct iface {
  virtual ~iface() = default;

  virtual $type$ async_virtual(int idx) = 0;
};

std::unique_ptr<iface> make_iface();

// impl.hpp
#include "iface.hpp"

struct impl : iface {
  $type$ async_virtual(int idx) override;

  $type$ async_member();
};

// impl.cpp
#include "impl.hpp"

#include <execution>

sender auto some_sender(int idx) {
  // some arbitrary composition of senders
}

sender auto other_sender() {
  // some other arbitrary composition of senders
}

$type$ impl::async_virtual(int idx) {
  // return some function of some_sender(idx);
}

$type$ impl::async_member() {
  // return some function of other_sender();
}

std::unique_ptr<iface> make_iface() {
  return std::unique_ptr<impl>();
}

// main.cpp
#include "impl.hpp":

int main() {
  auto p = make_iface();

  sync_wait(p->async_virtual(42));

  auto* q = static_cast<impl*>(p.get());

  // async_member's declaration is visible, but not its definition
  sync_wait(q->async_member());
}
```

If we imagine we have `sender auto type_erase(sender auto&&)` and use it in `impl.cpp`, I think you have this:
```c++
$type$ impl::async_virtual(int idx) {
  return type_erase(some_sender(idx));
}

$type$ impl::async_member() {
  return type_erase(other_sender());
}
```

Given the declaration of `some_sender` and `other_sender` are not visible outside `impl.cpp`, `main.cpp` doesn't know how big the concrete sender returned from `type_erase` is, so I think you inevitably have to do the standard small buffer-optimized type-erasure dance and risk an allocation. The only alternative I can see is for the specific types of the composed senders to leak out of the interface through something equivalent to `using $type$ = decltype(type_erase(...));`, which I think makes it inappropriate for use as the return type of virtual member functions, or separately-compiled functions of any kind.

If `$type$` is an instantiation of the proposed `function`, then `impl.cpp` looks like this (assuming the senders returned from `some_sender` and `other_sender` complete with `set_value_t()` for simplicity):

```c++
exec::function<void(int)> impl::async_virtual(int idx) {
  return exec::function<void(int)>(
      std::move(idx), // it bugs me that this move is required in the current code…
      [](int idx) { return some_sender(idx); });
}

exec::function<void()> impl::async_member() {
  return exec::function<void()>([] { return other_sender(); });
}
```
The type-erased sender factory is of constant size (because I restrict the factory to be trivially-copyable types no bigger than two pointers to accommodate regular function pointers and member function pointers), and the storage for the factory's arguments is knowable and a function of the class template parameters because they're expressed in the function type.

Does that answer your question? Do you see another way to accomplish the goal? A `function<int(???)>` is like a `task<int>` but with better-timed allocation because it erases less.

ericniebler:
thanks, i understand now what problem you're solving. it's very interesting. :thinking:

ispeters:
The test failure on GCC 14 Debug with ASAN enabled is going to be tricky; GCC doesn't support ASAN on Apple silicon, which is my only dev environment.

ispeters:
Unfortunately for me, the (gcc 14, Debug, ASAN) failure doesn't repro with Homebrew Clang 22.1.4, Debug, ASAN.

The failures are in `test_function.cpp` here:
https://github.com/NVIDIA/stdexec/blob/67066b85227b88dc5e9344e1c85209c925dfa9e9/test/exec/test_function.cpp#L374-L375
https://github.com/NVIDIA/stdexec/blob/67066b85227b88dc5e9344e1c85209c925dfa9e9/test/exec/test_function.cpp#L391
and here:
https://github.com/NVIDIA/stdexec/blob/67066b85227b88dc5e9344e1c85209c925dfa9e9/test/exec/test_function.cpp#L465-L466
https://github.com/NVIDIA/stdexec/blob/67066b85227b88dc5e9344e1c85209c925dfa9e9/test/exec/test_function.cpp#L478

I'm wondering if GCC's ASAN implementation is mucking with this `operator==` implementation in some way, like initializing the `std::byte make_sender_[sizeof(void *) * 2]` member with garbage, or something:
https://github.com/NVIDIA/stdexec/blob/67066b85227b88dc5e9344e1c85209c925dfa9e9/include/exec/function.hpp#L199-L205

include/exec/function.hpp
```diff
@@ -0,0 +1,627 @@
...
+// TODO: split this header into pieces
+#include "any_sender_of.hpp"
+
+#include <cstddef>
+#include <cstring>
+#include <memory>
+
+// This file defines function<ReturnType(Arguments...)>, which is a
+// type-erased sender that can complete with
+//  - set_value(ReturnType&&)
```
ericniebler: i would prefer if it completed with `ReturnType` instead of `ReturnType&&`.

include/exec/function.hpp
```diff
@@ -0,0 +1,627 @@
...
+        return get_frame_allocator(env);
+      }
+      else if constexpr (requires { get_allocator(env); })
+      {
+        return get_allocator(env);
+      }
+      else
+      {
+        return std::allocator<std::byte>();
+      }
```
ericniebler: could be:
```c++
inline constexpr auto choose_frame_allocator = STDEXEC::__first_callable{
  get_frame_allocator,
  get_allocator,
  STDEXEC::__always{std::allocator<std::byte>()};
```
`__first_callable` is handy for these sorts of things.

include/exec/function.hpp
```diff
@@ -0,0 +1,627 @@
...
+    {
+      return ((__get<I>(lhs) == __get<I>(rhs)) && ...);
+    }
+
+    template <class... Args>
+    bool _equal(__tuple<Args...> const &lhs, __tuple<Args...> const &rhs)
+      noexcept(noexcept(_equal(std::index_sequence_for<Args...>{}, lhs, rhs)))
+    {
+      return _equal(std::index_sequence_for<Args...>{}, lhs, rhs);
+    }
```
ericniebler: maybe we should just add `operator==(__tuple const&, __tuple const&) = default;` to the `__tuple` structs.

include/exec/function.hpp
```diff
@@ -0,0 +1,627 @@
...
+        {
+          auto &make_sender = *__std::start_lifetime_as<Factory>(storage);
+
+          auto alloc = choose_frame_allocator(get_env(rcvr));
+
+          return _any::_any_opstate_base(__in_place_from,
+                                         std::allocator_arg,
+                                         alloc,
+                                         STDEXEC::connect,
+                                         std::invoke(make_sender, static_cast<Args &&>(args)...),
```
ericniebler: prefer `STDEXEC::__invoke` over `std::invoke`. it compiles 2x faster. (also use `__invocabable` above.)

include/exec/function.hpp
```diff
@@ -0,0 +1,627 @@
...
+        {
+          return queries<__msplice<sigs[Is]>...>();
+        };
+
+        return fn(__make_indices<sigs.size()>());
+      }
+    };
+
+    template <>
+    struct _canonical_fn<queries<>>
```
ericniebler: i think the specialization above should handle this case as well.

include/exec/function.hpp
```diff
@@ -0,0 +1,627 @@
...
+  // that invoking this specialization with Return set to sender_tag is a bug...
+  //
+  // the same question applies to all the specializations below that take explicit
+  // completion signatures
+  struct function<Return(Args...)>
+    : _func::_func_impl<STDEXEC::sender_tag(Args...), _func::_sigs_from_t<Return, false>, queries<>>
+  {
+    using base = _func::_func_impl<STDEXEC::sender_tag(Args...),
+                                   _func::_sigs_from_t<Return, false>,
+                                   queries<>>;
```
ericniebler, line 382: is `base` a part of `function`'s public-facing interface?

include/exec/function.hpp
```diff
@@ -0,0 +1,627 @@
...
+
+    using base::base;
```
ericniebler, line 388: i think everything from here down can be removed. the defaults are correct.

include/exec/function.hpp
```diff
@@ -0,0 +1,627 @@
...
+    }
+  };
+}  // namespace experimental::execution
+
+namespace exec = experimental::execution;
+
+namespace std
+{
+  template <class... FuncArgs, template <class> class TQual, template <class> class UQual>
+  struct basic_common_reference<exec::function<FuncArgs...>,
```
ericniebler, line 491: why is this needed?

ericniebler (commented):
comments inline..

include/exec/function.hpp
```diff
@@ -0,0 +1,627 @@
...
+
+     public:
+      using sender_concept = SndrCncpt;
+
+      template <std::invocable<Args...> Factory>
+        requires STDEXEC::__not_decays_to<Factory, _func_impl>  //
+              && (STDEXEC_IS_TRIVIALLY_COPYABLE(Factory))       //
+              && (sizeof(Factory) <= sizeof(make_sender_))      //
+              && STDEXEC::sender_to<STDEXEC::__invoke_result_t<Factory, Args...>, _receiver_t>
+      constexpr explicit _func_impl(Args &&...args, Factory factory)
```
ericniebler, line 180: i would expect the factory to be the first parameter, followed by the arguments.

ericniebler:
> I'm wondering if GCC's ASAN implementation is mucking with this `operator==` implementation in some way, like initializing the `std::byte make_sender_[sizeof(void *) * 2]` member with garbage, or something:

i think you're accidentally comparing padding bits for equality. `ex::just` is 1 byte, but its value could be anything from 0x00 to 0xFF since the type is empty.

include/exec/function.hpp
ispeters: This comment is a bit ambiguous; the `completion_signatures` will contain `set_value_t(ReturnType)`; the reference here to `set_value(ReturnType&&)` is just the resulting signature of the `set_value` overload that will accept a value of type `ReturnType`. Robert had the same concern, though, so I obviously need to update the comment.

include/exec/function.hpp
ispeters: Or maybe I should remove `operator==`; it turns out `std::function` doesn't have it, except to compare to `nullptr_t` as a way of detecting an empty `function`: https://godbolt.org/z/s39axTPTs.

include/exec/function.hpp
ispeters: Perhaps it ought to, but it doesn't:
```c++
In file included from /Users/ianpetersen/git/stdexec/test/exec/test_function.cpp:18:
/Users/ianpetersen/git/stdexec/include/exec/function.hpp:311:14: error: no member named 'erase' in 'std::execution::__static_vector<std::execution::__type_index, 0>'
  311 |         sigs.erase(end, sigs.end());
      |         ~~~~ ^
/Users/ianpetersen/git/stdexec/include/exec/function.hpp:319:31: note: in instantiation of member function 'experimental::execution::_func::_canonical_fn<experimental::execution::queries<>>::get_sigs' requested here
  319 |         constexpr auto sigs = get_sigs();
      |                               ^
/Users/ianpetersen/git/stdexec/include/exec/function.hpp:345:35: note: in instantiation of member function 'experimental::execution::_func::_canonical_fn<experimental::execution::queries<>>::operator()' requested here
  345 |     using _canonical_t = decltype(_canonical<Sigs>());
      |                                   ^
/Users/ianpetersen/git/stdexec/include/exec/function.hpp:487:32: note: in instantiation of template type alias '_canonical_t' requested here
  487 |                         _func::_canonical_t<queries<Queries...>>>
      |                                ^
/Users/ianpetersen/git/stdexec/test/exec/test_function.cpp:81:47: note: in instantiation of template class 'experimental::execution::function<void (), experimental::execution::queries<>>' requested here
   81 |       exec::function<void(), exec::queries<>> sndr([]() noexcept { return ex::just(); });
      |                                               ^
1 error generated.
```
I'm hardly familiar with the compile-time facilities in stdexec; maybe I'm doing something wrong?

include/exec/function.hpp
ispeters: I don't know. If it is, it probably shouldn't be called `base`, but I'm undecided about whether there ought to be a way to access the base class's type; the benefit of including it in the interface would be that users could use it to access the canonicalized lists of completion signatures and environment queries, but maybe there's a better way to do that, or maybe its unnecessary.

ispeters:
> > I'm wondering if GCC's ASAN implementation is mucking with this `operator==` implementation in some way, like initializing the `std::byte make_sender_[sizeof(void *) * 2]` member with garbage, or something:
> 
> i think you're accidentally comparing padding bits for equality. `ex::just` is 1 byte, but its value could be anything from 0x00 to 0xFF since the type is empty.

I think you're right. I realized I've assumed that `is_trivially_copyable` implies "can be compared for equality with `memcmp`", which is probably not true.

I can see two ways to resolve this:
1. add a function pointer to the implementation that points at an `is_equal` function that does a type-safe comparison, or
2. follow `std::function`'s example and remove equality-comparableness.

Do you have an opinion on which direction makes more sense?

include/exec/function.hpp
ispeters: So, I'm open to that, but my reason for the current order is that it puts the arguments closer to the argument list of a lambda than the other order and, since `Args...` is deduced for the class rather than the constructor, there's no parsing ambiguity by doing it in this order.

Regarding "putting the arguments closer", I mean this:
```c++
function<int(foo, bar, baz, quux)> async(foo f, bar b1, baz b2, quux q) {
  return function<int(foo, bar, baz, quux)>(
    std::move(f), std::move(b1), std::move(b2), std::move(q),
    [](auto f, auto b1, auto b2, auto q) {
      // the function ctor arguments line up more nicely with this
      // lambda's arguments in this order, rather than having the
      // ctor arguments after the closing brace, far away from the
      // lambda's arguments
      return sender_algo(std::move(f), std::move(b1), std::move(b2), std::move(q));
    });
}
```

include/exec/function.hpp
ispeters: I added the extra details to make the following assertion work:
```c++
static_assert(std::assignable_from<
    function<void() noexcept>&,
    function<sender_tag(), completion_signatures<set_value_t(), set_stopped_t()>>>);
```

Those two `function` types are different types but, due to the canonicalization of the completion signatures and environment queries, they both derive from the same base class and are thus mutually assignable to each other but the `assignable` concept requires that `lhs = rhs` be valid and of type `decltype(lhs)` (i.e. `function<void() noexcept>::operator=` has to return `function<void() noexcept>&`). The assignment operator that could be inherited from the base class returns a reference-to-base, which breaks the satisfaction of `assignable`.

ispeters: Oh, that said, I think I can reduce duplication with a CRTP mix-in that expresses these declarations in one place.

include/exec/function.hpp
ispeters: This is also for `std::assignable_from` satisfaction.

include/exec/function.hpp
ispeters: We settled on deleting `operator==` instead of fixing it.

include/exec/function.hpp
ispeters: I added some dummy implementations of the missing member functions (`erase` and `operator[]`) to `__static_vector<T, 0>` so I don't need a special case anymore.

include/exec/function.hpp
ericniebler: "when in doubt, leave it out (of the public interface)"

ispeters: I'm using the existence of the publicly-accessible alias in the tests to confirm that different `function` specializations have the same base class; I think the tests are useful because that common descent is an important part of the interface. Would you recommend making the alias actually private and breaking that encapsulation in the tests somehow (does Catch have something like GTest's `FRIEND_TEST`?), or changing the name to something like `_base` to hint at logical privacy?

ispeters: Hmm… thinking about it a little more, the common descent is an implementation detail that makes interassignability work, which _is_ part of the interface. I can just delete the tests that depend on `base` being public and rely on the tests that validate interassignability to catch breakage to the thing I care about.

include/exec/function.hpp
ericniebler: have you considered an interface like:
```c++
function<int(foo, bar, baz, quux)> async(foo f, bar b1, baz b2, quux q) {
  return make_function(
    sender_algo, std::move(f), std::move(b1), std::move(b2), std::move(q));
}
```

it would be defined something like:
```c++
template <class Queries, class SenderFactory, class... Args>
using __result_t =
  value_types_of_t<__invoke_result_t<SenderFactory, Args...>,
                   __env_for_queries<Queries>,
                   type_identity_t,
                   type_identity_t>;

template <class Queries = queries<>, class SenderFactory, class... Args>
auto make_function(SenderFactory algo, Args... args)
  -> function<__result_t<Queries, SenderFactory, Args...>(Args...)>;
```

ericniebler:
~~#2065~~ #2066 should fix the clang-16 failures.

ispeters:
I've rebased locally onto main; I'm running a Clang 16 build to see if it's fixed.

include/exec/function.hpp
```diff
@@ -0,0 +1,529 @@
...
+
+  namespace _func
+  {
+    using namespace STDEXEC;
+
+    template <class Receiver, class Sigs, class... Queries>
+    class _func_op;
+
+    // The concrete operation state resulting from connecting a function<...> to a concrete
+    // receiver of type Receiver. This type manages a dynamically-allocated _derived_op instance,
```
ispeters: This should refer to `_any_opstate_base`.
```suggestion
    // receiver of type Receiver. This type manages an _any::_any_opstate_base instance,
```

PR #2051, opened 2026-04-27
as_awaitable: potential use-after-free when calling notify_one

maikel:
This PR does two things:

1. I believe there is currently a race between atomic<T>::fetch_sub and the call to atomic<T>::notify_one. By the time the await_suspend thread reaches the line with notify_one the receiver could have a resumed and free'd the coro frame. This is pretty critical. Maybe we write a relacy test that checks the issue.

2. @ispeters proposed an optimization to save state for the sender_awaitable. Instead of using a ref counter this improved algorithm uses thread ids directly. That's awesome in any case.

I ask @ispeters for review. Lets check whether (1) is a real issue.

### DIFF
--- a/include/stdexec/__detail/__as_awaitable.hpp
+++ b/include/stdexec/__detail/__as_awaitable.hpp
@@ -26,6 +26,7 @@
 #include "__connect.hpp"
 #include "__meta.hpp"
 #include "__queries.hpp"
+#include "__spin_loop_pause.hpp"
 #include "__type_traits.hpp"
 #include "__variant.hpp"
 
@@ -145,8 +146,24 @@ namespace STDEXEC
     template <class _Value>
     struct __sender_awaiter_base<_Value, false> : __sender_awaiter_base<_Value, true>
     {
-      __std::atomic<int>    __refcount_{2};
-      std::thread::id const __starting_thread_{std::this_thread::get_id()};
+      // This is used to coordinate between await_suspend (T1) and the receiver (T2).
+      // It can hold three different values:
+      //
+      //  T1_id  - await_suspend stored the awaiting thread's id before calling start().
+      //           The receiver has not completed yet.
+      //
+      //  {}     - The empty/cleared state. Two cases depending on who wrote it:
+      //           (a) The receiver completed on the same thread: await_suspend should
+      //               resume the continuation directly.
+      //           (b) await_suspend wrote it after start() returned (its last frame
+      //               write): a cross-thread receiver that is spinning will observe
+      //               this and proceed to call resume().
+      //
+      //  T2_id  - A cross-thread receiver (T2 != T1) wrote its own id here before
+      //           await_suspend did its exchange. T2 is spinning on load() until it
+      //           observes {} (written by T1's exchange). No notify_one is used;
+      //           the value change itself is the signal.
+      __std::atomic<std::thread::id> __thread_id_{std::this_thread::get_id()};
     };
 
     template <class _Value>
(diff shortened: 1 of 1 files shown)


include/stdexec/__detail/__as_awaitable.hpp
```diff
@@ -289,45 +309,32 @@ namespace STDEXEC
...
-
-      constexpr auto
+      STDEXEC_CONSTEXPR_CXX23 auto
       await_suspend([[maybe_unused]] __std::coroutine_handle<> __continuation) noexcept
         -> __std::coroutine_handle<>
       {
         STDEXEC_ASSERT(this->__continuation_.handle() == __continuation);
 
+        std::thread::id const __current_id = std::this_thread::get_id();
+        this->__thread_id_.store(__current_id, std::memory_order_relaxed);
```
ispeters: This is the initial value, right? I think this store is unnecessary.
```suggestion
```

include/stdexec/__detail/__as_awaitable.hpp
```diff
@@ -289,45 +309,32 @@ namespace STDEXEC
...
-        int const __old_refcount = this->__refcount_.fetch_sub(1, __std::memory_order_acq_rel);
-
-        if (__old_refcount == 1)
+        // This exchange({}) is T1's last write to the frame. After this point:
+        //  - If T2 is spinning waiting for our exchange, it will observe {} and
+        //    proceed to resume().
+        //  - If T2 hasn't run yet, it will see {} from its load in __done() and
+        //    skip the spin entirely
+        std::thread::id const __old_id = this->__thread_id_.exchange(std::thread::id{},
+                                                                     std::memory_order_acq_rel);
```
ispeters: I think this can be a store-release; I'd be curious to see whether TSAN agrees.

There are three cases to consider:
1. `__old_id` is `{}`; the write happened on the same thread so the store is irrelevant and the load-relaxed is fine;
2. `__old_id` is `std::this_thread::get_id()`; the continuation is still running so we need a store-release (although I'm not sure why—I'd be curious to see if relaxed passes TSAN), and the load is irrelevant; or
3. `__old_id` is the completion thread's ID; similar to above—we need to unblock the spin wait with a store-release but don't care about the load.

include/stdexec/__detail/__as_awaitable.hpp
```diff
@@ -240,31 +257,34 @@ namespace STDEXEC
...
+          // Completed on the same thread as the awaiter. Clear __thread_id_ so that
+          // await_suspend sees {} after start() returns and resumes the continuation.
+          // This cannot race with await_suspend because await_suspend is still on the same thread.
+          __awaiter.__thread_id_.store(std::thread::id{}, std::memory_order_relaxed);
+          return;
         }
-        else if (__old_refs == 1)
+        // Completing on a different thread. Write our id so await_suspend knows we already ran.
+        std::thread::id const __old_id = __awaiter.__thread_id_.exchange(__current_id,
+                                                                         std::memory_order_acq_rel);
```
ispeters: I think this can be an acquire-load; if we're running this code then the current thread is going to be responsible for finalizing the work of the sender, so we need information to flow from the awaiter to the completer, but not the other way.
```suggestion
                                                                         __std::memory_order_acquire);
```

include/stdexec/__detail/__as_awaitable.hpp
```diff
@@ -240,31 +257,34 @@ namespace STDEXEC
...
+        if (__old_id != std::thread::id{})
         {
-          // await_suspend already decremented first (refcount was 1 when we decremented,
-          // meaning await_suspend had already brought it from 2 to 1). Resume now.
-          STDEXEC::__coroutine_resume_nothrow(__awaiter.__get_continuation());
+          // await_suspend is between start() returning and its own exchange({}).
+          // That exchange is T1's last write to the frame. Spin until we observe
+          // it so that we don't call resume() while T1 is still touching the frame.
+          // The window is just a few instructions, so a yield-spin is fine.
+          while (__awaiter.__thread_id_.load(std::memory_order_acquire) == __current_id)
```
ispeters: If you make the condition `!= std::thread::id{}`, the underlying comparison will be with zero, rather than an arbitrary number; does that make any difference?
```suggestion
          while (__awaiter.__thread_id_.load(std::memory_order_acquire) != std::thread::id{})
```

ispeters (approved):
Looks good to me! I left a few nit-picky comments; I'd be curious to see the TSAN results if we relaxed the memory ordering a little, but I think the current memory orderings are safe.

include/stdexec/__detail/__as_awaitable.hpp
```diff
@@ -240,31 +257,34 @@ namespace STDEXEC
...
-        // await_suspend can determine the correct action from a single fetch_sub without
-        // any secondary reads that could race with frame destruction.
-        int const __decrement = __on_other_thread ? 2 : 1;
-        int const __old_refs  = __awaiter.__refcount_.fetch_sub(__decrement,
-                                                               __std::memory_order_acq_rel);
-        if (__on_other_thread && __old_refs == 2)
+        auto&                 __awaiter     = static_cast<__awaiter_t&>(this->__awaiter_);
+        std::thread::id const __current_id  = std::this_thread::get_id();
+        std::thread::id const __awaiting_id = __awaiter.__thread_id_.load(
+          std::memory_order_relaxed);
```
ispeters: I think your build error is the missing `__`.
```suggestion
          __std::memory_order_relaxed);
```

include/stdexec/__detail/__as_awaitable.hpp
```diff
@@ -240,31 +257,34 @@ namespace STDEXEC
...
         {
-          // We decremented first on a different thread. await_suspend will observe
-          // old==0 and return noop_coroutine; it will call notify_one to signal us.
-          // Wait until await_suspend has decremented, then resume the continuation.
-          __awaiter.__refcount_.wait(0, __std::memory_order_acquire);
-          STDEXEC::__coroutine_resume_nothrow(__awaiter.__get_continuation());
+          // Completed on the same thread as the awaiter. Clear __thread_id_ so that
+          // await_suspend sees {} after start() returns and resumes the continuation.
+          // This cannot race with await_suspend because await_suspend is still on the same thread.
+          __awaiter.__thread_id_.store(std::thread::id{}, std::memory_order_relaxed);
```
ispeters:
```suggestion
          __awaiter.__thread_id_.store(std::thread::id{}, __std::memory_order_relaxed);
```

include/stdexec/__detail/__as_awaitable.hpp
ispeters: Regardless, you need `__`.
```suggestion
          while (__awaiter.__thread_id_.load(__std::memory_order_acquire) == __current_id)
```

include/stdexec/__detail/__as_awaitable.hpp
ispeters: Oops, meant to leave a suggested diff to make the suggestion explicit.
```suggestion
                                                                     __std::memory_order_release);
```

ispeters (commented):
I think you need some leading `__` to appease the GPU builds.

include/stdexec/__detail/__as_awaitable.hpp
maikel: Is this guaranteed. Like always?

maikel: I guess, you are right because its an internal type that is only used when co_awaiting senders?

include/stdexec/__detail/__as_awaitable.hpp
maikel: I will test the memory orderings, including making it relaxed for this algorithm.

include/stdexec/__detail/__as_awaitable.hpp
maikel: yes. you are right. My GPU-fu is very weak

include/stdexec/__detail/__as_awaitable.hpp
maikel: I don't think it makes any difference. I will make the change

include/stdexec/__detail/__as_awaitable.hpp
maikel: Removed!

include/stdexec/__detail/__as_awaitable.hpp
maikel: Changed to release. Relaxed will still emit TSAN warnings. I believe that the particular TSAN warning is a false-positive but not sure whether its worth it fighting that.

include/stdexec/__detail/__as_awaitable.hpp
maikel: Changed

PR #2063, opened 2026-05-04
shrink the size of the type-erasing wrappers

ericniebler:
`__any`-based type-erasing wrappers are wasting space by over-aligning the internal buffer by default. this PR changes the default buffer alignment from `alignof(std::max_align_t)` to `alignof(void*)`. this eliminates two pointers worth of padding in the representation of `__any<ifooable>`.

### DIFF
--- a/include/stdexec/__detail/__any.hpp
+++ b/include/stdexec/__detail/__any.hpp
@@ -186,7 +186,7 @@ namespace STDEXEC::__any
             class _Base,
             class _BaseInterfaces   = __extends<>,
             size_t _BufferSize      = __default_buffer_size,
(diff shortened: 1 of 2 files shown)


include/stdexec/__detail/__any.hpp
```diff
@@ -186,7 +186,7 @@ namespace STDEXEC::__any
             class _Base,
             class _BaseInterfaces   = __extends<>,
             size_t _BufferSize      = __default_buffer_size,
-            size_t _BufferAlignment = alignof(std::max_align_t)>
+            size_t _BufferAlignment = alignof(void *)>
```
romintomasetti, line 189: Would it be worth adding a small test that checks the `sizeof` ?

PR #2068, opened 2026-05-05
run_loop: add public get_run_loop() accessor on scheduler

guptapratykshh:
Add a public `get_run_loop()` accessor on `__basic_run_loop::__scheduler` so consumers holding a `run_loop::scheduler` can recover the owning `run_loop&` through a public API.

### DIFF
--- a/include/stdexec/__detail/__run_loop.hpp
+++ b/include/stdexec/__detail/__run_loop.hpp
@@ -244,6 +244,17 @@ namespace STDEXEC
        public:
         using scheduler_concept = scheduler_tag;
 
+        // Returns a reference to the parent run_loop. The derived type is
+        // recovered via static_cast from the stored __basic_run_loop pointer.
+        // This is provided so that consumers holding only a `scheduler` can
+        // recover the owning loop without reaching into the schedule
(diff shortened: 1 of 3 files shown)


include/stdexec/__detail/__run_loop.hpp
```diff
@@ -244,6 +244,17 @@ namespace STDEXEC
        public:
         using scheduler_concept = scheduler_tag;
 
+        // Returns a reference to the parent run_loop. The derived type is
+        // recovered via static_cast from the stored __basic_run_loop pointer.
+        // This is provided so that consumers holding only a `scheduler` can
+        // recover the owning loop without reaching into the schedule
+        // sender's environment for the private __loop_ member.
+        STDEXEC_ATTRIBUTE(nodiscard, host, device)
```
romintomasetti, line 252: I think a test to demonstrate your use case would help.

ericniebler:
this change sets off my library-design Spidey sense. i don't think it is a good idea. you really don't want to give every bit of code an easy way to, e.g., shut down the main run loop. someone will do it. the separation between execution context and scheduler is deliberate.

PR #2069, opened 2026-05-06
unify the storage of intermediate results

ericniebler:
many async operations need to store the results of a child operation and send the results to a receiver later. in many places we are computing the type of a `__variant` of `__tuple`s for holding these results, and then using a visitor to send the results. it's a lot of code repetition.

unify all of that into a `__results_storage<Signatures...>` type that has a `.__complete(receiver)` member function that encapsulates the visitation.

### DIFF
--- a/include/stdexec/__detail/__storage.hpp
+++ b/include/stdexec/__detail/__storage.hpp
@@ -1,0 +1,86 @@
+/*
+ * Copyright (c) 2021-2025 NVIDIA Corporation
+ *
+ * Licensed under the Apache License Version 2.0 with LLVM Exceptions
+ * (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ *   https://llvm.org/LICENSE.txt
+ *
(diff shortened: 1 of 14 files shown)


include/stdexec/__detail/__storage.hpp
```diff
@@ -0,0 +1,86 @@
...
+    {
+      template <class _Receiver, class _Tuple>
+      constexpr void operator()(_Receiver& __rcvr, _Tuple&& __tupl) noexcept
+      {
+        __apply(__applier(), static_cast<_Tuple&&>(__tupl), __rcvr);
+      }
+    };
+
+    // Add storage for an exception_ptr if the result datums are not all nothrow
+    // decay-copyable.
```
ispeters, line 51: Why build the decay-copy into this utility? Robert has an analogous facility that stores references where this facility would transform references into decay-copied values. Could you layer the decay-copy on top?

ericniebler: I don't know where I would use a storage type that stores references. None of the algorithms are specified to do that.

PR #2081, opened 2026-05-18
core(basic sender): fix trailing return type

romintomasetti:
This PR changes a `decltype(auto)` to a proper trailing return type, avoiding instantiation of stuff that ends up trying to instantiate stuff for incomplete types.

I've added a new test that does not compile if the trailing return type is turned back to `decltype(auto)`.

### DIFF
--- a/test/stdexec/algos/adaptors/test_write_env.cpp
+++ b/test/stdexec/algos/adaptors/test_write_env.cpp
@@ -21,6 +21,7 @@
 #include <type_traits>
 #include <utility>
 
+#include <exec/single_thread_context.hpp>
 #include <stdexec/execution.hpp>
 #include <test_common/receivers.hpp>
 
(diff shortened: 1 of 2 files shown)


romintomasetti:
@maartenarnst

romintomasetti:
@ericniebler Not sure about the quality of the trailing return type but we need it in our code base to avoid incomplete type issues. Would you have a minute to take a look ? Thks!

test/stdexec/algos/adaptors/test_write_env.cpp
```diff
@@ -21,6 +21,7 @@
 #include <type_traits>
 #include <utility>
 
+#include <exec/single_thread_context.hpp>
```
ericniebler, line 25:
```suggestion
#include <exec/completion_signatures.hpp>
#include <exec/single_thread_context.hpp>
```

test/stdexec/algos/adaptors/test_write_env.cpp
```diff
@@ -68,4 +69,115 @@ namespace
                                  receiver<state>{{}, &s});
     ::STDEXEC::start(op);
   }
+
+  template <typename IncompleteType, typename Env = STDEXEC::env_of_t<IncompleteType>>
```
ericniebler:
```suggestion
  template <class IncompleteType, class Env = STDEXEC::env_of_t<IncompleteType>>
```

test/stdexec/algos/adaptors/test_write_env.cpp
```diff
@@ -68,4 +69,115 @@ namespace
...
+  struct SenderIncomplete
+  {
+    using sender_concept = STDEXEC::sender_tag;
+
+    template <class Self, class... Env>
+    static consteval auto get_completion_signatures()
+      -> STDEXEC::__completion_signatures_of_t<STDEXEC::__copy_cvref_t<Self, Sndr>, Env...>
+    {
+      return {};
+    }
```
ericniebler, line 136:
```suggestion
    template <class Self, class... Env>
    static consteval auto get_completion_signatures()
    {
      return exec::get_child_completion_signatures<Self, Sndr, Env...>();
    }
```

PR #2083, opened 2026-05-19
exec: fork join trailing return type

romintomasetti:
To avoid issues with incomplete types.

In the same vein as:
* #2081


include/exec/fork_join.hpp
```diff
@@ -293,6 +293,9 @@ namespace experimental::execution::__fork_join
 
     static constexpr auto __connect =
       []<class _Receiver, class _Sender>(_Sender&& __sndr, _Receiver&& __rcvr) noexcept
```
ericniebler: you found a bug:
```suggestion
      []<class _Receiver, class _Sender>(_Sender&& __sndr, _Receiver __rcvr) noexcept
```

ericniebler:
thanks!

PR #2092, opened 2026-05-22
2091 compilation failed with fno exception

### LINKED ISSUE #2091: Compilation failed with -fno-exception
As title says.

To reproduce run build with `cmake -DCMAKE_CXX_FLAGS=-fno-exceptions`.

BartolomeyKant:
solves https://github.com/NVIDIA/stdexec/issues/2091

### DIFF
--- a/include/exec/asio/noexcept_boost_throw.hpp
+++ b/include/exec/asio/noexcept_boost_throw.hpp
@@ -1,0 +1,45 @@
+/*
+ * Copyright (c) 2026 NVIDIA Corporation
+ *
+ * Licensed under the Apache License Version 2.0 with LLVM Exceptions
+ * (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ *   https://llvm.org/LICENSE.txt
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#pragma once
+
+#include <exec/asio/asio_config.hpp>
+#include <stdexec/__detail/__config.hpp>
+
+/**
+ * This is a boost::throw_exception implementation provided for fno-exceptions builds
+ */
+#if STDEXEC_ASIO_USES_BOOST && STDEXEC_NO_STDCPP_EXCEPTIONS()
+
+#  include <boost/asio/thread_pool.hpp>
+#  include <boost/throw_exception.hpp>
+
+#  include <cstdlib>
+#  include <exception>
+
+namespace boost
+{
+  void throw_exception(std::exception const &)
+  {
+    std::abort();
+  }
+
+  void throw_exception(std::exception const &, boost::source_location const &)
+  {
+    std::abort();
+  }
+}  // namespace boost
+#endif
--- a/test/exec/test_function.cpp
+++ b/test/exec/test_function.cpp
@@ -103,6 +103,16 @@ namespace
       REQUIRE(fortytwo == 42);
     }
 
+    SECTION("void() from just_stopped()")
+    {
+      exec::function<void()> sndr([]() noexcept { return ex::just_stopped(); });
+
+      auto ret = ex::sync_wait(std::move(sndr));
+
+      REQUIRE_FALSE(ret.has_value());
+    }
+
+#if !STDEXEC_NO_STDCPP_EXCEPTIONS()
     SECTION("void() from throwing factory")
     {
       exec::function<void()> sndr([]() -> decltype(ex::just()) { throw "oops"; });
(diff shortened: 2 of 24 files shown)


test/exec/test_function.cpp
```diff
@@ -92,6 +92,7 @@ namespace
     }
   }
 
+ #if !STDEXEC_NO_STDCPP_EXCEPTIONS()
```
ispeters: I think this choice of scope excludes too many tests; I think it's only the scenarios that say `throw` that cause problems for `-fno-exceptions` builds. I'd prefer to have the other tests run in the no-exceptions build by narrowing the scope of this `#if` to only exclude the problematic scenarios.

ispeters (changes requested):
I don't actually have any authority here, but I'd prefer to see the changes made slightly narrower.

BartolomeyKant:
There is problems form Catch I forgot about. In local build I just remove Werror temporally.

```
/home/coder/stdexec/build/_deps/catch2-src/src/catch2/../catch2/catch_translate_exception.hpp:32:73: warning: unused parameter 'it' [-Wunused-parameter]
   32 |             std::string translate( ExceptionTranslators::const_iterator it, ExceptionTranslators::const_iterator itEnd ) const override {
      |                                                                         ^
/home/coder/stdexec/build/_deps/catch2-src/src/catch2/../catch2/catch_translate_exception.hpp:32:114: warning: unused parameter 'itEnd' [-Wunused-parameter]
   32 |             std::string translate( ExceptionTranslators::const_iterator it, ExceptionTranslators::const_iterator itEnd ) const override {
      |                                                                                                                  ^
2 warnings generated.

```

test/exec/test_function.cpp
BartolomeyKant: Do you mean exclude only throwing sections here and keep the "int() noexcept from just(42)" ?

include/exec/asio/noexcept_boost_throw.hpp
```diff
@@ -0,0 +1,45 @@
...
+ * limitations under the License.
+ */
+
+#pragma once
+
+#include <exec/asio/asio_config.hpp>
+#include <stdexec/__detail/__config.hpp>
+
+/**
+ * This is a boost::throw_exception implementation provided for fno-exceptions builds
```
BartolomeyKant, line 23: I'm not sure if it's a right way or not. 

The problem is - there is no implementation for `boost::throw_exception` is case of `-fno-exceptions` which is used in `boost::thread_pool`.

BartolomeyKant:
> There is problems form Catch I forgot about. In local build I just remove Werror temporally.
> 
> ```
> /home/coder/stdexec/build/_deps/catch2-src/src/catch2/../catch2/catch_translate_exception.hpp:32:73: warning: unused parameter 'it' [-Wunused-parameter]
>    32 |             std::string translate( ExceptionTranslators::const_iterator it, ExceptionTranslators::const_iterator itEnd ) const override {
>       |                                                                         ^
> /home/coder/stdexec/build/_deps/catch2-src/src/catch2/../catch2/catch_translate_exception.hpp:32:114: warning: unused parameter 'itEnd' [-Wunused-parameter]
>    32 |             std::string translate( ExceptionTranslators::const_iterator it, ExceptionTranslators::const_iterator itEnd ) const override {
>       |                                                                                                                  ^
> 2 warnings generated.
> ```

https://github.com/catchorg/Catch2/issues/3114

Can we suppress this warning for catch2?

test/exec/test_function.cpp
ispeters, line 135: This error doesn't require exceptions, so it could be included in a `-fno-exceptions` build, too.

test/exec/test_function.cpp
ispeters: Yep! Thanks for making the change. I noted below, though, that the `just_error(42)` case doesn't throw anything and should be included in a `-fno-exceptions` build, too.

ispeters (approved):

include/exec/asio/noexcept_boost_throw.hpp
ericniebler: rather than adding a `boost::throw_exception` function, you can use the `STDEXEC_THROW` macro.

ericniebler (dismissed):

include/exec/asio/noexcept_boost_throw.hpp
BartolomeyKant: > you can use the STDEXEC_THROW macro.

I don't get it.
Can you provide a code snippet?

The problem comes from third-party library, from `boost::thread_pool` constructor.

ericniebler: ah, i misunderstood the problem. i'll take a look when i get a sec.

BartolomeyKant:
Rebase to main.

Exclude calls for std::make_exception because despite it compiles it invokes `std::abort` in case of `-fno-exceptions`.

Also remove build relacy tests - it's not ready for `-fno-exceptions`.

BartolomeyKant:
Refactor with clang-format 21.
Clang-format 22 adds new options and formats a little different with the same config.

Also I can't reproduce the test fail.
https://github.com/NVIDIA/stdexec/actions/runs/27765872931/job/82152327268
I see `expect_error_receiver` there. https://github.com/NVIDIA/stdexec/pull/2092/changes#diff-3d9a806b6ad94214f5d9087b14ab7fdcdf186ccc290e917ea38f4b29109862d7R269
 But does it actually mean `make_exception_ptr` is calling somewhere here? 

UPD: Oh, I get it! `expect_error_receiver` expects `std::exception_ptr`.

ericniebler (approved):

PR #2100, opened 2026-05-28
Hackery to narrow down an intermittent crash

ispeters:
Repro the crash with
```sh
$ cmake -S . -B build -GNinja \
  -DCMAKE_BUILD_TYPE=Debug \
  -DCMAKE_CXX_FLAGS="-fsanitize=address" \
  -DCMAKE_CXX_STANDARD:STRING=20 \
  -DCMAKE_CXX_EXTENSIONS:BOOL=OFF \
  -DSTDEXEC_BUILD_TESTS:BOOL=ON \
  -DCMAKE_CXX_COMPILER=$(which clang++)
$ cmake --build build && ./build/test/test.stdexec
```

This is obviously not for merging; I'm just sharing my narrowed-down repro now that it doesn't depend on apple-clang. Maybe it repros on Linux or Windows, too?

PR #2102, opened 2026-05-28
Give struct __rcvr user-defined ctors in __basic_sender

ispeters:
This looks like it fixes a miscompilation leading to UB in combinations of `read_env` and `write_env` in Apple Clang.

Moots #2100.

### DIFF
--- a/include/stdexec/__detail/__basic_sender.hpp
+++ b/include/stdexec/__detail/__basic_sender.hpp
@@ -236,6 +236,16 @@ namespace STDEXEC
     using receiver_concept = receiver_tag;
     using __index_t        = __msize_t<_Idx>;
 
+    STDEXEC_ATTRIBUTE(always_inline)
+    constexpr explicit __rcvr(_State& __state) noexcept
+      : __state_(__state)
+    {}
+
+    STDEXEC_ATTRIBUTE(always_inline)
+    constexpr __rcvr(__rcvr const & __other) noexcept
+      : __state_(__other.__state_)
+    {}
+
     template <class... _Args>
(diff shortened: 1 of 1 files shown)


include/stdexec/__detail/__basic_sender.hpp
```diff
@@ -236,6 +236,16 @@ namespace STDEXEC
...
 
+    STDEXEC_ATTRIBUTE(always_inline)
+    constexpr explicit __rcvr(_State& __state) noexcept
+      : __state_(__state)
+    {}
+
+    STDEXEC_ATTRIBUTE(always_inline)
+    constexpr __rcvr(__rcvr const & __other) noexcept
+      : __state_(__other.__state_)
+    {}
```
ericniebler, line 249: maybe leave a comment about why these ctors exist. otherwise, i'm likely to accidentally delete them.

ispeters: I was wondering if you'd want to exclude them from non-Apple-Clang builds with the preprocessor.

ericniebler, line 249:
```suggestion

#if STDEXEC_APPLE_CLANG()
    // These constructors are a work-around for bad codegen with apple-clang
    STDEXEC_ATTRIBUTE(always_inline)
    constexpr explicit __rcvr(_State& __state) noexcept
      : __state_(__state)
    {}

    STDEXEC_ATTRIBUTE(always_inline)
    constexpr __rcvr(__rcvr const & __other) noexcept
      : __state_(__other.__state_)
    {}
#endif  // STDEXEC_APPLE_CLANG()
```

PR #2103, opened 2026-05-28
make `affine` compatible with [exec.affine]

Cra3z:
make `affine` compatible with [[exec.affine]](https://eel.is/c++draft/exec.affine#5)

### DIFF
--- a/include/stdexec/__detail/__affine.hpp
+++ b/include/stdexec/__detail/__affine.hpp
@@ -17,7 +17,7 @@
 
 #include "__basic_sender.hpp"
 #include "__completion_behavior.hpp"
-#include "__continues_on.hpp"
+#include "__finally.hpp"
 #include "__schedulers.hpp"
 #include "__senders.hpp"
 #include "__unstoppable.hpp"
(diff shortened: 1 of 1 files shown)


include/stdexec/__detail/__affine.hpp
```diff
@@ -84,6 +114,11 @@ namespace STDEXEC
         // we can just return the child sender. Otherwise, we need to wrap it.
         return STDEXEC::__forward_like<_Sender>(__child);
       }
+      else if constexpr (__affine::__has_affine_member<__cv_child_t>)
+      {
+        // If the child has member function `affine`, We use it.
+        return STDEXEC::__forward_like<_Sender>(__child).affine();
+      }
```
ericniebler, line 96: please revert all changes except this one. the use of `__finally_` is intentional; it is used here as a stronger version of `continues_on`, because it connects the schedule operation before starting the predecessor, so that once the predecessor completes, the schedule operation can be started without having to deal with exceptions thrown by `connect`.

i believe the standard should be changed to this formulation (although i haven't filed an issue for it yet).

ericniebler (changes requested):

ericniebler (approved):

PR #2104, opened 2026-05-28
Clarify parallel scheduler build

vasama:
* Configure `STDEXEC_PARALLEL_SCHEDULER_HEADER_ONLY` appropriately based on `STDEXEC_BUILD_PARALLEL_SCHEDULER`.

* If `STDEXEC_BUILD_PARALLEL_SCHEDULER` is off, define the `parallel_scheduler` target as an empty interface target.

  Create the `STDEXEC::parallel_scheduler` alias target unconditionally.

* Build `example.sudoku` regardless of configuration.
* Build tests using the user-defined stdexec namespace.
* Remove local header-only macro override in parallel scheduler test.

* Remove unneeded weak attribute from default parallel scheduler definition.

  The status quo with the weak attribute simply means that in the case of a missing definition, instead of producing an error, the linker ignores the symbol and creates a null reference. This ultimately results in a runtime segmentation fault when the parallel scheduler backend is queried.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -260,6 +260,10 @@ if (NOT STDEXEC_NAMESPACE STREQUAL "stdexec")
   target_compile_definitions(stdexec INTERFACE STDEXEC_NAMESPACE=${STDEXEC_NAMESPACE})
 endif()
 
+if (NOT STDEXEC_BUILD_PARALLEL_SCHEDULER)
+  target_compile_definitions(stdexec INTERFACE STDEXEC_PARALLEL_SCHEDULER_HEADER_ONLY)
+endif()
+
 option(STDEXEC_ENABLE_EXTRA_TYPE_CHECKING "Enable extra type checking that is costly at compile-time" OFF)
 if (STDEXEC_ENABLE_EXTRA_TYPE_CHECKING)
   target_compile_definitions(stdexec INTERFACE STDEXEC_ENABLE_EXTRA_TYPE_CHECKING)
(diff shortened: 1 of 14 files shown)


ericniebler (approved):

vasama:
I see there are many build failures, but they don't look directly related to my changes. Perhaps applying the header only configuration is what is breaking the CI builds? 🤔

vasama:
I pushed some fixes. The MSVC and GCC-13 issues I was able to test locally. For the MacOS issue, missing support for `__has_include` was the only thing I could think of. But it isn't necessary here anyway.

vasama:
Okay. A more thorough reading of the MacOS failure revealed the problem. It's a cyclic include dependency:
```
libdispatch_queue.hpp -> execution.hpp -> ... -> __parallel_scheduler_default_impl.hpp -> libdispatch_queue.hpp
```

This should be solved by including only the necessary implementation headers in `libdispatch_queue.hpp`, but it's not clear to me what that set is. I also don't have a MacOS device for testing, unfortunately...

vasama:
I added a fix for the Windows build issue and removed the libdispatch commit. A fix is still needed for the cyclic include dependency I described above.

ericniebler:
@vasama could you have a look at the CI failures?

vasama:
I hit a roadblock not having access to a MacOS device. I'll look into setting up a virtual machine to test things out. However, I'm unfortunately much too busy this summer and won't be able to get back to this until some time in august.

PR #2107, opened 2026-06-01
Implement p4223 attrs

ispeters:
[P4223R0](https://wg21.link/p4223r0) calls for `function` to support a type parameter of the form `attrs<…>` that specifies the required attributes of the type-erased sender. This PR started out as an attempt to implement this missing feature in `exec::function`, but, along the way, I discovered that's not as trivial as I expected when I added the requirement to the paper. Unlike `exec::any_sender_of`, `function` doesn't know the type of the sender it's erasing except in its constructor and, at that point, it knows _only_ the type—it doesn't have the sender as a value until `connect` gets called—so we can't check that the erased sender's attributes meet the specified requirements.

The simplest resolution to this conflict is obviously to just not support sender attributes in `function` but, after consulting with @ericniebler on the subject, he pointed out that `get_completion_domain` is a function of a sender's attributes, and it's a critical part of the algorithm customization logic. So, this PR is an implementation of the compromise: you can specify the required completion domains of the erased sender because we need only the sender's type to validate the constraint is met, and we can do that in the `function` constructor.

New in this PR:
 * `template <class...> exec::attrs`, which is like `template <class...> exec::queries` but for sender attribute queries instead of receiver environment queries;
 * some more specializations of `exec::__make_function` to support specifying the completion domain requirements of the erased sender;
 * new constraints on `function`'s constructor to validate that the erased sender meets the aforementioned new domain requirements; and
 * some tests to validate that everything works as intended.

The new requirement is that `__common_domain_t<ExpectedDomain, ActualDomain>` is the same as `ExpectedDomain` (i.e. the completion domain of the erased sender must be, or inherit from, the completion domain specified in the `function`'s type).

TODO:
- [ ] negative tests to confirm the constructor constraints prevent invalid constructions
- [ ] more specializations of `__make_function` so you can specify required sender attributes without having to fully specify all of the type parameters
- [ ] tests for the above
- [ ] probably more documentation
- [ ] maybe confirmation that the only required sender attributes are `get_completion_domain`
- [ ] an R1 of P4223 that updates the spec to reflect this new implementation experience

### DIFF
--- a/include/exec/function.hpp
+++ b/include/exec/function.hpp
@@ -17,6 +17,7 @@
 
 #include "../stdexec/__detail/__completion_signatures.hpp"
 #include "../stdexec/__detail/__concepts.hpp"
+#include "../stdexec/__detail/__domain.hpp"
 #include "../stdexec/__detail/__meta.hpp"
 #include "../stdexec/__detail/__read_env.hpp"
 #include "../stdexec/__detail/__receivers.hpp"
@@ -55,6 +56,11 @@
 // queries to pick the frame allocator from the environment without relying on TLS.
 namespace experimental::execution
 {
+  // for specifying required sender attributes in exec::function
+  template <_query::_query_signature... Sigs>
+  struct attrs
+  {};
+
   namespace __func
   {
     using namespace STDEXEC;
@@ -192,7 +198,7 @@ namespace experimental::execution
       }
     };
 
-    template <class _Sigs, class _Queries, class... _Args>
+    template <class _Sigs, class _Queries, class _Attrs, class... _Args>
     class __function;
 
     //! the main implementation of the type-erasing sender function<...>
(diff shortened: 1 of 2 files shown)


include/exec/function.hpp
```diff
@@ -106,13 +112,13 @@ namespace experimental::execution
...
       constexpr auto get_env() const noexcept  //
-        -> __join_env_t<__prop_t, env_of_t<_Receiver>>
+        -> __join_env_t<__prop_t const &, env_of_t<_Receiver>>
       {
         return __env::__join(*__env_, STDEXEC::get_env(*static_cast<_Receiver const *>(this)));
       }
 
      private:
-      __prop_t *__env_;
+      __prop_t const *__env_;
```
ispeters, line 121: @ericniebler, what do you think of this change? It resolves a compiler error where `__join` receives a `__prop_t&` but is expecting a `__prop_t` and can't find a conversion. The other thing that worked was to invoke `__join` like so:
```c++
return __env::__join(auto(*__env_), …);
```
which seemed like a bunch of unnecessary copying.

Issue #2098, opened 2026-05-27
SIGSEGV with `co_await (stdexec::get_scheduler() | ...);`


ReneOschmann:
# Summary
Using `co_await (ex::get_scheduler() | ex::let_value([](auto sch) { return ex::schedule(sch); }));` in a `task`  leads to a premature exit of `sync_wait` and a subsequent use-after-free of the coroutine.
This is a regression introduced with 482c2605918d3aadbd472960e207126a21809140 #2078

Tested with Clang and GCC on x86_64

# Workaround
```c++
            auto sch = co_await ex::get_scheduler();
            co_await (ex::just(sch) | ex::let_value([](auto sch) { return ex::schedule(sch); }));
```

# Reproducer
https://godbolt.org/z/WE9fs5WPv

```c++
#include <coroutine>
#include <stdexec/execution.hpp>

namespace ex = STDEXEC;

void foo()
{
    int order[2] = {0, 0};
    size_t idx = 0;
    ex::sync_wait(
        [&]() -> ex::task<void> {
            order[idx++] = 1;
            co_await (ex::get_scheduler() | ex::let_value([](auto sch) { return ex::schedule(sch); }));
            order[idx++] = 2;
        }());
}

void foo_workaround()
{
    int order[2] = {0, 0};
    size_t idx = 0;
    ex::sync_wait(
        [&]() -> ex::task<void> {
            order[idx++] = 1;
            auto sch = co_await ex::get_scheduler();
            co_await (ex::just(sch) | ex::let_value([](auto sch) { return ex::schedule(sch); }));
            order[idx++] = 2;
        }());
}

int main()
{
    foo(); // crashes with SIGSEGV probaly use after free
    foo_workaround();
}

/* my usage:
namespace Tools
{
    // Give control back to the scheduler.
    struct Yield : ex::sender_adaptor_closure<Yield>
    {
        auto operator()() const
        {
            return ex::get_scheduler() | ex::let_value([](auto sch) { return ex::schedule(sch); });
        }

        template<class Sndr>
        auto operator()(Sndr sndr) const;
    };

    inline constexpr Yield yield{};

    template<class Sndr>
    auto Yield::operator()(Sndr sndr) const
    {
        return std::move(sndr) | ex::let_value([](auto... value) {
                   return ex::when_all(ex::just(std::move(value)...), Tools::yield());
               });
    }
} 

void foo_intended_use()
{
    int order[4] = {0, 0, 0, 0};
[...]

PR #2112, opened 2026-06-09
fix(read_env): only advertise inline_completion for set_value_t

nebkat:
`read_env`'s `__attrs::query` was a member template, so it returned `inline_completion` for every completion tag - even `set_error_t` and `set_stopped_t`, which read_env never sends. That made `let_value` mistakenly classify chains like

    read_env(q) | let_value([](auto v) { return some_async_sender; })

as inline-completing. The check `__never_sends || behavior == inline_completion` passed for every tag because read_env claimed inline across the board.

`__as_awaitable` then picked the inline `__sender_awaiter`, which holds the operation state as a local in `await_suspend` and destroys it at the closing brace. For any inner sender that hadn't actually completed by then, this is a use-after-free.

Example: [Godbolt](https://godbolt.org/z/b448xor4e)

----

Drop the `_SetTag` template parameter and bind the query directly to `set_value_t` - the only tag read_env sends - matching what `__just::__attrs` already does. The other tags now fall through to `__unknown`, the inline check correctly fails, and the non-inline awaiter gets selected.

### DIFF
--- a/include/stdexec/__detail/__read_env.hpp
+++ b/include/stdexec/__detail/__read_env.hpp
@@ -80,9 +80,16 @@ namespace STDEXEC
     template <class _Query>
     struct __attrs
     {
-      template <class _SetTag>
       STDEXEC_ATTRIBUTE(nodiscard)
-      constexpr auto query(__get_completion_behavior_t<_SetTag>) const noexcept
+      constexpr auto query(__get_completion_behavior_t<set_value_t>) const noexcept
+      {
+        return __completion_behavior::__inline_completion;
+      }
+
+      template <class _Env>
+        requires(!__nothrow_callable<_Query, _Env>)
+      STDEXEC_ATTRIBUTE(nodiscard)
+      constexpr auto query(__get_completion_behavior_t<set_error_t>, _Env const &) const noexcept
       {
         return __completion_behavior::__inline_completion;
       }


Cra3z:
I think the more fundamental cause lies in a flaw of the current `let` algorithm: right now, the attributes of a `let` sender merely **forward** those of **the predecessor sender**. This means that as long as **the predecessor sender** is inline-completing, the `let` sender becomes inline-completing, without taking **the completion behavior** of **the successor sender** into account.
See https://github.com/NVIDIA/stdexec/issues/2098#issuecomment-4603915829.

nebkat:
Right, I did think that was unusual!

I presume this is still the correct completion behavior for a sender which only completes with `set_value`? `just()` uses the `_JustTag::__tag_t` to template `__attrs<_SetTag>` so it only defines the `set_value` completion behavior.

ericniebler:
> right now, the attributes of a let sender merely forward those of the predecessor sender.

right, pr #2077 is intended to fix this, but the pr is not finished yet, and i've been unavoidably detained. #2077 exposed a problem with the attributes of `continues_on` that needs to be addressed first.

include/stdexec/__detail/__read_env.hpp
```diff
@@ -80,9 +80,8 @@ namespace STDEXEC
     template <class _Query>
     struct __attrs
     {
-      template <class _SetTag>
       STDEXEC_ATTRIBUTE(nodiscard)
-      constexpr auto query(__get_completion_behavior_t<_SetTag>) const noexcept
+      constexpr auto query(__get_completion_behavior_t<set_value_t>) const noexcept
```
ericniebler, line 84: it is also possible for the `read_env` sender to complete inline with an error if the query expression throws.

include/stdexec/__detail/__read_env.hpp
```diff
@@ -80,9 +80,14 @@ namespace STDEXEC
...
+      {
+        return __completion_behavior::__inline_completion;
+      }
+
+      STDEXEC_ATTRIBUTE(nodiscard)
+      constexpr auto query(__get_completion_behavior_t<set_error_t>) const noexcept
```
ericniebler: this should be constrained on whether the query expression is potentially throwing.

ericniebler (dismissed):

PR #2116, opened 2026-06-16
Isp

dontepanlin:


dontepanlin:
Fixed gcc (Ubuntu 13.3.0-6ubuntu2~24.04) 13.3.0 build
Removed strange github.com file download on every configure

include/stdexec/__detail/__domain.hpp
```diff
@@ -312,7 +312,7 @@ namespace STDEXEC
     }
 
     template <class _Attrs, class... _Env>
-    static constexpr auto __get_domain() noexcept
```
ericniebler, line 315: why?

CMakeLists.txt
```diff
@@ -24,15 +24,8 @@ if(POLICY CMP0167)
 endif()
 
 ##############################################################################
-# - Download and initialize RAPIDS CMake helpers -----------------------------
-
-# Fetch rapids-cmake
-if(NOT EXISTS ${CMAKE_CURRENT_BINARY_DIR}/RAPIDS.cmake)
-  file(DOWNLOAD https://raw.githubusercontent.com/rapidsai/rapids-cmake/branch-24.02/RAPIDS.cmake
-       ${CMAKE_CURRENT_BINARY_DIR}/RAPIDS.cmake)
-endif()
```
ericniebler, line 33: what problem does this fix?

dontepanlin: blocked github

include/stdexec/__detail/__domain.hpp
dontepanlin: Only this thing works on gcc13.3 version on big project. 
Single compilation is succed, but something strange occures on auto type declaration if this file included in many places. I cant explain exactly nature of this bug.

CMakeLists.txt
BartolomeyKant: Confirm. I also has issues with downloading this file on some machines from time to time. I don't know why.

include/stdexec/__detail/__domain.hpp
ericniebler: the problem is that this change is wrong. `__get_domain()` must return by value.

PR #2119, opened 2026-06-18
Destroy parallel bulk predecessor values

fallintoplace:
## Summary

This fixes the parallel scheduler bulk receiver so predecessor values stored for bulk execution have a real lifetime and are destroyed on every terminal path. The old implementation placement-newed a tuple into raw storage and even noted that the tuple was never destroyed, which leaks resources for owning predecessor values.

The storage now uses `__manual_lifetime`, destroys the stored tuple before forwarding `set_value`, `set_error`, or `set_stopped`, and adds coverage for value, error, and stopped completion paths. The test uses an owning value that tracks live instances so the old behavior leaves the count non-zero.

While adding coverage for value-carrying predecessor senders, the same path also needed a few adjacent fixes that were previously not instantiated by tests:

- expose the value completion domain from `parallel_scheduler`
- match `bulk_item_receiver_proxy::execute(size_t, size_t)`
- pass `(count, receiver, scratch)` to backend `schedule_bulk_*`
- invoke the connect initializer as an rvalue

History check: the raw-storage tuple and non-destruction comment were introduced in `b67a1e58` when `parallel_scheduler` moved into the `STDEXEC` namespace, then persisted through later formatting/refactors.

## Testing

- `cmake --build build-parallel-scheduler --target test.stdexec -j8`
- `./build-parallel-scheduler/test/test.stdexec "bulk on parallel_scheduler destroys stored predecessor values*"`
- `./build-parallel-scheduler/test/test.stdexec "[scheduler][parallel_scheduler]"`
- `cmake --build build-parallel-scheduler --target test.parallel_scheduler_replacement -j8`
- `./build-parallel-scheduler/test/test.parallel_scheduler_replacement`
- `git diff --check upstream/main...HEAD`

Not run locally: `clang-format --dry-run --Werror include/stdexec/__detail/__parallel_scheduler.hpp test/stdexec/schedulers/test_parallel_scheduler.cpp`, because `clang-format` is not installed on this machine.


ericniebler (approved):

fallintoplace:
@ericniebler Should I resolve the merge conflict too and push it?

PR #2117, opened 2026-06-18
Fix schedule_all chunking for small ranges

fallintoplace:
## Summary

Fix `static_thread_pool::schedule_all` batching so small non-empty ranges always make progress when the pool has more workers than items. The batching code now asserts the existing non-zero worker invariant, clamps non-empty chunk sizes to at least one item, and resets the pending task count after each intermediate flush.

## Testing

- `cmake --build build --target test.exec -j 8`
- `./build/test/exec/test.exec "schedule_all on static_thread_pool handles ranges smaller than available parallelism"`
- `./build/test/exec/test.exec "[types][static_thread_pool]"`
- `./build/test/exec/test.exec`

PR #2120, opened 2026-06-19
Fix schedule_all chunking for small ranges

fallintoplace:
## Summary

Fix `exec::schedule_all` on `static_thread_pool` for small non-empty ranges where the range size is less than the pool thread count.

The previous chunk-size calculation could produce `0`, which meant the batching loop never advanced. Giving non-empty work a minimum chunk size of `1` avoids that hard hang. While testing that path, I also found that chunk flushes moved the staged task queue without resetting `tasks_size_`, so the next flush could try to drain more tasks than were actually queued. The patch resets the staged count after each flush.

## Testing

- `cmake --build build --target test.exec`
- `./build/test/exec/test.exec "schedule_all on static_thread_pool handles fewer items than threads"`
- `./build/test/exec/test.exec "[types][static_thread_pool]"`
- `ASAN_OPTIONS=abort_on_error=0:symbolize=1 ./build-asan/test/exec/test.exec "schedule_all on static_thread_pool handles fewer items than threads"`
- `./build/test/exec/test.exec`

### DIFF
--- a/include/exec/static_thread_pool.hpp
+++ b/include/exec/static_thread_pool.hpp
@@ -1749,7 +1749,8 @@ namespace experimental::execution
           std::size_t nthreads     = this->pool_.available_parallelism();
           bwos_params params       = this->pool_.params();
           std::size_t local_size   = params.blockSize * params.numBlocks;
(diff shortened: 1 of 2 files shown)


ericniebler:
how is this different than #2117

PR #2151, opened 2026-07-28
Extend the stdexec module's coverage

ispeters:
This diff extends the `stdexec` module to cover more of what's defined by `#import <stdexec/execution.hpp>`. With this change in place, the `test.stdexec` binary has this output in modules mode:
```
All tests passed (2134 assertions in 322 test cases)
```
compared to this output in traditional mode:
```
All tests passed (3460 assertions in 598 test cases)
```
so we're now covering almost two thirds of the assertions in just more than half of the test cases.

The majority of the work was:
 * normalizing the includes section of headers in `include/stdexec` to follow the module-sensitive pattern
 * deciding which symbols in `STDEXEC` to export from the module
 * normalizing the includes section of test files in `test/stdexec` to follow the module-sensitive pattern


include/stdexec/__detail/__config.hpp
```diff
@@ -925,7 +925,7 @@ namespace STDEXEC
 #endif
 
 ////////////////////////////////////////////////////////////////////////////////////////////////////
-#if !defined(STDEXEC_DEMANGLE_SENDER_NAMES) && STDEXEC_MSVC()
+#if !defined(STDEXEC_DEMANGLE_SENDER_NAMES) && (STDEXEC_MSVC() || STDEXEC_USE_MODULES())
```
ispeters, line 934: This might be a notable change. I don't really understand the consequences, except that it works around a Clang bug.

PR #2153, opened 2026-07-31
Reference-Aware Algorithms

RobertLeahy:
Implements [Stop the Decay P4288R0](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2026/p4288r0.pdf).

Also contains the following fixes:

- `continues_on` now implements the standard behavior of scheduling when storing the result datums of the predecessor throws
- `finally` now runs the successor even if storing the result datums throws

Does not fully implement P4288R0's design for `storage_for_completion_signature` because `std::tuple` is not device compatible.


test/exec/test_when_any.cpp
```diff
@@ -215,9 +226,13 @@ namespace
...
 
-  TEST_CASE("when_any - with duplicate completions", "[adaptors][when_any]")
+  TEST_CASE("when_any rejects ambiguous normalized completions", "[adaptors][when_any]")
   {
-    REQUIRE_THROWS(STDEXEC::sync_wait(exec::when_any(dup_sender{})));
+    using sender_t = decltype(exec::when_any(dup_sender{}));
+    using completions_t =
+      decltype(sender_t::template get_completion_signatures<sender_t, ex::env<>>());
+
+    static_assert(ex::__merror<completions_t>);
```
ericniebler, line 235: once i fix the constexpr exception support in `get_completion_signatures`, this test will fail. i think you could change this to `static_assert(!ex::sender_in<sender_t, ex::env<>>)`.

include/stdexec/__detail/__tuple.hpp
```diff
@@ -232,7 +233,10 @@ namespace STDEXEC
     template <class... _Ts>
     STDEXEC_HOST_DEVICE_DEDUCTION_GUIDE __tuple(_Ts...) -> __tuple<_Ts...>;
 
-#  define STDEXEC_TUPLE_GET(_Idx) , static_cast<_Tuple&&>(__tupl).__val##_Idx
+#  define STDEXEC_TUPLE_GET(_Idx)                                                                  \
+    , static_cast<                                                                                 \
+        __copy_cvref_t<_Tuple&&, decltype(static_cast<_Tuple&&>(__tupl).__val##_Idx)>>(            \
+        static_cast<_Tuple&&>(__tupl).__val##_Idx)
```
ericniebler, line 239: why is this change necessary? shouldn't `static_cast<_Tuple&&>(__tupl).__val##_Idx` already have the correct value category if `__val##_Idx` has reference type?

include/stdexec/__detail/__spawn_future.hpp
```diff
@@ -96,24 +88,41 @@ namespace STDEXEC
...
-                           __decayed_signature_t<_Sigs>...>;
+      using __stored_signatures_t =
+        __concat_completion_signatures_t<completion_signatures<set_stopped_t()>,
+                                         completion_signatures<_Sigs...>>;
+      using __storage_t = storage_for_completion_signatures<__stored_signatures_t>;
+      using type =
+        __concat_completion_signatures_t<typename __storage_t::completion_signatures,
+                                         completion_signatures<set_error_t(std::exception_ptr)>>;
+
+      static consteval auto get_completion_signatures()
```
ericniebler, line 116: should the return type of this function always be the same as the `type` type alias?

include/stdexec/__detail/__as_awaitable.hpp
```diff
@@ -111,11 +150,11 @@ namespace STDEXEC
                               && __completes_inline_for<set_error_t, _Sender, _Env...>
                               && __completes_inline_for<set_stopped_t, _Sender, _Env...>;
 
-    template <class _Value, bool _Inline>
+    template <class _Sender, class _Promise, class _Value, bool _Inline>
```
ericniebler, line 153: i try to keep sender types out of the types of receivers (and awaitables) because it explodes the types of operation states.

include/stdexec/__detail/__as_awaitable.hpp
```diff
@@ -91,14 +98,46 @@ namespace STDEXEC
...
 
     template <class _Value>
-    using __expected_t = __variant<__value_or_void_t<_Value>, std::exception_ptr>;
+    struct __await_result
+    {
+      template <class _Arg>
+      constexpr auto operator()(_Arg&& __arg) const -> _Value
+      {
+        return static_cast<_Arg&&>(__arg);
+      }
```
ericniebler, line 123: i think if you change this to:
```suggestion
      template <class... _Args>
      constexpr auto operator()(_Args&&... __args) const -> _Value
      {
        return _Value(static_cast<_Args&&>(__args)...);
      }
```
you no longer need the specialization for `std::tuple`.

also, should this be be conditionally `noexcept`?

ericniebler (changes requested):
partial review, more coming...

> Also contains the following fixes:
> 
> * `continues_on` now implements the standard behavior of scheduling when storing the result datums of the predecessor throws
> * `finally` now runs the successor even if storing the result datums throws

would be great if these fixes were in a separate PR so i could merge them early.

include/stdexec/__detail/__tuple.hpp
RobertLeahy: Given:

```c++
struct foo { int&& bar; };
```

[Both `decltype((std::declval<foo>().bar))` and `decltype((std::declval<foo&>().bar))` are `int&`.](https://godbolt.org/z/bMWWaxG5r)

[`std::tuple<int&&>` comparison here.](https://godbolt.org/z/n874e47Pc)

RobertLeahy:
> would be great if these fixes were in a separate PR so i could merge them early.

I can make separate PRs for both of them but a separate fix wasn't/isn't a natural consequence of this work. The fixes just happened to fall out by accident from converting both of them over to `storage_for_completion_signatures`.

include/stdexec/__detail/__tuple.hpp
ericniebler: TIL. thanks.

Issue #2023, opened 2026-04-13
flaky test for ASIO completion tokens


ericniebler:
as noted [here](https://github.com/NVIDIA/stdexec/blob/849f9899c2f4684c4c08fafb9f02fdf34d86fe17/test/exec/asio/test_completion_token.cpp#L354-L366), one of the completion token tests is flaky and needs to be investigated.

https://github.com/NVIDIA/stdexec/blob/849f9899c2f4684c4c08fafb9f02fdf34d86fe17/test/exec/asio/test_completion_token.cpp#L354-L366

PR #2155, opened 2026-08-05
asioexec::completion_token: Fix Miscompiling Test

RobertLeahy:
GCC 14 miscompiles the affected test in release without -fno-ipa-modref. GCC 13 and 15 don't seem to have this issue, hence the scope of the workaround.

Reverted 2681e4f09707929b076507edf3f5dba038ec6ce4 (which didn't address the problem fully) and removed the [!mayfail] tag added by 849f9899c2f4684c4c08fafb9f02fdf34d86fe17.

Addresses #2023.

### DIFF
--- a/test/exec/asio/test_completion_token.cpp
+++ b/test/exec/asio/test_completion_token.cpp
@@ -38,6 +38,11 @@
 using namespace STDEXEC;
 using namespace exec::asio;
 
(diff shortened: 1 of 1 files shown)


ericniebler (approved):
very nice. i wonder if there is a source-level pragma that does the same job as the `-fno-ipa-modref` command line flag. not everybody builds with cmake.

PR #2157, opened 2026-08-07
do not use sccache distributed builds when building with C++20 modules

ericniebler:
a bug in sccache is causing modularized builds to fail in CI. turn off distributed builds for those CI jobs.


ispeters (approved):
LGTM

PR #2159, opened 2026-08-08
fix(__any_allocator): add friend declaration so the cross-specialization converting constructor compiles

### LINKED ISSUE #2158: task_scheduler fails to compile: __any_allocator cross-specialization converting constructor accesses a private member
## Summary

`stdexec::task_scheduler` fails to compile when the type-erased backend takes the heap-allocation fallback path: `__any_allocator<std::byte>` has a converting constructor that reads the private member of *another specialization* of the same class template:

```cpp
template <class _Uy>
constexpr __any_allocator(__any_allocator<_Uy> __other) noexcept
  : __alloc_(std::move(__other.__alloc_))   // cross-specialization private access
{ }
```

That access is rejected by **all major compilers** — not just MSVC. Verified locally:

- MSVC cl 19.51.36252: `C2248: cannot access private member`
- Clang (clang-cl 22.1.8): `error: '__alloc_' is a private member of 'stdexec::__any_allocator<std::byte>'`
- GCC 13.3: `error: ... is private within this context`

## Why upstream never noticed

The converting constructor is a member function template, so it is only instantiated when used. In the upstream test matrix every scheduler's operation state fits in `task_scheduler`'s inline storage, so the backend never falls back to the heap and the constructor is never instantiated. It *is* instantiated when the operation state is large — e.g. `stdexec::task` on an asio-based scheduler (`asio::post` + `use_sender`, whose opstate exceeds the inline storage) — which breaks the build on every compiler, not just MSVC.

## Minimal repro — language level

```cpp
#include <utility>

template <class T>
struct A {
  template <class U>
  A(A<U> other) noexcept : x_(std::move(other.x_)) {}  // reads other.x_ (private)
  A() = default;

 private:
  T x_;
};

int main() {
  A<int> a;
  A<long> b(a);   // rejected by MSVC, Clang and GCC
}
```

## stdexec repro

```cpp
#include <exec/asio/asio_config.hpp>
#include <exec/asio/use_sender.hpp>
#include <asio/io_context.hpp>

struct Sched {
  using scheduler_concept = stdexec::scheduler_tag;
  asio::io_context* ioc_ = nullptr;

  stdexec::sender auto schedule() const noexcept {
    return exec::asio::asio_impl::post(*ioc_, exec::asio::use_sender)
         | stdexec::upon_error([](std::exception_ptr) noexcept {})
         | stdexec::upon_stopped([]() noexcept {});
  }
  bool operator==(const Sched&) const noexcept = default;
};

int main() {
  asio::io_context ioc;
  Sched s{&ioc};
  stdexec::task_scheduler ts(s, std::allocator<Sched>{});
  (void)ts;
}
```

```
__any_allocator.hpp(83): error C2248: "stdexec::__any_allocator<std::byte>::__alloc_":
cannot access private member (declared in "stdexec::__any_allocator<std::byte>")
```

## Notes

- Wrapping `inline_scheduler`, `run_loop::scheduler`, or `exec::asio::asio_thread_pool`'s scheduler compiles fine (small operation states, inline storage, converting constructor never instantiated). Only schedulers whose schedule operation state exceeds the inline storage (e.g. asio `post` + `use_sender`) hit it.
- Since the constructor is instantiated only on that path, a plain minimal project may or may not reproduce depending on the include set and exact opstate size (`STDEXEC_TASK_SCHEDULE_OPSTATE_SIZE`).

## Suggested fix

Add an explicit friend declaration to `__any_allocator` (legal on every compiler):

```cpp
// in struct __any_allocator, next to the default constructor:
template <class>
friend struct __any_allocator;
```

See PR #2159 for the implementation and a regression test.

deretame:
## Summary

`stdexec::task_scheduler` fails to compile when its type-erased backend takes the heap-allocation fallback path: `__any_allocator<std::byte>` has a converting constructor that reads the private member of *another specialization* of the same class template. That access is **rejected by all major compilers** (MSVC `C2248`, Clang and GCC report it as accessing a private member) — verified with MSVC cl 19.51.36252, clang-cl 22.1.8 and GCC 13.3.

The constructor is only instantiated on the heap-allocation fallback path (operation state larger than `task_scheduler`'s inline storage), which the upstream test matrix never exercises — e.g. `stdexec::task` on an asio-based scheduler (`asio::post` + `use_sender`). So the build breaks on every compiler, not just MSVC. See #2158 for the full analysis and repro.

## Fix

Add an explicit friend declaration to `__any_allocator` (makes the cross-specialization access legal on every compiler):

```cpp
template <class>
friend struct __any_allocator;
```

## Test

Add `test/stdexec/detail/test_any_allocator.cpp`, which instantiates the cross-specialization converting constructor directly. The test fails on **every** compiler without the friend declaration (verified: MSVC C2248, Clang private-member error, GCC private-within-this-context), so it guards the fix on all platforms.

Verified locally on MSVC cl 19.51.36252:

| | without friend | with friend |
|---|---|---|
| `__any_allocator<int>` → `__any_allocator<std::byte>` conversion | ❌ C2248 | ✅ compiles |
| `test.stdexec` full build + run (`*any_allocator*`) | — | ✅ 1 test case passed |

clang-cl 22.1.8 also rejects the conversion without the friend declaration and compiles with it.

Fixes #2158

### DIFF
--- a/include/stdexec/__detail/__any_allocator.hpp
+++ b/include/stdexec/__detail/__any_allocator.hpp
@@ -67,10 +67,9 @@ namespace STDEXEC
 
     __any_allocator() = default;
 
-    // MSVC does not implement [class.access]: members of a class template
-    // cannot access private members of other specializations of the same
-    // template. The converting constructor below relies on that rule, so
-    // declare the friendship explicitly.
+    // The converting constructor below accesses the private member of another
+    // specialization of this template, which is rejected by MSVC, Clang and
+    // GCC. Declare the friendship explicitly.
     template <class>
     friend struct __any_allocator;
 
--- a/include/stdexec/__detail/__any_allocator.hpp
+++ b/include/stdexec/__detail/__any_allocator.hpp
@@ -67,6 +67,13 @@ namespace STDEXEC
 
     __any_allocator() = default;
 
+    // MSVC does not implement [class.access]: members of a class template
+    // cannot access private members of other specializations of the same
+    // template. The converting constructor below relies on that rule, so
+    // declare the friendship explicitly.
+    template <class>
+    friend struct __any_allocator;
+
     template <__not_same_as<__any_allocator> _Alloc>
       requires __is_not_instance_of<_Alloc, __any_allocator> && __simple_allocator<_Alloc>
     __any_allocator(_Alloc __alloc) noexcept
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -75,6 +75,7 @@ set(stdexec_test_sources
     stdexec/algos/consumers/test_sync_wait.cpp
     stdexec/algos/consumers/test_spawn.cpp
     stdexec/detail/test_any.cpp
+    stdexec/detail/test_any_allocator.cpp
     stdexec/detail/test_common_domain.cpp
     stdexec/detail/test_completion_signatures.cpp
     stdexec/detail/test_demangle.cpp
--- a/test/stdexec/detail/test_any_allocator.cpp
+++ b/test/stdexec/detail/test_any_allocator.cpp
@@ -22,12 +22,15 @@
 #include <utility>
 
 // The converting constructor __any_allocator(_Uy) reads the private member of
-// another specialization of the same class template. This is legal since C++17
-// ([class.access]) but MSVC rejects it with C2248; the friend declaration on
-// __any_allocator makes the conversion compile everywhere. The conversion is
-// instantiated whenever task_scheduler's type-erased backend needs to copy an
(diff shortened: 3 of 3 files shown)


deretame:
Hi @ericniebler, thank you very much for the earlier `/ok to test`!

I pushed a follow-up commit (`9f40f646`) addressing the clang-format check that failed: the new test file had two alignment violations (`AlignConsecutiveDeclarations` / `AlignTrailingComments`). I verified locally with clang-format 21.1.0 (the same version CI uses) that the whole tree now passes `--dry-run --Werror`.

Could you please run `/ok to test 9f40f646` again when you get a chance? Much appreciated!

PR #2162, opened 2026-08-08
Forward rooted environment in start_detached

fallintoplace:
## Summary

- forward the rooted environment into `apply_sender`
- add coverage for a move-only environment

## Why

The two-argument overload stores the rooted environment in `__env2` and then passes that named variable as an lvalue. The default implementation stores the decayed environment by value, so a move-only environment cannot be used. Forwarding `__env2` keeps the value category expected by the constraints.

## Testing

- `cmake --build build --target test.exec -j 8`
- `./build/test/exec/test.exec '[start_detached]'`


ericniebler (approved):

ericniebler:
attn @fallintoplace, this needs formatting

PR #2164, opened 2026-08-08
Preserve completion scheduler in timed fallbacks

fallintoplace:
## Summary

The timed scheduler fallbacks currently use `let_value(just(), ...)` to keep `now()` lazy. `let_value` does not expose a completion scheduler, so both fallback senders lose `get_completion_scheduler<set_value_t>` even when the delegated timed sender advertises it.

This adds a small internal wrapper that forwards the fallback sender’s existing completion behavior and signatures and adds the set-value completion scheduler. Tests cover both fallback directions.

## Testing

- `build/test/exec/test.exec`
- `build/test/exec/test.exec "[timed_scheduler]"`
- `build/test/exec/test.exec "[timed_thread_scheduler]"`

### DIFF
--- a/include/exec/timed_scheduler.hpp
+++ b/include/exec/timed_scheduler.hpp
@@ -25,7 +25,20 @@ namespace STDEXEC::__timed_scheduler_fallback
   struct __tag
   {};
 
-  template <class _Scheduler, class _Sender>
+  template <class _NativeSender, class _Scheduler>
+  concept __completion_scheduler_matches =
+    __callable<get_completion_scheduler_t<set_value_t>, env_of_t<_NativeSender>>
+    && __decays_to<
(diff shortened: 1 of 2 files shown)


include/exec/timed_scheduler.hpp
```diff
@@ -20,6 +20,71 @@
 
 #include <chrono>
 
+namespace STDEXEC::__timed_scheduler_fallback
```
ericniebler: should this be in namespace `experimental::execution` instead of `STDEXEC`?

ericniebler:
i'm ambivalent about merging this since it is a work-around for a problem in the `let_` senders that #2077 aims to fix.

fallintoplace:
Looking at #2077, I don’t think the current predecessor-forwarding case handles this composition: just() has no completion scheduler, and the timed sender that does have one is only produced lazily by the let_value function.

ericniebler:
Good point

PR #2169, opened 2026-08-09
Fix inplace_stop_source request_stop return value

fallintoplace:
This fixes the boolean result returned by `inplace_stop_source::request_stop()`.

The implementation currently returns `true` when a stop request was already made and `false` after this call successfully requests stop. That is reversed. The first call should return `true`, and later calls should return `false`.

The regression test checks both calls and confirms that the stop request is visible.

Tests:
- `cmake --build build --target test.exec -j 4`
- `build/test/exec/test.exec '[stop_token]'`
- `build/test/exec/test.exec`


ericniebler (approved):
nice catch!

ericniebler:
thanks!

PR #2175, opened 2026-08-10
Fix merge_each error propagation

fallintoplace:
## Summary

- preserve an existing error when nested work also signals a break
- complete normally after the delayed error item is consumed so sequence consumers can publish it
- cover errors from nested value senders, nested sequences, and the outer sequence

## Testing

- full `test.exec` run with `-fno-exceptions`


ericniebler (approved):

ericniebler:
@fallintoplace please have a look at these failures

PR #2176, opened 2026-08-10
Fix thread_pool_base bulk capture failure

fallintoplace:
## Summary

- stop bulk execution after predecessor value storage throws
- preserve the original capture exception as the only completion
- add a direct `thread_pool_base` regression test

The catch block already completed the receiver with `set_error`, but then continued into bulk dispatch with a valueless variant. The equivalent `static_thread_pool` path already returns at this point.

## Testing

- `ctest --test-dir build --output-on-failure -j8`
- full `test.exec` run with exceptions disabled


ericniebler (approved):
nice find!

PR #2179, opened 2026-08-10
Fix constrained worker selection

fallintoplace:
## Summary

- select the round-robin target from the eligible worker count
- preserve the existing fallback when no pool workers match the mask
- add a regression test with a synthetic two-node NUMA policy

The constrained selection path reduced its target modulo the full pool size before walking only the eligible workers. When the eligible set was smaller than the pool, indices beyond that set fell through the traversal and were returned as raw worker indices, allowing work to be queued on an excluded node.

The regression test uses four workers split across two synthetic nodes. Before this change, two of four consecutive submissions ran on the excluded node.

## Testing

- full test.exec suite: 3,302 assertions in 327 test cases
- full test.exec suite with -fno-exceptions: 3,244 assertions in 309 test cases
- regression test repeated 100 times

### DIFF
--- a/include/exec/static_thread_pool.hpp
+++ b/include/exec/static_thread_pool.hpp
@@ -853,8 +853,8 @@ namespace experimental::execution
     {
       thread_local std::uint64_t start_index{std::uint64_t(std::random_device{}())};
       start_index += 1;
-      std::size_t target_index = start_index % thread_count_;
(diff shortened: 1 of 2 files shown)


ericniebler (approved):
LGTM

fallintoplace:
Some clang format errors appear but I can only fix tomorrow.

PR #2185, opened 2026-08-10
Preserve nvexec repeat_n child errors

fallintoplace:
## Summary

- derive `repeat_n` completions from its child sender
- preserve and decay child error types while keeping the existing CUDA and exception paths
- add coverage with a custom child error

## Testing

- `git diff --check`

### DIFF
--- a/include/nvexec/stream/repeat_n.cuh
+++ b/include/nvexec/stream/repeat_n.cuh
@@ -155,7 +155,7 @@ namespace nv::execution::_strm
       using sender_concept = STDEXEC::sender_tag;
 
       template <class...>
-      using _set_value_t = STDEXEC::completion_signatures<STDEXEC::set_value_t()>;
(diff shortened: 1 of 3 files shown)


include/nvexec/stream/repeat_n.cuh
```diff
@@ -154,13 +154,26 @@ namespace nv::execution::_strm
     {
       using sender_concept = STDEXEC::sender_tag;
 
-      template <class>
-      static consteval auto get_completion_signatures()
+      template <class...>
+      using _set_value_t = STDEXEC::completion_signatures<STDEXEC::set_value_t()>;
```
ericniebler:
```suggestion
      using _set_value_t = STDEXEC::completion_signatures<>;
```

you're already adding the `set_value_t()` signature below.

ericniebler (changes requested):

PR #2187, opened 2026-08-11
Fix repeat receiver lifetime after cleanup

fallintoplace:
## Summary

- save the parent operation-state pointer before cleaning up the child operation
- use the saved pointer for downstream completion in `repeat_n` and `repeat_until`
- add regression coverage for value, error, and stopped completions

## Why

Cleanup resets the child operation, which owns the repeat receiver that is currently completing. Reading `__state_` from that receiver after cleanup accesses an object whose lifetime has ended.

## Testing

- full `test.exec` suite: 3,326 assertions across 328 test cases
- focused `[repeat_n]` and `[repeat_until]` test groups
- focused test groups with `-fno-exceptions`

### DIFF
--- a/include/exec/repeat_n.hpp
+++ b/include/exec/repeat_n.hpp
@@ -76,25 +76,27 @@ namespace experimental::execution
       template <class _Error>
       constexpr void set_error(_Error &&__err) noexcept
       {
+        auto *__state = __state_;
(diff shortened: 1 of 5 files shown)


ericniebler (approved):
nice!

PR #2188, opened 2026-08-11
Remove duplicate algorithm include

fallintoplace:
## Summary

- remove a duplicate `<algorithm>` include from the nvexec launch test

## Testing

- `git diff --check`


test/nvexec/launch.cpp
ericniebler: remove this one instead, pls.

ericniebler (changes requested):

ericniebler (approved):

PR #2189, opened 2026-08-11
Destroy nvexec ensure_started completion storage before freeing it

fallintoplace:
## Summary

- wait for the pending completion copy before destroying managed storage
- destroy the managed completion variant before freeing its allocation
- cover both consumed and detached senders with a non-trivial completion value

## Why

`ensure_started` placement-constructs its completion storage in managed memory. The shared state freed that memory without running the variant destructor, so the active completion value was never destroyed.

The direct stream path copies the completion asynchronously. Its recorded event now synchronizes that copy before host destruction. If recording the event fails after the copy is queued, destruction synchronizes the stream instead.

### DIFF
--- a/include/nvexec/stream/ensure_started.cuh
+++ b/include/nvexec/stream/ensure_started.cuh
@@ -209,6 +209,7 @@ namespace nv::execution::_strm
 
         if (data_)
         {
+          data_->~variant_t();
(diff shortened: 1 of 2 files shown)


ericniebler:
why is this a draft PR?

fallintoplace:
I want to check until I am sure that everything is working correctly and there is no hidden regression.

PR #2190, opened 2026-08-11
Destroy nvexec continues_on storage on CUDA errors

fallintoplace:
## Summary

- destroy host-constructed completion storage before propagating CUDA setup errors
- keep the existing deferred destruction on successful paths
- add a regression using pinned host memory to force `cudaMemPrefetchAsync` to fail

## Why

`continues_on` constructs non-trivial completion storage on the host before several CUDA calls. An early error returned without running the active completion value destructor, and the operation state only released the raw allocation.

Cleanup now happens before the downstream completion operation because that operation may destroy the operation state.


ericniebler:
why is this a draft PR?

PR #2198, opened 2026-08-15
Fix nvexec launch move-only callable forwarding

fallintoplace:
## What changed

- Forward the stored launch callable with the sender's value category.
- Add coverage for a trivially copyable callable that cannot be copied.

## Why

nvexec::launch accepts movable callables, but connect() passed the stored callable as an lvalue. Consuming a sender with a move-only callable therefore tried to copy it.

## Testing

- git diff --check
- Standalone C++ syntax check for the move-only callable
- CUDA coverage is included for GPU CI


ericniebler (approved):

fallintoplace:
@ericniebler Hi, I just pushed the clang format fix. [6d12001](https://github.com/NVIDIA/stdexec/pull/2198/commits/6d120015b0d79da99318dd74b1709d99de1c477d)

PR #2201, opened 2026-08-15
Stop libdispatch bulk after value capture failure

fallintoplace:
## Summary

- stop libdispatch bulk after predecessor value capture fails
- add regression coverage for a throwing predecessor value

The receiver already sends `set_error` when value capture fails. The old code continued into the completion path after that error.

Tests:
- `cmake --build build --target test.exec -j 8`
- `./build/test/exec/test.exec --reporter compact`


ericniebler (approved):

fallintoplace:
Hey Eric, thanks for approving. I found one more edge case in libdispatch bulk capture for set_value(T&). I’m going to push a small follow-up with the forwarding fix and a regression test. The existing tests pass with it.

PR #2202, opened 2026-08-15
Enforce header self containment

ispeters:
Claude wrote this diff for me to add CMake's auto-generated tests that stdexec's headers are self-contained to CI.

There are a handful of headers that are deliberately not self-contained so they're excluded from the auto-generated validation with `set_source_files_properties(<file> PROPERTIES SKIP_LINTING ON)` (documented in `CMakeLists.txt`). `__parallel_scheduler_default_impl_entry.hpp` is *almost* self-contained—it has a precondition that the includer define `STDEXEC_PARALLEL_SCHEDULER_INLINE` before including it but it is otherwise self-contained so there's a "hand-written" `.cpp` file that does nothing but define `STDEXEC_PARALLEL_SCHEDULER_INLINE` and include the header, which is added as a dependency to the auto-generated tests.

### DIFF
--- a/.github/workflows/ci.cpu.yml
+++ b/.github/workflows/ci.cpu.yml
@@ -114,6 +114,7 @@ jobs:
             -DCMAKE_BUILD_TYPE=${{ matrix.build }} \
             -DCMAKE_CXX_FLAGS="${{ matrix.cxxflags }}" \
             -DSTDEXEC_ENABLE_TBB:BOOL=${{ !contains(matrix.cxxflags, '-fsanitize') && !contains(matrix.name, 'modules') }} \
+            -DSTDEXEC_ENABLE_TASKFLOW:BOOL=${{ matrix.name == 'CPU (clang 22, Debug)' }} \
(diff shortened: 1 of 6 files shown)


.github/workflows/ci.cpu.yml
```diff
@@ -128,6 +128,14 @@ jobs:
           # to result in fewer Clang ICEs.
           cmake --build build -v -j ${{ contains(matrix.name, 'modules') && 1 || 512 }};
 
+          # Verify that public headers are self-contained. Header
+          # verification is independent of the rest of the build matrix, so
+          # it only needs to run once; pick a single plain (non-modules,
+          # non-sanitizer) Debug build to run it in.
```
ispeters: This is perhaps a lie. Some of the excluded files are excluded for being Windows-only, or similar. This diff is certainly adding value, but maybe it makes sense to attach this validation to more elements of the matrix to ensure at least one build includes each of the conditionally-excluded files.

PR #2203, opened 2026-08-15
Fix schedule_all move-only range support

fallintoplace:
## What changed

- forward the range when building the schedule_all operation state
- add a regression test with a move-only owning_view

## Why

The consuming subscribe() path moves the range, but opstate_base_with_receiver passed its by-value range parameter as an lvalue. That made valid move-only ranges fail to compile.

## Testing

- build/test/exec/test.exec 'schedule_all on static_thread_pool accepts move-only ranges'
- build/test/exec/test.exec

### DIFF
--- a/include/exec/static_thread_pool.hpp
+++ b/include/exec/static_thread_pool.hpp
@@ -1652,7 +1652,7 @@ namespace experimental::execution
       struct opstate_base_with_receiver : opstate_base<Range>
       {
         explicit opstate_base_with_receiver(Range range, _static_thread_pool& pool, Receiver rcvr)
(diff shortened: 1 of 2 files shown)


ericniebler (approved):

ericniebler:
@fallintoplace can you pls look into the regression tests failure on this PR and the others you've opened?

fallintoplace:
@ericniebler Let me have a look now.

ericniebler (approved):

ericniebler:
@fallintoplace this PR has test failures that need your attention.

ericniebler:
thanks!

PR #2208, opened 2026-08-16
Fix read_with_default connect noexcept

fallintoplace:
## What changed

- Make `read_with_default` member `connect()` derive its exception specification from constructing the returned operation state.
- Add a regression for a default value whose move throws during `connect()`.

## Why

The operation state already accounts for a potentially throwing default move, but `connect()` only considered moving the receiver.

## Testing

- `build/test/exec/test.exec` (3345 assertions passed)
- `ctest --test-dir build --output-on-failure -j 8` (948 passed)


ericniebler (approved):

ericniebler:
@fallintoplace please clang-format all commits _before_ you make a pull request.

fallintoplace:
@ericniebler roger that. I will follow next time.

PR #2212, opened 2026-08-16
Fix just_from submit exception specification

fallintoplace:
## What changed

- Make both `just_from::submit()` overloads conditionally `noexcept`.
- Add coverage for a callable that throws while being moved during `submit()`.

## Why

`submit()` calls `connect()`, but both overloads were unconditionally `noexcept`. A throwing callable move could therefore terminate instead of propagating the exception.

## Testing

- `build/test/exec/test.exec [just_from]`

All 953 configured tests passed.

### DIFF
--- a/include/exec/just_from.hpp
+++ b/include/exec/just_from.hpp
@@ -172,7 +172,7 @@ namespace experimental::execution
 
       template <class Rcvr>
       STDEXEC_ATTRIBUTE(host, device)
-      auto submit(Rcvr rcvr) && noexcept -> void
+      auto submit(Rcvr rcvr) && noexcept(STDEXEC::__nothrow_decay_copyable<Rcvr, Fn>) -> void
       {
         auto op = static_cast<_sndr_base&&>(*this).connect(static_cast<Rcvr&&>(rcvr));
         STDEXEC::start(op);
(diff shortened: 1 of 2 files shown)


include/exec/just_from.hpp
```diff
@@ -172,15 +172,16 @@ namespace experimental::execution
 
       template <class Rcvr>
       STDEXEC_ATTRIBUTE(host, device)
-      auto submit(Rcvr rcvr) && noexcept -> void
+      auto submit(Rcvr rcvr) && noexcept(STDEXEC::__nothrow_decay_copyable<Rcvr, Fn>) -> void
```
ericniebler: receivers are now required to be nothrow move constructible, so this check is unnecessary.
```suggestion
      auto submit(Rcvr rcvr) && noexcept(STDEXEC::__nothrow_decay_copyable<Fn>) -> void
```

include/exec/just_from.hpp
```diff
@@ -172,15 +172,16 @@ namespace experimental::execution
...
       {
         auto op = static_cast<_sndr_base&&>(*this).connect(static_cast<Rcvr&&>(rcvr));
         STDEXEC::start(op);
       }
 
       template <class Rcvr>
       STDEXEC_ATTRIBUTE(host, device)
-      auto submit(Rcvr rcvr) const & noexcept -> void
+      auto submit(Rcvr rcvr) const & noexcept(STDEXEC::__nothrow_decay_copyable<Rcvr, Fn const &>)
```
ericniebler:
```suggestion
      auto submit(Rcvr rcvr) const & noexcept(STDEXEC::__nothrow_decay_copyable<Fn const &>)
```

PR #2213, opened 2026-08-16
Make exec::without conditionally noexcept

fallintoplace:
## What changed

- Make `exec::without` derive its exception specification from the selected return object construction.
- Keep lvalue environments nothrow when `without` only stores a reference.
- Add regressions for throwing rvalue environment moves in both queryable and missing-query paths.

## Why

`exec::without` was unconditionally `noexcept`, even though it can move an rvalue environment into its return object. A throwing move could terminate instead of propagating the exception.

## Testing

- `build/test/exec/test.exec [env]`

All 953 configured tests passed.

### DIFF
--- a/include/exec/env.hpp
+++ b/include/exec/env.hpp
@@ -61,16 +61,20 @@ namespace experimental::execution
     struct __without_t
     {
       template <class _Env, class _Query>
-      constexpr auto operator()(_Env&& __env, _Query) const noexcept
+        requires STDEXEC::__queryable_with<_Env, _Query>
+      constexpr auto operator()(_Env&& __env, _Query) const
+        noexcept(STDEXEC::__nothrow_constructible_from<__without<_Env, _Query>, _Env>)
+          -> __without<_Env, _Query>
       {
-        if constexpr (STDEXEC::__queryable_with<_Env, _Query>)
-        {
-          return __without<_Env, _Query>{static_cast<_Env&&>(__env)};
-        }
-        else
-        {
-          return static_cast<_Env&&>(__env);
-        }
+        return __without<_Env, _Query>{static_cast<_Env&&>(__env)};
+      }
+
+      template <class _Env, class _Query>
+        requires(!STDEXEC::__queryable_with<_Env, _Query>)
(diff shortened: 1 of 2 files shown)


include/exec/env.hpp
```diff
@@ -61,7 +61,10 @@ namespace experimental::execution
     struct __without_t
     {
       template <class _Env, class _Query>
-      constexpr auto operator()(_Env&& __env, _Query) const noexcept
+      constexpr auto operator()(_Env&& __env, _Query) const noexcept(
+        STDEXEC::__nothrow_constructible_from<
+          STDEXEC::__if_c<STDEXEC::__queryable_with<_Env, _Query>, __without<_Env, _Query>, _Env>,
```
ericniebler: this isn't quite right because the function currently always returns by value, and `_Env` could be a reference type. Arguably, the function should either be returning a `__without<_Env, _Query>` or an `_Env`. it would probably be easiest to split this into two overloads instead of using an `if constexpr`.

ericniebler (changes requested):

include/exec/env.hpp
```diff
@@ -61,16 +61,20 @@ namespace experimental::execution
...
-        }
-        else
-        {
-          return static_cast<_Env&&>(__env);
-        }
+        return __without<_Env, _Query>{static_cast<_Env&&>(__env)};
+      }
+
+      template <class _Env, class _Query>
+        requires(!STDEXEC::__queryable_with<_Env, _Query>)
```
ericniebler:
```suggestion
```

include/exec/env.hpp
```diff
@@ -61,16 +61,20 @@ namespace experimental::execution
...
+      constexpr auto operator()(_Env&& __env, _Query) const
+        noexcept(STDEXEC::__nothrow_constructible_from<_Env, _Env>) -> _Env
```
ericniebler:
```suggestion
        noexcept(STDEXEC::__nothrow_move_constructible<_Env>) -> _Env
```

ericniebler (approved):

PR #2219, opened 2026-08-21
Fix schedule_from task cleanup before start

fallintoplace:
## Problem

`schedule_from` releases the owning pointer for its continuation task during `connect()`, but the operation state only keeps a raw pointer after that. If the operation is destroyed without calling `start()`, the task never reaches the queue and its device and pinned allocations are not freed.

## Fix

- Keep the continuation task owned by the operation state until the inner operation is started.
- Release ownership immediately before calling `start(inner_op_)`.
- Keep cleanup correct when setup reports a CUDA error or connecting the inner sender unwinds.

## Test

- Added a regression test for destroying an unstarted operation.
- Added a setup-error regression that fails the environment allocation after storage and task allocation succeed, starts the operation, checks `cudaErrorMemoryAllocation`, and verifies balanced cleanup.
- `clang-format-21 --dry-run --Werror`
- CMake configure and build with CUDA disabled

### DIFF
--- a/include/nvexec/stream/schedule_from.cuh
+++ b/include/nvexec/stream/schedule_from.cuh
@@ -19,9 +19,8 @@
 #pragma once
 
 #include "../../stdexec/execution.hpp"
+#include <memory>
 #include <utility>
-
-#include "../detail/cuda_atomic.cuh"  // IWYU pragma: keep
 
 #include "common.cuh"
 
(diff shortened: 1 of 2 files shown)


include/nvexec/stream/schedule_from.cuh
```diff
@@ -97,23 +109,21 @@ namespace nv::execution
 
         void start() & noexcept
         {
-          started_.test_and_set(::cuda::std::memory_order::relaxed);
-
```
ericniebler, line 101: @gevtushenko does this `atomic_flag` `test_and_set` have a purpose? the `started_` flag is not referenced anywhere else.

fallintoplace: I checked the history: started_ was introduced in 6302fbef as part of the P3826 rewrite, but it isn’t used elsewhere. It isn’t needed here, so I removed it and the unused CUDA atomic include in 14562036. Task ownership is handled by task_ptr_ until inner_op_ is started.

ericniebler:
@fallintoplace could you look into these test failures?

ericniebler (approved):

PR #2221, opened 2026-08-21
Fix start_scheduler query of __receiver_proxy_base in __task_scheduler.hpp

yulin25:
## Background
parallel_scheduler causes a compilation error when used in an environment where it is the start scheduler.

## What's changed
- Fall back to inline_scheduler rather than failing compilation if the start scheduler cannot construct a task_scheduler.
- Fall back to inline_scheduler rather than terminating if constructing the task_scheduler throws.
- Add a corresponding parallel_scheduler regression test.

### DIFF
--- a/include/stdexec/__detail/__task_scheduler.hpp
+++ b/include/stdexec/__detail/__task_scheduler.hpp
@@ -1,4 +1,4 @@
-/*
+﻿/*
  * Copyright (c) 2026 NVIDIA Corporation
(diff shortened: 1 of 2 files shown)


ericniebler:
i need to think about this. silently falling back to `inline_scheduler` is almost never the right thing to do.

yulin25:
Superseded by #2225.

PR #2223, opened 2026-08-22
fix(exec): any_sender completion signatures for sequence with 3+ senders

### LINKED ISSUE #2101: `exec::any_sender` fails `get_completion_signatures` inside a `sequence` of 3+ senders (works with 2)
*AUTHOR NOTE*: our recent attempted upgrade to using the latest stdexec failed badly due to the below issue, as reported by Claude. There is a workaround of nested senders, eg. `sequence(a, sequence(b, c))`, which ensures each sequence is limited to two senders, but that feels a bit rubbish so we are reverting to the old version for now. 

### Summary
A type-erased `exec::any_sender` composes fine in a 2-element `exec::sequence`, but putting it in a `sequence` of **three or more** senders fails to compile with `no matching function for call to 'get_completion_signatures'`. A 3-element sequence of **concrete** senders is fine.

### Reproducer
```cpp
// clang++ -std=c++23 -I<stdexec>/include -DCASE=N -fsyntax-only repro.cpp
#include <exception>
#include <exec/any_sender_of.hpp>
#include <exec/sequence.hpp>
using namespace stdexec;
using namespace exec;

using AnyVoid = any_sender<any_receiver<
    completion_signatures<set_value_t(), set_error_t(std::exception_ptr)>>>;
using AnyBool = any_sender<any_receiver<
    completion_signatures<set_value_t(bool), set_error_t(std::exception_ptr)>>>;

#if CASE == 1   // OK: 2-element sequence containing a type-erased sender
auto make() { return sequence(AnyVoid(just()), AnyBool(just(true))); }
#elif CASE == 2 // ERROR: 3-element sequence containing a type-erased sender
auto make() { return sequence(just(), AnyVoid(just()), AnyBool(just(true))); }
#elif CASE == 3 // OK: 3-element sequence of concrete senders
auto make() { return sequence(just(), just(), just(true)); }
#endif
int main() { (void)make(); }
```

| Case | Senders | Result |
|------|---------|--------|
| 1 | `sequence(any, any)` | ✅ compiles |
| 2 | `sequence(just(), any, any)` | ❌ compile error |
| 3 | `sequence(just(), just(), just(true))` | ✅ compiles |

### Error (CASE 2, paths shortened)
```
__sequence.hpp:395: error: no matching function for call to 'get_completion_signatures'
note: __has_get_completion_signatures<const exec::any_sender<…>&,
                                      const cprop<get_domain_t,…>&> evaluated to false
note: candidate (consteval static, with env): constraints not satisfied
      [_Sender = const exec::any_sender<…>&]
note: candidate (consteval static, no env): too many template arguments
```

### Analysis
`exec::sequence(s...)` computes its completion signatures **recursively**: it peels off the first sender and re-queries the **tail as a sub-sequence** (`__seq::__sndr<…>`), which holds its children by **`const&`** (the error shows `__seq::__sndr<const __sexpr&, const any_sender<…>&>`). With 2 senders there is no tail recursion — the senders are queried by value; with 3+ there is.

`any_sender` exposes its completions via `_isender::_interface_`:
```cpp
template <std::derived_from<_interface_> Self, class... Env>
static consteval auto get_completion_signatures();
```
The discovery machinery `STDEXEC_GET_COMPLSIGS` invokes it as
`REMOVE_REFERENCE(_Sender)::template get_completion_signatures<_Sender, _Env...>()`,
forwarding `_Sender` **un-decayed** as `Self`:

- by value → `Self = any_sender` → `derived_from<any_sender, _interface_>` is **true** → works (2-element case).
- through a reference → `Self = const any_sender&` → `derived_from<const any_sender&, _interface_>` is **false** (a reference type is never `derived_from` anything) → no viable overload, and `any_sender` has no `::completion_signatures` typedef / env-taking member to fall back on → the whole `__with` discovery disjunction fails.

So a type-erased `any_sender` cannot have its completion signatures queried through a cv/ref-qualified type, and `sequence`'s ≥3-arg recursion introspects its children precisely that way (`const&`). That's why 2 works, 3+ fails, and 3-concrete works (concrete senders don't have the `derived_from<Self>` constraint).

### Suggested fix
Decay `Self` before the `derived_from` check — either in `any_sender`'s constraint
(`std::derived_from<std::remove_cvref_t<Self>, _interface_>`) or by having the query
machinery pass a decayed sender type as `Self`.

### Environment
- stdexec `main` @ `fee4d651`
- clang 22.1.5, libc++, `-std=c++23`, macOS arm64

alwaysprince05:
When exec::sequence has 3+ senders, it recursively computes completion signatures by peeling off the first sender and querying the tail as a sub-sequence (__seq::__sndr). This tail holds child senders by const&. The get_completion_signatures discovery machinery passes the sender type un-decayed (as const any_sender&), but any_sender's constraint std::derived_from<Self, _interface_> fails because reference types are never derived_from anything.

Fix by decaying Self before the derived_from check in _isender::_interface_::get_completion_signatures().

Fixes #2101

### DIFF
--- a/include/stdexec/__detail/__sequence.hpp
+++ b/include/stdexec/__detail/__sequence.hpp
@@ -183,7 +183,7 @@ namespace STDEXEC
     };
 
     template <class _Child1, class _Child2, class... _Rest>
-    struct __attrs<_Child1, _Child2, _Rest...> : __attrs<__sndr<_Child1, _Child2>, _Rest...>
+    struct __attrs<_Child1, _Child2, _Rest...> : __attrs<__sndr<__decay_t<_Child1>, __decay_t<_Child2>>, __decay_t<_Rest>...>
     {};
 
     //////////////////////////////////////////////////////////////////////////////////////
@@ -379,17 +379,25 @@ namespace STDEXEC
       {
         using __cv_sender1_t = __copy_cvref_t<_Self, _Sender1>;
 
+        // If the first sender is a nested sequence (held by const& in attrs),
+        // decay it for completion signature queries since it's not the "true"
+        // first sender that gets connected directly (see line 328).
+        using __sender1_for_completions_t =
+          __if_c<__is_instance_of<__decay_t<__cv_sender1_t>, __seq::__sndr>,
+                 __decay_t<__cv_sender1_t>,
+                 __cv_sender1_t>;
+
         if constexpr (!__decay_copyable<_Self>)
         {
           return STDEXEC::__throw_compile_time_error<_SENDER_TYPE_IS_NOT_DECAY_COPYABLE_,
                                                      _WITH_PRETTY_SENDER_<_Self>>();
         }
-        else if constexpr (!__sends<set_value_t, __cv_sender1_t, __fwd_env_t<_Env>...>)
+        else if constexpr (!__sends<set_value_t, __sender1_for_completions_t, __fwd_env_t<_Env>...>)
         {
           // If the first sender has no set_value completions, then the second sender will
           // never be started, so just return the (error and stopped) completions of the
           // first sender.
-          return STDEXEC::get_completion_signatures<__cv_sender1_t, __fwd_env_t<_Env>...>();
+          return STDEXEC::get_completion_signatures<__sender1_for_completions_t, __fwd_env_t<_Env>...>();
         }
         else
         {
@@ -398,7 +406,7 @@ namespace STDEXEC
 
           auto __completions1 =  //
             STDEXEC::__transform_completion_signatures(
-              STDEXEC::get_completion_signatures<__cv_sender1_t, __fwd_env_t<_Env>...>(),
+              STDEXEC::get_completion_signatures<__sender1_for_completions_t, __fwd_env_t<_Env>...>(),
               __eat_value_signatures<_Self>{});
 
           auto __completions2 =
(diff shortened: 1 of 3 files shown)


ericniebler:
a `const&` `any_sender` cannot be `connect`-ed. that is why its `get_completion_signatures` is constrained the way it is. i don't know what the right fix is, but this is not it.

include/stdexec/__detail/__sequence.hpp
```diff
@@ -389,7 +389,7 @@ namespace STDEXEC
           // If the first sender has no set_value completions, then the second sender will
           // never be started, so just return the (error and stopped) completions of the
           // first sender.
-          return STDEXEC::get_completion_signatures<__cv_sender1_t, __fwd_env_t<_Env>...>();
+          return STDEXEC::get_completion_signatures<std::remove_cvref_t<__cv_sender1_t>, __fwd_env_t<_Env>...>();
```
ericniebler: No, the cvref-ness of the first sender matters, because it doesn't get decay-copied into the operation state like the other senders. See how it gets `connect`-ed on line 328.

ericniebler (changes requested):

alwaysprince05:
Superseded by #2244 — a simpler 2-line fix at the root cause in any_sender_of.hpp rather than patching __sequence.hpp. The constraint on get_completion_signatures was too restrictive since it's a static consteval function that never connects the sender.

PR #2226, opened 2026-08-24
Fix `exec::thread_pool_base` customization of bulk

### LINKED ISSUE #1687: thread pools based on `execpools::thread_pool_base` are not customizing the `bulk*` algorithms correctly
this problem effects the `asio_thread_pool`, the `taskflow_threadpool`, and the `tbb_thread_pool`. `thread_pool_base` attempts to customize `bulk` (and only `bulk`) as follows:

```c++
      template <stdexec::sender S, std::integral Shape, class Fun>
      STDEXEC_MEMFN_DECL(
        auto bulk)(this const scheduler& sch, S&& sndr, Shape shape, Fun fun) noexcept
        -> bulk_sender_t<S, Shape, Fun> {
        // ...
      }
```

the first thing wrong is that this function does not accept an execution policy. the second wrong thing is that the `STDEXEC_MEMFN_DECL` macro expands to a hidden overload of `tag_invoke`:

```c++
      template <stdexec::sender S, std::integral Shape, class Fun>
      friend auto tag_invoke(bulk_t, const scheduler& sch, S&& sndr, Shape shape, Fun fun) noexcept
        -> bulk_sender_t<S, Shape, Fun> {
        // ...
      }
```

`tag_invoke` is no longer used to find algorithm customizations, so this overload will never be considered.

`thread_pool_base` should do like `static_thread_pool` does by defining a domain with a `transform_sender` member constrained to accept `bulk_chunked_t` and `bulk_unchunked_t` senders that transforms those senders into ones that use `thread_pool_base::bulk_enqueue`.

Cra3z:
Fix the missing customization of `bulk_unchunked` in `exec::thread_pool_base`. Additionally, the current `bulk` customization in `exec::thread_pool_base` executes in parallel regardless of the execution policy, which is incorrect. It should only execute in parallel across multiple execution agents when the policy is `std::execution::par` or `std::execution::par_unseq`; otherwise, it should fall back to sequential execution on a single execution agent.

fixes #1687

### DIFF
--- a/include/exec/thread_pool_base.hpp
+++ b/include/exec/thread_pool_base.hpp
@@ -227,15 +227,12 @@ namespace experimental::execution
         [[nodiscard]]
         auto num_agents_required() const -> std::uint32_t
         {
-          if (!parallelize_)
-          {
-            return 1;
-          }
+          auto const parallelism = parallelize_ ? pool_.available_parallelism()
+                                                : static_cast<std::uint32_t>(1);
 
           // With work stealing, is std::min necessary, or can we feel free to ask for more agents (tasks)
           // than we can actually deal with at one time?
-          return static_cast<std::uint32_t>(
-            (std::min) (shape_, static_cast<Shape>(pool_.available_parallelism())));
+          return static_cast<std::uint32_t>((std::min) (shape_, static_cast<Shape>(parallelism)));
         }
 
         template <class F>
--- a/include/exec/thread_pool_base.hpp
+++ b/include/exec/thread_pool_base.hpp
@@ -28,6 +28,7 @@ import stdexec;
 #else
 #  include "../stdexec/__detail/__execution_fwd.hpp"
 
+#  include "../stdexec/__detail/__bulk.hpp"
 #  include "../stdexec/__detail/__connect.hpp"
 #  include "../stdexec/__detail/__env.hpp"
 #  include "../stdexec/__detail/__meta.hpp"
@@ -81,7 +82,9 @@ namespace experimental::execution
 
     struct domain : STDEXEC::default_domain
     {
-      template <sender_for<STDEXEC::bulk_chunked_t> Sender, class Env>
+      template <class Sender, class Env>
+        requires sender_for<Sender, STDEXEC::bulk_chunked_t>
+              || sender_for<Sender, STDEXEC::bulk_unchunked_t>
       static constexpr auto transform_sender(STDEXEC::set_value_t, Sender&& sndr, Env const & env)
       {
         auto& [tag, data, child] = sndr;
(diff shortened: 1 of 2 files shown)


ericniebler (approved):
LGTM, thanks!

fixes #1687

ericniebler:
> ```
> The following tests FAILED:
> 	818 - exec::bwos::lifo_queue - high contention stress (Timeout)
> ```

@Cra3z, looks like you need to bump up the timeout for this test.

Cra3z:
@ericniebler Uh... odd, I've never touched the `lifo_queue` code. On my local `MSVC 14.51`, the `exec::bwos::lifo_queue - high contention stress` test case doesn't seem to time out.

Cra3z:
@ericniebler I dug into this one — it does reproduce, but I don't think raising the
timeout is the right fix: the test's runtime is a function of how many cores the
runner happens to give us, so any fixed timeout is just a guess.

### Reproduction

Pinning ctest to a single core (plus a few competing spinners, to emulate a
contended runner vCPU) reproduces the CI failure almost exactly:

```
local:  1/1    Test #292: exec::bwos::lifo_queue - high contention stress ...***Timeout  32.10 sec   (ctest exit=8)
CI:     818/1022 Test #818: exec::bwos::lifo_queue - high contention stress ...***Timeout  32.31 sec   (ctest: 8)
```

### Root cause

The owner's push loop yields, but the thief loop has no backoff at all:
owner, line 357 — yields:
https://github.com/NVIDIA/stdexec/blob/67208375d32c90c7cd9285a89f4617a74d61ff20/test/exec/test_bwos_lifo_queue.cpp#L352-L364

thief, line 376 — pure busy-spin:
https://github.com/NVIDIA/stdexec/blob/67208375d32c90c7cd9285a89f4617a74d61ff20/test/exec/test_bwos_lifo_queue.cpp#L372-L389

With 8 thieves burning full quanta on a mostly-failing `steal_front()`, the owner
gets roughly 1/9 of the scheduler's share. Throughput collapses as cores are removed:

| cores | wall time |
|------:|----------:|
| 16    | 0.02 s    |
| 4 (CI runner) | 2.5 s |
| 2     | 10.8 s    |
| 1     | 25.5 s    |

A watchdog confirms `pushCount` climbs linearly the entire time, so this is
throughput collapse rather than a livelock or a deadlock. Two things I ruled out:
ASan is not the driver (0.010 s → 0.018 s at 16 cores), and neither is
msvc 14.50 vs 14.51. The 4-vCPU runner alone only gets to ~2.5 s — the rest
appears to come from the Docker `--isolation=process` overhead and a contended
shared vCPU, which is why it's flaky rather than consistently failing.

### Possible fix

Give the thief loop the same backoff the push loop already has:

```diff
           {
             break;
           }
+          else
+          {
+            std::this_thread::yield();
+          }
         }
         thiefCounts[t] = localCount;
       });
```

ericniebler:
wow thanks @Cra3z! terrific analysis. have you done any benchmarking with the fix?

ericniebler (approved):

PR #2227, opened 2026-08-24
Fix `libdispatch_queue` bulk customization

Cra3z:
Fix `libdispatch_queue` bulk customization and address the issue #2152.


ericniebler (approved):
awesome, thanks!

PR #2228, opened 2026-08-25
Make `get_completion_signatures` align with the standard

yulin25:
According to the [standard](https://eel.is/c++draft/exec.getcomplsigs), `STDEXEC::get_completion_signatures` shall have only one constraint: environment count <= 1.
When the customization does not exist or the member `get_completion_signatures` customization is ill‑formed upon instantiation,  `STDEXEC::get_completion_signatures` should throw an exception, while the invocation of `STDEXEC::get_completion_signatures` remains well-formed.  
The internal `STDEXEC::__compsigs::__get_completion_signatures_helper` already does what the standard says, but `STDEXEC::get_completion_signatures` adds additional constraints, which does not align with the standard.  
This PR removes those additional constraints.

### DIFF
--- a/include/stdexec/__detail/__get_completion_signatures.hpp
+++ b/include/stdexec/__detail/__get_completion_signatures.hpp
@@ -1,4 +1,4 @@
-/*
+﻿/*
  * Copyright (c) 2021-2024 NVIDIA Corporation
  *
  * Licensed under the Apache License Version 2.0 with LLVM Exceptions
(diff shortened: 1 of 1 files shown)


include/stdexec/__detail/__get_completion_signatures.hpp
```diff
@@ -372,11 +372,9 @@ namespace STDEXEC
   //! See @ref get_completion_signatures() for the full description.
   STDEXEC_MODULE_EXPORT
   template <class _Sender, class _Env>
-    requires __has_get_completion_signatures<_Sender, _Env>
```
ericniebler, line 375: please also delete the concept `__has_get_completion_signatures`, and the `__cmplsigs::__with` concept it depends on.

ericniebler (changes requested):

include/stdexec/__detail/__get_completion_signatures.hpp
yulin25: Removed in [a87f907](https://github.com/NVIDIA/stdexec/pull/2228/commits/a87f907217a44ca93075fd98d341c568a0037116).

ericniebler (approved):

Issue #2230, opened 2026-08-25
timed_thread_context can miss an enqueue notification and sleep with a command pending


daniel-schmidt:
[repro.cpp](https://github.com/user-attachments/files/31419075/repro.cpp)

Switching unittests in our codebasis from the Asio thread pool to `timed_thread_context` for timeouts, I observed spurious test failures in about 1 of 10,000 runs. With AI assistence, I was able to trace the problem down to a potential issue in the context's interaction with the `command_queue` and can provide a two-line fix, although I am certainly not fully aware of possible performance implications or unwanted side-effects this change may have.

## Summary

`exec::timed_thread_context` occasionally does not process a newly enqueued timer-cancellation command until its current `wait_until` deadline. This can make cancellation take the complete timer duration (or the scheduler's two-second idle interval), and can change which child of `when_any` wins.

The problem appears to be a mismatch between `__intrusive_mpsc_queue::pop_front()` and the timed scheduler's conditional notification policy:

- `pop_front()` is allowed to return `nullptr` while a producer is between updating `__head_` and  linking the new node through `__prev->next_`.
- `timed_thread_context::schedule()` only notifies its worker when `push_back()` returns `true`. In the problematic interleaving, `push_back()` returns `false`, so the worker goes to sleep even though the producer has now finished linking a pending command.

## Reproducer

Save the attached `repro.cpp`, then build it against the stdexec headers:

```console
g++ -std=c++20 -O2 -pthread -I /path/to/stdexec/include repro.cpp -o repro
./repro
```

An immediate `just(0)` races a timer returning `1`; that result passes through two `continues_on` / `when_any` stages whose timers return `2` and `3`. A normal iteration completes almost immediately with `0`. On my machine, a timer occasionally wins after its complete 100 ms duration or after  2 s (likely the deadline of the `timed_thread_context`) elapsed:

```text
wrong winner 3 in iteration 21029 after 100 ms
```

The iteration number and delay vary. All five consecutive 1,000,000-iteration runs reproduced the problem.

Also a simplified version without the `middle` and the `outer` sender reproduces the issue, but in a more subtle way. Just racing a 100 ms timer against `just(0)`, I observed an unexpectedly long execution time of 2 s instead of immediate completion of the sender during runs with 1,000,000 iterations:

```text
long duration 2000ms with winner 0 in iteration 51535
```

## Expected behavior

[...]

PR #2231, opened 2026-08-25
Fix missing enqueue notifications in timed_thread_context.

daniel-schmidt:
The push_back of the command queue can return false during insertion of a new command.

See #2230 for details.

### DIFF
--- a/include/exec/timed_thread_scheduler.hpp
+++ b/include/exec/timed_thread_scheduler.hpp
@@ -192,7 +192,7 @@ namespace experimental::execution
           while (
             !n_submissions_in_flight_.compare_exchange_weak(expected,
                                                             context_closed,
-                                                            STDEXEC::__std::memory_order_relaxed))
+                                                            STDEXEC::__std::memory_order_acquire))
           {
             STDEXEC::__spin_loop_pause();
             expected = 0;
@@ -235,7 +235,7 @@ namespace experimental::execution
         ready_ = true;
       }
       cv_.notify_one();
-      n_submissions_in_flight_.fetch_sub(1, STDEXEC::__std::memory_order_relaxed);
+      n_submissions_in_flight_.fetch_sub(1, STDEXEC::__std::memory_order_release);
     }
 
     void request_stop()
--- a/include/exec/timed_thread_scheduler.hpp
+++ b/include/exec/timed_thread_scheduler.hpp
@@ -229,12 +229,12 @@ namespace experimental::execution
                                                          STDEXEC::__std::memory_order_relaxed);
         return;
       }
-      if (command_queue_.push_back(op))
+      command_queue_.push_back(op);
       {
         std::scoped_lock lock{ready_mutex_};
         ready_ = true;
-        cv_.notify_one();
       }
+      cv_.notify_one();
       n_submissions_in_flight_.fetch_sub(1, STDEXEC::__std::memory_order_relaxed);
     }
 


include/exec/timed_thread_scheduler.hpp
```diff
@@ -229,12 +229,12 @@ namespace experimental::execution
...
         return;
       }
-      if (command_queue_.push_back(op))
+      command_queue_.push_back(op);
       {
         std::scoped_lock lock{ready_mutex_};
         ready_ = true;
-        cv_.notify_one();
       }
+      cv_.notify_one();
```
ericniebler, line 237: @lewissbaker i am remembering a situation in which a cv notify needed to happen within a critical section. is this one of those situations?

ispeters: I'm about 80% confident this is not required to be done inside the critical section.

The circumstance where you would need to notify inside the critical section is if the awoken work could cause the destruction of the condition variable before the notify method returns.

In this case, you're notifying the cv with a positive number of inflight submissions, which means, in the worst case, the `run` function is spinning on that count, waiting for it to go back to zero, which it can't do until `notify_one` returns. Since the only path to the cv's destruction is through the queue-draining loop in `run`, and since that loop must be no further advanced than spinning on the count of active submissions, I don't think you can run into UB here.

include/exec/timed_thread_scheduler.hpp
```diff
@@ -229,12 +229,12 @@ namespace experimental::execution
                                                          STDEXEC::__std::memory_order_relaxed);
```
ispeters, line 229: Incidentally to this PR, I don't understand why this CAS is here.

If we're running this code then `n < 0` must be true; that can only happen if `run` has exited the busy-loop that's trying to set `n_submissions_in_flight_` to `context_closed`. Given that `run` has successfully exited the busy loop, there's nothing I can see that prevents the current context from being destroyed out from under you here. We're already playing with fire by having discovered that `*this` might be about to be destroyed by dereferencing `this` on line 214; continuing to dereference `this` after we know it's in its death throes seems like all downside and no upside.

Maybe there's a concern that, at shut-down, such a large number of attempts to schedule will happen "all at once" that `context_closed` gets incremented up to 0. I suppose that could theoretically happen and, if it did, it would be disastrous. That could be fixed by changing how "closed" is represented. I'd split `n_submissions_in_flight_` into a one-bit "closed?" field in the LSB, and a count of in-flight submissions in the remaining bits; incrementing and decrementing the count could then be done with `fetch_add(2, …)` and `fetch_sub(2, …)`, and the busy-loop that's trying to mark the context as closed could spinning around `compare_exchange_weak(0, 1, …)`. The check for a negative `n` here would instead be a check for *odd* `n`. It's impossible to accidentally overflow the count field and corrupt the closed field with this strategy.

include/exec/timed_thread_scheduler.hpp
lewissbaker, line 227: Why is this line trying to change 'n' back to 'context_closed'?
The 'n' value obtained above was the value before 1 was added to it, so in the no-contention case this is going to fail (as the value is likely still n+1 rather than n).

This seems like it could potentially just set the value back to 'context_closed' or just decrement the value instead? Or alternatively, turn the above fetch_add into a compare_exchange loop that leaves the value unchanged if it's currently 'context_closed'.

Either way, some commentary here would be helpful for future readers.

lewissbaker, line 200: This loop seems like it could spin for an unbounded amount of time if other threads keep queuing new items onto it as a single other thread could potentially continuously win the race of incrementing the counter to 1, then this loop checks the counter and spins, then the other thread decrements, then immediately enqueues another item (which increments the count to 1).

An alternative would be to atomically set a flag that means that new schedule operations are not permitted once the count reaches zero when the flag is set. This would put a bound on the number of other items that can be enqueued to the number of threads currently calling `schedule()` at the time that the shutdown begins.

Having said that, at a higher level I would probably try to avoid a situation where the users can still potentially schedule more items once the `timed_thread_context` destructor starts. I would instead make it a precondition of the destructor that all scheduled items have completed and that no further items will be scheduled onto the context. This would greatly simplify the shutdown approach - once you see the 'stop-requested' flag you just assert that the queue is empty and exit the `run()` function. You could also then completely get rid of the `n_submissions_in_flight_` data-member.

ericniebler:
@daniel-schmidt could you please look into the clang-16 test failure?

include/exec/timed_thread_scheduler.hpp
lewissbaker: The one thing I'd want to check here is whether or not there is a possibility of the `run()` thread seeing the decrement of `n_submissions_in_flight` before it sees the effects of `notify_one()`.

The decrement below has relaxed memory order and I don't think that the `notify_one` member-function provides acquire semantics that would prevent the subsequent access of `n_submissions_in_flight_` from being reordered above the `notify_one()` call. So whether or not this is safe might depend on exactly how the `notify_one()` call is implemented - I'm not 100% sure on this one.

~I would probably keep the call to `notify_one()` inside the scope of the lock just to be safe here.~

EDIT: Actually, on reflection I think the better solution is to make the decrement below a release operation and then the compare-exchange in run() an acquire operation.

daniel-schmidt:
> @daniel-schmidt could you please look into the clang-16 test failure?

There is a failure in test_parallel_scheduler.cpp:633 As far as I'm aware, this is not using a timed_thread_scheduler or any of the code I touched.

ericniebler:
> There is a failure in test_parallel_scheduler.cpp:633 As far as I'm aware, this is not using a timed_thread_scheduler or any of the code I touched.

agreed, this is likely an unrelated failure. i am suspicious of the relaxed atomic operations in `destructor_tracked_value`. i'll make a separate PR for that.

ericniebler:
@lewissbaker i am inclined to merge this PR as-is and file issues for the points you raise above. objections?

include/exec/timed_thread_scheduler.hpp
ericniebler: patches welcome. i'm unlikely to investigate this unless there's a bug here.

PR #2234, opened 2026-08-26
Fix `__stop_when` completion signatures

yulin25:
When computing `__stop_when`'s completion signatures, the implementation does not give the child sender the correct environment.

### DIFF
--- a/include/stdexec/__detail/__spawn_future.hpp
+++ b/include/stdexec/__detail/__spawn_future.hpp
@@ -230,7 +230,7 @@ namespace STDEXEC
 
       using __op_t = connect_result_t<__future_spawned_sender<_Sender, _Env>, __receiver_t>;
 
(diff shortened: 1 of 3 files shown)


include/stdexec/__detail/__spawn_future.hpp
```diff
@@ -230,7 +230,7 @@ namespace STDEXEC
 
       using __op_t = connect_result_t<__future_spawned_sender<_Sender, _Env>, __receiver_t>;
 
-      using __base = __spawn_future_state_base<completion_signatures_of_t<_Sender, _Env>>;
+      using __base = typename __spawn_future_state::__spawn_future_state_base;
```
ericniebler:
```suggestion
      using __base = __spawn_future_state::__spawn_future_state_base;
```

PR #2237, opened 2026-08-26
Fix missing trailing return type in `__when_all_impl::__get_env`

maartenarnst:
In #2124, the trailing return type was removed from `__when_all_impl::__get_env`. This forces early instantiation of downstream `get_env`, which in turn may cause compilation errors when downstream types are still incomplete. This PR restores the trailing return type.

@RobertLeahy


ericniebler (approved):
thanks!

PR #2243, opened 2026-08-31
ci: build modules with Clang 23 only; document modular-build support

ispeters:
Clang's module-serialization ICE (a crash inside
ASTReader::readTypeRecord / FunctionProtoType::Profile while deserializing from a BMI) makes the modular build unreliable on Clang
22. It reproduces on every released 22.1.x, including 22.1.8, which is the last planned 22.1.x release, so no released Clang 22 will ever build the modular tests. Which translation unit crashes is not stable -- it shifts with incidental state such as the build environment or unrelated source changes -- so an unrelated contribution can fail CI for reasons that have nothing to do with it, and the exclusion list cannot be made reliable by adding entries to it.

The crash is gone in Clang 23.1.0. Move the modular CI job to Clang 23, declare Clang 23 the supported toolchain for modular builds in the README, and drop the ICE exclusion list: those tests all build once the compiler is fixed.

rapidsai/devcontainers does not publish an llvm23 image yet, so the job installs Clang 23 from apt.llvm.org on top of the llvm22 image. The install step no-ops once the image tag matches the requested LLVM version, so de-hacking is a two-line matrix change.

test/stdexec/detail/test_any_allocator.cpp stays excluded in modules mode for an unrelated, stdexec-side reason: __any_allocator is reachable but not exported from the stdexec module.


ericniebler (approved):

ispeters:
This is clearly not going to work given that Clang 23.1.0 ICEs with the same ICE as Clang 22.whatever did. I'm going to work on reducing the repro with `cvise` and see if I can get Clang fixed. Meanwhile, I'm sad about losing CI coverage of the modular build but I don't want to impose too much burden on other contributors. Is there a way to keep it running so I can see the signal it generates without ICE failures blocking other PRs?

ericniebler:
ok, closing

PR #2244, opened 2026-08-31
fix(exec): ignore sender cvref-qualification when computing sender attributes

### LINKED ISSUE #2101: `exec::any_sender` fails `get_completion_signatures` inside a `sequence` of 3+ senders (works with 2)

alwaysprince05:
Fixes #2101 (supersedes #2223 and the earlier approach in this PR).

When exec::sequence has 3+ senders, the recursive __seq::__sndr nesting stores intermediate types by const& in __attrs. This causes __copy_cvref_t to propagate reference qualifiers into the sender type passed to get_completion_signatures. Reference types never satisfy derived_from, which breaks type-erased senders (e.g. any_sender).

Root cause

The __attrs 3+ children specialization creates intermediate __seq::__sndr types with the raw (possibly reference) child types. When get_completion_signatures is called on these intermediate types, __copy_cvref_t<_Self, _Sender1> produces a reference type, and derived_from fails because reference types can't derive from anything.

Fix (2 targeted changes in __sequence.hpp only)

Decay child types in __attrs 3+ children specialization to prevent intermediate __seq::__sndr types with reference types.
Use __decay_t<__cv_sender1_t> for completion-signature queries only; the connect path still uses the cvref-qualified type.
No changes to any_sender_of.hpp — its derived_from constraint is correct.

### DIFF
--- a/include/nvexec/stream/let_xxx.cuh
+++ b/include/nvexec/stream/let_xxx.cuh
@@ -72,8 +72,7 @@ namespace nv::execution::_strm
     inline constexpr auto _mk_sch_env =
       []<class CvSender, class Receiver, class SetTag>(CvSender&& sndr, Receiver&& rcvr, SetTag)
     {
-      using cv_fn = __copy_cvref_fn<CvSender>;
-      return __mk_secondary_env_t<SetTag>()(cv_fn{}, sndr, STDEXEC::get_env(rcvr));
+      return __mk_secondary_env_t<SetTag>()(sndr, STDEXEC::get_env(rcvr));
     };
 
     template <class CvSender, class Receiver, class SetTag>
--- a/include/stdexec/__detail/__finally.hpp
+++ b/include/stdexec/__detail/__finally.hpp
@@ -225,7 +225,6 @@ namespace STDEXEC
       using __env2_t            = __final::__env2_t<_CvInitialSender, env_of_t<_Receiver>>;
       using __base_t            = __final_opstate_t<_CvInitialSender, _CvFinalSender, _Receiver>;
       using __initial_results_t = __base_t::__results_t;
-      using __cv_fn             = __copy_cvref_fn<_CvInitialSender>;
 
       constexpr explicit __opstate(_CvInitialSender&& __initial,
                                    _CvFinalSender&&   __final,
@@ -233,7 +232,7 @@ namespace STDEXEC
         : __base_t(&__cleanup_initial_opstate,
                    static_cast<_CvFinalSender&&>(__final),
                    static_cast<_Receiver&&>(__rcvr),
-                   __mk_secondary_env_t{}(__cv_fn{}, __initial, get_env(__rcvr)))
+                   __mk_secondary_env_t{}(__initial, STDEXEC::get_env(__rcvr)))
       {
         __initial_opstate_.__emplace_from(STDEXEC::connect,
                                           static_cast<_CvInitialSender&&>(__initial),
--- a/include/stdexec/__detail/__let.hpp
+++ b/include/stdexec/__detail/__let.hpp
@@ -169,16 +169,15 @@ namespace STDEXEC
       using __env2_t        = _Env2;
       using __second_rcvr_t = __rcvr_env<_Receiver, _Env2>;
 
-      template <class _CvFn, class _Sender>
+      template <class _Sender>
       constexpr explicit __opstate_base(_SetTag,
-                                        _CvFn           __cv,
                                         _Sender const & __sndr,
                                         _Fun            __fn,
                                         _Receiver&&     __rcvr) noexcept
         : __rcvr_(static_cast<_Receiver&&>(__rcvr))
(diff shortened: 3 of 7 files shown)


include/exec/any_sender_of.hpp
```diff
@@ -514,7 +514,8 @@ namespace experimental::execution
         using sender_concept = STDEXEC::sender_tag;
         using _base_t::_base_t;
 
-        template <__std::derived_from<_interface_> Self, class... Env>
+        template <class Self, class... Env>
+          requires __std::derived_from<__decay_t<Self>, _interface_>
```
ericniebler: i'm afraid this fix is wrong too. if `sender_in<S, E>` is `true`, then a sender of type `S` should be `connect`-able with a receiver whose environment has type `E`. after this change, that is no longer true for `any_sender`. only rvalue `any_sender`s can be `connect`ed, so `any_sender::get_completion_signatures` is right to reject lvalue `any_sender`s.

ericniebler (changes requested):

ericniebler (approved):

ericniebler:
thanks @alwaysprince05 !

PR #2245, opened 2026-08-31
fix the atomic memory orderings in `destructor_tracked_value` in test_parallel_scheduler.cpp

ericniebler:
... to address a spurious failure in the `"bulk on parallel_scheduler destroys stored predecessor values"` test.

### DIFF
--- a/test/stdexec/schedulers/test_parallel_scheduler.cpp
+++ b/test/stdexec/schedulers/test_parallel_scheduler.cpp
@@ -517,7 +517,7 @@ struct destructor_tracked_value
 
   ~destructor_tracked_value()
   {
-    live_->fetch_sub(1, std::memory_order_relaxed);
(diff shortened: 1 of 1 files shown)


test/stdexec/schedulers/test_parallel_scheduler.cpp
```diff
@@ -637,7 +637,7 @@ TEST_CASE("bulk on parallel_scheduler destroys stored predecessor values",
     REQUIRE(result.has_value());
   }
 
-  CHECK(live->load(std::memory_order_relaxed) == 0);
+  CHECK(live->load(std::memory_order_acquire) == 0);
```
lewissbaker: I don't think this needs to be an 'acquire' operation.
There aren't any other operations that occur after the load for which you need visibility of things that occurred before the corresponding 'release' (at least that I can tell...).

ericniebler:
this is not fixing the real issue

PR #2248, opened 2026-09-03
Fix starts_on's sender-attributes

### LINKED ISSUE #2238: sync_wait / co_await of starts_on(task_scheduler{...}, task) fails a completion-domain static_assert
Starting a `stdexec::task` with `starts_on` on a `stdexec::task_scheduler` produces a sender
that I cannot `sync_wait`, and cannot `co_await` from another task. The same code with the
concrete scheduler compiles.

```cpp

namespace ex = stdexec;

auto child() -> ex::task<void> { co_return; }

int main() {
  ex::run_loop loop;
  ex::sync_wait(
      ex::starts_on(ex::task_scheduler{loop.get_scheduler()}, child()));
}
```

```
g++ -std=c++26 -Iinclude -fsyntax-only repro.cpp
```

fails with:

```
include/stdexec/__detail/__domain.hpp:297:21: error: static assertion failed: the sender claims to complete on a domain that is not the domain of its completion scheduler
  297 |       static_assert(__same_as<_Domain, __detail::__scheduler_domain_t<_Sch, _Env const &...>>,
```

clang 21.1.8 reports the same static_assert at the same line.

The same failure appears when the sender is awaited from another task:

```cpp
auto parent() -> ex::task<void> {
  auto sched = co_await ex::get_start_scheduler();
  co_await ex::starts_on(sched, child());
}

int main() {
  ex::run_loop loop;
  ex::sync_wait(ex::starts_on(loop.get_scheduler(), parent()));
}
```

Here `sched` deduces to `stdexec::task_scheduler`, even though `parent()` itself is
started on the concrete `run_loop::scheduler`.

Variants, all with the same `child()`:

| variant | result |
| --- | --- |
| `sync_wait(starts_on(task_scheduler{s}, child()))` | **fails** |
| `co_await starts_on(task_scheduler{s}, child())` inside a task | **fails** |
| `sync_wait(starts_on(s, child()))` with the concrete scheduler `s` | compiles |
| `sync_wait(starts_on(task_scheduler{s}, just()))` | compiles |
| `sync_wait(starts_on(task_scheduler{s}, just() \| let_value(child)))` | compiles |
| `sync_wait(child() \| write_env(prop{get_start_scheduler, task_scheduler{s}}))` | compiles |
| `connect(starts_on(task_scheduler{s}, child()), some_receiver{})` | compiles |
| `spawn(starts_on(task_scheduler{s}, child()) \| upon_error(...), scope.get_token())` | compiles |
| `completion_signatures_of_t<decltype(starts_on(task_scheduler{s}, child())), env<>>` | compiles |

Is `starts_on` with a `task_scheduler` and a task child expected to be usable with
`sync_wait` / `co_await`, or is the type-erased scheduler meant to be used differently here?

Tested at `f91f6363` with g++ 15.2.0 and clang 21.1.8, both `-std=c++26`.

Cra3z:
Fixes #2238:
`__starts_on::__attrs` answers `get_completion_scheduler_t<Tag>` in two ways:

* if the child completes where it starts (`__completes_where_it_starts`), it returns `_Scheduler`, because the child is started on `_Scheduler` and never leaves it;
* otherwise it forwards the query to the child's attributes with the augmented environment `env2`.

But it answered `get_completion_domain_t<Tag>` in only one way: always forward to the child's attributes with `env2`.

For a `task<>` child this matters. `task`'s completion behavior is `asynchronous_affine | inline_completion`: affine, but not *purely* inline. `get_completion_domain_t::__get_domain` only consults `get_domain(env2)` for a purely inline child, and `task` exposes neither a domain nor a completion scheduler, so the forwarded query falls through to `default_domain`. Meanwhile the completion-scheduler query returns `task_scheduler`, whose domain is `task_scheduler_domain`. `default_domain != task_scheduler_domain` and the static_assert fires.

### DIFF
--- a/include/stdexec/__detail/__starts_on.hpp
+++ b/include/stdexec/__detail/__starts_on.hpp
@@ -65,32 +65,9 @@ namespace STDEXEC
         , __attr_(__attr)
       {}
 
-      // Query for completion scheduler to use for algorithm dispatching.
-      // NOT TO SPEC
-      template <class... _Env>
-      STDEXEC_ATTRIBUTE(nodiscard, always_inline, host, device)
-      constexpr auto query(get_completion_domain_t<>, _Env&&...) const noexcept
-        -> __completion_domain_of_t<set_value_t, _Scheduler, _Env...>
-      {
-        return {};
-      }
-
-      // Query for completion scheduler
-      template <class _SetTag, class... _Env>
-        requires __completes_where_it_starts<_SetTag, env_of_t<_Child>, __env2_t<_Env>...>
-      STDEXEC_ATTRIBUTE(nodiscard, always_inline, host, device)
-      constexpr auto query(get_completion_scheduler_t<_SetTag>, _Env&&...) const noexcept
-        -> _Scheduler
-      {
-        // If the child completes where it starts, then starts_on(Sch,Child) completes on
-        // scheduler Sch.
-        return __sched_;
-      }
-
       // Query for completion scheduler - delegates to child's env with augmented
       // environment
       template <class _SetTag, class... _Env>
-        requires(!__completes_where_it_starts<_SetTag, env_of_t<_Child>, __env2_t<_Env>...>)
       STDEXEC_ATTRIBUTE(nodiscard, always_inline, host, device)
       constexpr auto query(get_completion_scheduler_t<_SetTag> __query,
                            _Env&&... __env) const noexcept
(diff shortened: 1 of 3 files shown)


ericniebler (changes requested):
> `get_completion_domain_t::__get_domain` only consults `get_domain(env2)` for a purely inline child

this is the real problem. i've created #2250 to fix that. if you rebase this PR on that one, i think you can simplify the `get_completion_scheduler` query of `__starts_on::__attrs` to mirror the current `get_completion_domain` query.

Cra3z:
@ericniebler Updated, please review when you get a chance.

ericniebler (approved):

ericniebler:
thanks!

PR #2249, opened 2026-09-03
fix(exec): query the frame allocator from the receiver's environment

SamGaaWaa:
exec::function chose its frame allocator by calling __choose_frame_allocator with the *receiver* itself instead of the receiver's environment. Allocation always fell back to std::allocator.

### DIFF
--- a/include/exec/function.hpp
+++ b/include/exec/function.hpp
@@ -141,14 +141,15 @@ namespace experimental::execution
       using __receiver_t   = __receiver_wrapper<__any_receiver_ref<_Sigs, _Queries>>;
       using __prop_t       = __receiver_t::__prop_t;
       using __stop_token_t = stop_token_of_t<env_of_t<__receiver_t>>;
-      using __adaptee_t    = decltype(__choose_frame_allocator(__declval<_Receiver const &>()));
+      using __adaptee_t    = decltype(__choose_frame_allocator(
+        __declval<env_of_t<_Receiver> const &>()));
 
       __memory_resource_adaptor_t<__adaptee_t> __resource_;
       __prop_t                                 __env_;
       _any::_state<_Receiver, __stop_token_t>  __rcvr_;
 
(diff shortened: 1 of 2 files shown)


include/exec/function.hpp
```diff
@@ -141,14 +141,15 @@ namespace experimental::execution
       using __receiver_t   = __receiver_wrapper<__any_receiver_ref<_Sigs, _Queries>>;
       using __prop_t       = __receiver_t::__prop_t;
       using __stop_token_t = stop_token_of_t<env_of_t<__receiver_t>>;
-      using __adaptee_t    = decltype(__choose_frame_allocator(__declval<_Receiver const &>()));
+      using __adaptee_t    = decltype(__choose_frame_allocator(
+        __declval<env_of_t<_Receiver> const &>()));
```
ericniebler:
```suggestion
      using __adaptee_t    = __result_of<__choose_frame_allocator, env_of_t<_Receiver>>;
```

include/exec/function.hpp
```diff
@@ -141,14 +141,15 @@ namespace experimental::execution
...
 
       __memory_resource_adaptor_t<__adaptee_t> __resource_;
       __prop_t                                 __env_;
       _any::_state<_Receiver, __stop_token_t>  __rcvr_;
 
       explicit __opstate_base(_Receiver __rcvr)
-        : __resource_(__choose_frame_allocator(std::as_const(__rcvr)))
+        : __resource_(__choose_frame_allocator(STDEXEC::get_env(std::as_const(__rcvr))))
```
ericniebler: preexisting issue, but `as_const` is unnecessary here. it instantiates a function template for no reason.
```suggestion
        : __resource_(__choose_frame_allocator(STDEXEC::get_env(__rcvr)))
```

ericniebler (approved):
nice catch! a couple of minor tweaks and this one is good.

PR #2251, opened 2026-09-04
fix(examples): replace switch statement with std::popcount in sudoku example

### LINKED ISSUE #2105: example.sudoku: assert in examine_potentials aborts the example in non-NDEBUG builds
The `example.sudoku` example aborts on its very first solve in any build that does not define `NDEBUG` (e.g. `-DCMAKE_BUILD_TYPE=Debug`, sanitizer builds, distro packages built without `-DNDEBUG`):

```
$ cmake -S . -B build -DCMAKE_BUILD_TYPE=Debug -DSTDEXEC_BUILD_EXAMPLES=ON
$ cmake --build build --target example.sudoku
$ ./build/examples/example.sudoku
Assertion failed: (!"potential set is not a power of 2"),
function examine_potentials, file sudoku.cpp, line 282.
Abort trap: 6     # exit 134
```

Reproduces on current `main` ([`fee4d651`](https://github.com/NVIDIA/stdexec/commit/fee4d651)), macOS arm64, Apple Clang, `-std=gnu++20`. The flags emitted by CMake for the Debug configuration are simply `-g -std=gnu++20 ...`, with no `-DNDEBUG`, so `assert` is active.

### Diagnosis

Adding a print before the failing assert shows the very first call into `examine_potentials` already hits the `default` branch:

```
DBG: cell i=0 solved_element=1 potential_set=0x0
```

i.e. cell 0 of `init_values0` (which starts as a fixed `1`) reaches `examine_potentials` with `potential_set == 0`, falls through the early "empty set" check, and lands in the `switch`'s `default`.

### Root cause

[`calculate_potentials`](https://github.com/NVIDIA/stdexec/blob/main/examples/sudoku.cpp#L183-L199) unconditionally clears `potential_set` for every cell and only fills in candidate bits for **unsolved** cells:

```cpp
for (unsigned i = 0; i < BOARD_SIZE; ++i) {
  b[i].potential_set = 0;
  if (!b[i].solved_element) {
    // ... compute candidate bitmask ...
    b[i].potential_set |= 1 << (potential - 1);
  }
}
```

Every already-solved cell therefore enters `examine_potentials` with `potential_set == 0`. The early-out at the top of [`examine_potentials`](https://github.com/NVIDIA/stdexec/blob/main/examples/sudoku.cpp#L217-L289) only treats `potential_set == 0` as "empty set / dead branch" when the cell is **also** unsolved:

```cpp
if (b[i].solved_element == 0 && b[i].potential_set == 0)  // empty set
  return false;
switch (b[i].potential_set) {
  case 1:  /* singleton -> promote */ break;
  ...
  case 256: ...
  default:
    assert(!"potential set is not a power of 2");   // sudoku.cpp:282
}
```

For an already-solved cell (`solved_element != 0`, `potential_set == 0`) control falls through into `switch(0)` -> `default` -> `assert`. There is also a second legal state the `switch` does not cover: an unsolved cell with multiple candidates, where `potential_set` is e.g. `3 = 1|2` or `7 = 1|2|4` and is not a power of two.

In other words, the assertion encodes an invariant ("`potential_set` is always a power of two here") that the function itself never establishes. Its actual intent is to **promote singleton candidate sets** into `solved_element` and leave every other cell untouched.

### Suggested fix

Make the singleton filter explicit, instead of relying on the `default` `assert` being compiled out:

```cpp
bool examine_potentials(board_element *b, bool *progress)
{
  bool singletons = false;
  for (unsigned i = 0; i < BOARD_SIZE; ++i)
  {
    if (b[i].solved_element != 0)
      continue;                      // already fixed; potential_set is 0 by construction
    if (b[i].potential_set == 0)     // empty set -> dead branch
      return false;
    // multiple candidates (not a power of 2): leave for partial_solve to branch on
    if ((b[i].potential_set & (b[i].potential_set - 1)) != 0)
      continue;
    switch (b[i].potential_set) {
      /* unchanged singleton -> solved_element mapping */
    }
  }
  *progress = singletons;
  return valid_board(b);
}
```

`x & (x - 1) == 0` is the standard power-of-two test (`x == 0` already excluded above). With the filter in place the `default` branch becomes genuinely unreachable; I'd remove it, but happy to keep it as a defensive `assert` if maintainers prefer.

After this change, `example.sudoku` runs to completion in both Debug and Release on all four built-in boards, in both `find-one` and full-enumeration modes.

alwaysprince05:
Fixes #2105 - Replace switch statement with std::popcount in sudoku example to fix assert failure in Debug mode


examples/sudoku.cpp
```diff
@@ -221,67 +222,13 @@ bool examine_potentials(board_element *b, bool *progress)
   {
     if (b[i].solved_element == 0 && b[i].potential_set == 0)  // empty set
       return false;
-    switch (b[i].potential_set)
+    // Check if potential_set is a power of 2 (only one bit set)
+    if (b[i].potential_set != 0 && std::popcount(b[i].potential_set) == 1)
```
ericniebler: couldn't this simply be:
```suggestion
    if (std::popcount(b[i].potential_set) == 1)
```

alwaysprince05: Done! Simplified to if (std::popcount(b[i].potential_set) == 1) — the zero-check was redundant since std::popcount(0) == 0 which never equals 1.

alwaysprince05:
Hi @ericniebler — I rebased this branch onto current `main` (now up to date, head is `e25bbd7`). The review feedback is applied, so could you run `/ok to test e25bbd7`? The previous `/ok to test f7055d2` pointed at the old pre-rebase SHA, so CI is still waiting on approval. Thanks!

PR #2256, opened 2026-09-14
fix: provide the scheduler_concept nested alias on all stdexec schedulers (#2134)

alwaysprince05:
Fixes #2134

## What happened

Users couldn't start an `exec::task` on an `exec::static_thread_pool` or `exec::timed_thread_context` scheduler — the code failed to compile:

```
include/exec/task.hpp:163:33: error: no viable overloaded '='
```

The task's internal `__any_scheduler` converting constructor is guarded by `__semi_scheduler`, which requires a `scheduler_concept` typedef derived from `scheduler_tag`. Per [exec.sched], that nested alias is part of the `scheduler` concept, so scheduler types are expected to define it — but a number of the schedulers in stdexec did not.

## The fix (updated after review)

Rather than loosening the internal `__semi_scheduler` concept (the original approach of this PR), this PR now keeps `include/exec/task.hpp` exactly as it is on `main` and instead makes all of stdexec's own schedulers provide the required nested alias (`using scheduler_concept = scheduler_tag;`):

- `exec::static_thread_pool::scheduler`
- `exec::thread_pool_base<...>::scheduler` (also covers `exec::tbb::tbb_thread_pool` and `exec::taskflow::taskflow_thread_pool`)
- `exec::timed_thread_scheduler`
- `exec::libdispatch_scheduler`
- `exec::io_uring_context::__scheduler` (Linux)
- `exec::windows_thread_pool::scheduler` (Windows)
- `exec::trampoline_scheduler`'s scheduler
- `exec::reschedule`'s internal scheduler
- `stdexec::parallel_scheduler`
- `nvexec::stream_scheduler` and `nvexec::multi_gpu_stream_scheduler`

Schedulers that already defined the alias are unchanged (`inline_scheduler`, `run_loop`'s scheduler, `any_scheduler`, and the test-common schedulers).

## Tests

Two regression tests in `test/exec/test_task.cpp` co_start an `exec::task` on `exec::static_thread_pool` and `exec::timed_thread_context` and verify it actually runs, guarding the fix going forward.

## Verification

- `test.exec`: 3401 assertions / 359 cases pass (Release, Apple Clang)
- `test.stdexec`: 3528 assertions / 619 cases pass

Issue #2134, opened 2026-07-22
Build error when starting exec::task in some schedulers, because they miss the scheduler_concept typedef


miwechner:
Hi,

we recently noticed that you cannot build a program where an `exec::task` is started in some schedulers. In particular `exec::static_thread_pool` and `exec::timed_thread_context` are affecting us.

This was the most minimal example that i could find:

```cpp
#include <exec/static_thread_pool.hpp>
#include <exec/task.hpp>
#include <stdexec/execution.hpp>

auto main() -> int
{
  exec::static_thread_pool ctx;
  stdexec::sync_wait(stdexec::starts_on(ctx.get_scheduler(), [&]()-> exec::task<void>{
    co_return;
  }()));
  return 0;
}
```

<details>
<summary>Compile error</summary>

```
[main] Building folder: /home/coder/stdexec/build/base-debug example.hello_world
[build] Starting build
[proc] Executing command: /usr/bin/cmake --build /home/coder/stdexec/build/base-debug --target example.hello_world --
[build] [0/2] Re-checking globbed directories...
[build] [1/3] Building CXX object examples/CMakeFiles/example.hello_world.dir/hello_world.cpp.o
[build] FAILED: examples/CMakeFiles/example.hello_world.dir/hello_world.cpp.o 
[build] /usr/bin/g++ -DTBB_USE_DEBUG -isystem /home/coder/stdexec/include -isystem /home/coder/stdexec/build/base-debug/include -g -std=gnu++20 -Wno-maybe-uninitialized -fcoroutines -fconcepts-diagnostics-depth=10 -Wall -Werror=unused-parameter -Wno-non-template-friend -MD -MT examples/CMakeFiles/example.hello_world.dir/hello_world.cpp.o -MF examples/CMakeFiles/example.hello_world.dir/hello_world.cpp.o.d -o examples/CMakeFiles/example.hello_world.dir/hello_world.cpp.o -c /home/coder/stdexec/examples/hello_world.cpp
[build] In file included from /home/coder/stdexec/examples/hello_world.cpp:19:
[...]

PR #2257, opened 2026-09-15
test: add scheduler_concept regression guard for all stdexec schedulers

alwaysprince05:
Follow-up to #2256.

While working on #2134 I noticed nothing actually stops this from coming back — a
new scheduler can forget the `scheduler_concept` alias and users end up with the
same confusing `no viable overloaded '='` error inside `exec::task` again. So this
PR adds a small compile-time guard, test-only, that fails to build if any
scheduler in stdexec drops the alias.

What's here:

- `test/exec/test_scheduler_concept.cpp` (new): statically checks that the
  always-built schedulers — inline, run_loop, task_scheduler, parallel,
  static_thread_pool, thread_pool_base-derived pools, timed_thread, trampoline,
  and reschedule's internal scheduler — define the alias and satisfy
  `scheduler<>`.
- The gated ones (io_uring, windows_thread_pool, libdispatch, nvexec stream and
  multi-GPU stream) are asserted in their existing gated test files, so no extra
  CMake plumbing is needed on other platforms.

Verification: `test.exec` — 3417 assertions / 362 cases pass (Release, Apple Clang 21).

Happy to move or restructure this if there's a better home for the guard.


alwaysprince05:
Hi @ericniebler — small test-only follow-up to #2256: compile-time guards so the scheduler_concept alias from #2134 can't silently regress. Could you /ok to test when you get a chance? Thanks!

ericniebler (approved):

alwaysprince05:
Pushed a one-line clang-format fix (0f216e4) — the style check should be green now, but the new workflow run is sitting in "awaiting maintainer approval". Could you approve it when you get a chance? Thanks!

PR #2262, opened 2026-09-17
Fix issue-2140

### LINKED ISSUE #2140: Compile error of task<mp-units::quanitity> type on gcc. clang ok.
https://godbolt.org/z/59bsThYc5

I don't really know whats happening in stdexec itself. However clang seems to be ok. The godbolt like has a reproducable code snippet

Cra3z:
Fixes #2140:

`__cmplsigs::__partition_completion_signatures_t` grouped the completion signatures by
folding over an overloaded `operator*`, with the intentionally incomplete
`__undefined<__partitions<...>>` as the accumulator and `Tag(*)(Args...)` function
pointers as the operands.

When **ADL** for `operator*`, the associated namespaces of `set_value_t(*)(Litre)` include
those of `Litre` (that is `mp_units::quantity<mp_units::si::litre, double>`), so any `operator*` declared alongside the user's value or error type
becomes a candidate. `mp_units::quantity` declares hidden-friend multiplication operators
constrained with `std::derived_from<quantity> Q` (see https://github.com/mpusz/mp-units/blob/27d2def9082ce00d7eb4f75695dbead4a748f23f/src/core/include/mp-units/framework/quantity.h#L562-L584
). During overload resolution that
constraint is checked with `Q = __undefined<...>`, `__undefined` isn't a complete type, so **GCC** reports a hard error:

```
concepts:76:28: error: invalid use of incomplete type
  'struct stdexec::__undefined<stdexec::__cmplsigs::__partitions<...>>'
    concept derived_from = __is_base_of(_Base, _Derived)
```

The bug is not specific to `mp-units`: any type that comes with a constrained `operator*`
reachable by ADL possibly triggers it, in particular when the constraint is built on a concept
or type trait whose precondition requires a **complete** type. That covers most of the traits in
[[meta.unary.prop]](https://eel.is/c++draft/type.traits#tab:meta.unary.prop), such as
`std::is_constructible`, `std::is_trivially_copyable`, or `std::is_nothrow_move_assignable`. Minimal reproducer (see https://godbolt.org/z/Kq9bcb7Ph):

```cpp

namespace ns {
  struct Foo {
    template <std::derived_from<Foo> Q, class V>
    friend auto operator*(const Q& q, const V&) -> Q { return q; }
  };
}

int main() {
  stdexec::sync_wait(stdexec::just(ns::Foo{}));
}
```


include/stdexec/__detail/__completion_signatures.hpp
```diff
@@ -156,53 +156,40 @@ namespace STDEXEC
...
+    struct __partition_helper<__partitions<_Values, _Errors, _Stopped>, set_stopped_t(), _Sigs...>
+      : __partition_helper<__partitions<_Values, _Errors, __mlist<set_stopped_t()>>, _Sigs...>
+    {};
 
     template <class... _Sigs>
-    using __partition_completion_signatures_t =  //
-      decltype(__cmplsigs::__unpack_partitioned_completions(
-        (__declval<__undefined<__partitions<>>&>() * ... * static_cast<_Sigs*>(nullptr))));
+    using __partition_completion_signatures_t =
+      typename __partition_helper<__partitions<>, _Sigs...>::type;
```
ericniebler:
```suggestion
    using __partition_completion_signatures_t = __partition_helper<__partitions<>, _Sigs...>::type;
```

ericniebler (changes requested):

ericniebler:
thanks!

PR #2264, opened 2026-09-17
Fix(sender_concepts): argument order in implementation of `__has_eptr_completion`

### LINKED ISSUE #2263: Bug: argument order inverted in implementation of `__has_eptr_completion`
The order of `__q1<__is_eptr_t>` and `__qq<__mor_t>` in

https://github.com/NVIDIA/stdexec/blob/abcdee3d1cef5bd1ff5d22d5cb11c9364e58cf67/include/stdexec/__detail/__sender_concepts.hpp#L276-L280

appears to be inverted compared with the expected order:

https://github.com/NVIDIA/stdexec/blob/abcdee3d1cef5bd1ff5d22d5cb11c9364e58cf67/include/stdexec/__detail/__completion_signatures.hpp#L220-L222

As a consequence, the concept currently always yields false.

The impact of this bug may be quite significant. Indeed, for example, with

https://github.com/NVIDIA/stdexec/blob/abcdee3d1cef5bd1ff5d22d5cb11c9364e58cf67/include/stdexec/__detail/__starts_on.hpp#L210

and 

https://github.com/NVIDIA/stdexec/blob/abcdee3d1cef5bd1ff5d22d5cb11c9364e58cf67/include/stdexec/__detail/__sequence.hpp#L372-L376

the current behavior is that the exemptions are never true, so that a sender like

```
auto sched = STDEXEC::get_parallel_scheduler();
auto sndr = ex::starts_on(sched, ex::just() | ex::then([] {}))
```

is always a dependent sender:

```
static_assert(ex::dependent_sender<decltype(sndr)>);
```

As such, one example of the impact is that if the apparent bug is fixed, a `starts_on` sender is no longer necessarily a dependent sender.

I will propose a PR with the fix itself:
- #2264 

I am not sure about the best strategy to address the impact of the resolution. It may depend for example on whether or not a `starts_on` sender is expected to always report it is a dependent sender.

Observed while working on the impact of #2248 on our code base.

Joint analysis with @claude.

maartenarnst:
This PR solves the apparent bug reported in:
- #2263

Note that the PR does not address consequences of the resolution of the bug, such as the example described in the issue.


ericniebler:
the new test needs a clang-format

test/stdexec/concepts/test_concepts_sender.cpp
```diff
@@ -217,6 +217,12 @@ namespace
     STATIC_REQUIRE(ex::sender_of<my_sender_int, ex::set_value_t(int)>);
   }
 
+  TEST_CASE("sender that has an exception_ptr error completion signature",
+            "[concepts][sender]")
```
ericniebler:
```suggestion
  TEST_CASE("sender that has an exception_ptr error completion signature", "[concepts][sender]")
```

ericniebler (changes requested):

test/stdexec/concepts/test_concepts_sender.cpp
maartenarnst: Ok, done!

ericniebler (approved):

maartenarnst:
@ericniebler The modules build in the CPU pipeline failed. This is why I've added the `STDEXEC_MODULE_EXPORT_AUTHORING` above the concept.

ericniebler:
Don't worry about failures in the modules build. That job is allowed to fail since clang's modules support is still pretty flakey. Module build failures will not block PRs.

ericniebler:
thanks!

PR #2271, opened 2026-09-19
fix(exec): make sequence senders work with STDEXEC_ENABLE_EXTRA_TYPE_CHECKING

### LINKED ISSUE #2111: Sequence sender compilation failure with STDEXEC_ENABLE_EXTRA_TYPE_CHECKING
##Description

`any_sequence_sender | exec::transform_each(...)` fails to compile when `STDEXEC_ENABLE_EXTRA_TYPE_CHECKING=ON`.

The issue is that the `exec::sequence` validation paths in `sequence_senders.hpp` call `completion_signatures_of_t`, which under extra type checking routes into `__debug_sender` — a path designed for regular senders, not sequence senders.

## Steps to Reproduce

1. Clone stdexec at commit `61fb73d7`
2. Enable `STDEXEC_ENABLE_EXTRA_TYPE_CHECKING=ON`
3. Compose `any_sequence_sender | transform_each(...)`

Minimal reproducer:

```cpp
#include <exec/sequence/any_sequence_of.hpp>
#include <exec/sequence/transform_each.hpp>
#include <exec/sequence/iterate.hpp>

using item_sigs = stdexec::completion_signatures<
    stdexec::set_value_t(int),
    stdexec::set_error_t(std::exception_ptr),
    stdexec::set_stopped_t()>;

using erased_seq_t = exec::any_sequence_sender<
    exec::any_sequence_receiver<item_sigs>>;

int main() {
    // Build a concrete sequence sender, erase to any_sequence_sender,
    // then compose with transform_each — this is the failing pattern.
    auto concrete = exec::iterate(std::vector{1, 2});
    erased_seq_t erased = std::move(concrete);

    auto pipe = std::move(erased)
        | exec::transform_each(stdexec::then([](int v) { return v + 1; }));
    (void)pipe;
}
```

## Root Cause

Two places in `sequence_senders.hpp` use the regular `completion_signatures_of_t` for sequence sender checks:

- `__sequence_receiver_from` concept (line ~687)
- `subscribe_t::__type_check_arguments` (line ~758)

When extra type checking is ON, these instantiate `__debug_sender` which tries `connect()` — but sequence senders are not regular senders and don't provide `connect()`.

Additionally, `transform_each_t::get_env` returns the child environment **by value**, which fails when the child is `any_sequence_sender` (immovable environment proxy). The function also has a recursive `static_assert(sender_for<...>)` that re-enters the regular sender concept.

## Proposed Fix

**File 1: `include/exec/sequence_senders.hpp`**

Use `__sequence_completion_signatures_of_t` instead of `completion_signatures_of_t` in the two sequence-specific checks:

```diff
 // __sequence_receiver_from (line ~687)
-      STDEXEC::completion_signatures_of_t<_Sequence, STDEXEC::env_of_t<_Receiver>>>;
+      __sequence_completion_signatures_of_t<_Sequence, STDEXEC::env_of_t<_Receiver>>>;

 // subscribe_t::__type_check_arguments (line ~758)
-          using __checked_signatures [[maybe_unused]] = completion_signatures_of_t<_Sequence, env_of_t<_Receiver>>;
+          using __checked_signatures [[maybe_unused]] = __sequence_completion_signatures_of_t<_Sequence, env_of_t<_Receiver>>;
```

**File 2: `include/exec/sequence/transform_each.hpp`**

Return child environment by reference and remove the recursive sender_for check:

```diff
-      static auto get_env(_Sexpr const & __sexpr) noexcept -> env_of_t<__child_of<_Sexpr>>
+      static decltype(auto) get_env(_Sexpr const & __sexpr) noexcept
       {
-        static_assert(sender_for<_Sexpr, transform_each_t>);
         return __apply([]<class _Child>(__ignore, __ignore, _Child const & __child)
-                       { return STDEXEC::get_env(__child); },
+                       -> decltype(auto) { return STDEXEC::get_env(__child); },
                        __sexpr);
       }
```

## Full unified diff

```diff
diff --git a/include/exec/sequence/transform_each.hpp b/include/exec/sequence/transform_each.hpp
--- a/include/exec/sequence/transform_each.hpp
+++ b/include/exec/sequence/transform_each.hpp
@@ -217,7 +217,6 @@ struct transform_each_t
       }

       template <class _Sexpr>
-      static auto get_env(_Sexpr const & __sexpr) noexcept -> env_of_t<__child_of<_Sexpr>>
+      static decltype(auto) get_env(_Sexpr const & __sexpr) noexcept
       {
-        static_assert(sender_for<_Sexpr, transform_each_t>);
         return __apply([]<class _Child>(__ignore, __ignore, _Child const & __child)
-                       { return STDEXEC::get_env(__child); },
+                       -> decltype(auto) { return STDEXEC::get_env(__child); },
                        __sexpr);
       }
     };
   }  // namespace __transform_each

diff --git a/include/exec/sequence_senders.hpp b/include/exec/sequence_senders.hpp
--- a/include/exec/sequence_senders.hpp
+++ b/include/exec/sequence_senders.hpp
@@ -684,7 +684,7 @@ concept __sequence_receiver_from =
     sequence_sender_in<_Sequence, STDEXEC::env_of_t<_Receiver>>
     && STDEXEC::receiver_of<
       _Receiver,
-      STDEXEC::completion_signatures_of_t<_Sequence, STDEXEC::env_of_t<_Receiver>>>;
+      __sequence_completion_signatures_of_t<_Sequence, STDEXEC::env_of_t<_Receiver>>>;

   template <class _Receiver, class _Sequence>
   concept sequence_receiver_from =
@@ -753,11 +753,11 @@ struct subscribe_t
         if constexpr (sequence_sender_in<_Sequence, env_of_t<_Receiver>>)
         {
-          // Instantiate __debug_sender via completion_signatures_of_t and
+          // Instantiate sequence completion signatures and
           // item_types_of_t to check that the actual completions and item_types
           // match the expected completions and values.
           using __checked_signatures
-            [[maybe_unused]] = completion_signatures_of_t<_Sequence, env_of_t<_Receiver>>;
+            [[maybe_unused]] = __sequence_completion_signatures_of_t<_Sequence, env_of_t<_Receiver>>;
           using __checked_item_types
             [[maybe_unused]] = item_types_of_t<_Sequence, env_of_t<_Receiver>>;
         }
         else
         {
           __diagnose_sequence_sender_concept_failure<_Sequence, env_of_t<_Receiver>>();
         }
         return true;
       }
```

## Environment

| Item | Value |
|------|-------|
| stdexec | `61fb73d7` |
| Trigger | `STDEXEC_ENABLE_EXTRA_TYPE_CHECKING=ON` |
| Platform | Windows x64, C++20 |
| Build config | `STDEXEC_MAIN_PROJECT=OFF` (subdirectory) |

## Verification

Confirmed on `stdexec@61fb73d7` with `STDEXEC_ENABLE_EXTRA_TYPE_CHECKING=ON`:

- **Without patch**: `frame_source.cc` (which uses `any_sequence_sender | transform_each(stdexec::then(...))`) fails with `error C3889: connect_t` — `__debug_sender` tries `connect()` on a `transform_each` sequence expression.
- **With patch**: full project builds successfully (`Debug`, `Release`, `ASan` presets).

## Additional Context

The `exec::sequence` extension is experimental. This is not a compiler-specific issue — the extra type checking paths conflate sequence and regular sender metadata. A regression test for `any_sequence_sender | transform_each` with extra type checking enabled would be useful.

alwaysprince05:
Fixes #2111

## Problem

With `STDEXEC_ENABLE_EXTRA_TYPE_CHECKING=ON`, composing an `exec::any_sequence_sender` with `exec::transform_each` fails to compile, e.g.:

```cpp
exec::any_sequence_sender<exec::any_sequence_receiver<item_sigs>> erased =
  exec::iterate(std::views::all(values));
auto pipe = std::move(erased) | exec::transform_each(exec::then([](int v) { return v + 1; }));
```

fails with `no matching function for call to object of type 'stdexec::connect_t'` from `__debug_sender`.

## Root cause

Sequence senders are not regular senders (no `connect()`/`start()`), but three code paths treated them as such when extra type checking is enabled:

1. **`sequence_senders.hpp`** — `__sequence_receiver_from` (line ~700) and `subscribe_t::__type_check_arguments` (line ~772) used `STDEXEC::completion_signatures_of_t` on the *sequence* sender. With extra type checking, that alias instantiates `__checked_completion_signatures` → `__debug_sender`, which tries to `connect()` the sequence sender. The fix uses the sequence-aware `__sequence_completion_signatures_of_t` instead, as suggested in the issue.

2. **`transform_each.hpp`** — `transform_each_t::get_env` returned the child environment **by value** and had a recursive `static_assert(sender_for<_Sexpr, transform_each_t>)` that re-entered the regular `sender<>` concept (which under extra type checking drives the debug machinery). The erased sequence sender's environment is immovable, so the by-value return also fails to compile. Returns `decltype(auto)` and drops the redundant assert, per the issue's proposed fix.

3. **`sequence_senders.hpp` (new find)** — the forward declaration of `__debug_sequence_sender` had its template parameters in a different order than the definition at the bottom of the header (`<class _Env = env<>, class _Sequence>` vs `<class _CvSequence, class _Env>`), making them two distinct function templates. Every odr-use under extra type checking therefore referenced a declared-but-never-defined function template, which clang diagnoses as an error when instantiated with internal-linkage argument types (MSVC/GCC don't diagnose it because the body never actually runs — it's guarded by `__mnever`). The declaration is now aligned with the definition.

## Verification

- Reproduced the failure on `main` with Apple clang 21 (`-DSTDEXEC_ENABLE_EXTRA_TYPE_CHECKING=1`): same `connect_t` error signature as reported. After the fix, the reproducer compiles in both configurations (A/B-verified with an identical file).
- New regression test in `test/exec/sequence/test_sequence_any_sender.cpp`: fails to compile on pristine `main` with the flag, compiles with this fix; `test.exec` passes in the normal configuration (363 test cases / 3418 assertions).
- Full `test.stdexec` passes in the normal configuration (623 test cases / 3535 assertions) — no regressions.
- With the flag ON, the sequence test files' compile errors drop from 20/58/66 (pristine `main`) to 4/6/16 (this branch). The remainder are separate pre-existing issues in the debug machinery outside this PR's scope, found while verifying:
  - `exec/async_scope.hpp` `__future_opstate` accesses `__future_state` members while incomplete (`test_type_async_scope.cpp`, also fails on `main`);
  - `item_types_of_t` under the flag still probes sequence *item* senders with the regular `__debug_sender`;
  - `on`/`continues_on` transformation hits `__not_a_scheduler` diagnostics under the debug environment.
  Happy to file follow-up issues for these if maintainers agree they're separate.

Formatted with clang-format; no behavioral changes with `STDEXEC_ENABLE_EXTRA_TYPE_CHECKING` off (the changed aliases/members are only instantiated under the flag, and `decltype(auto)` forwarding preserves the previous value-return semantics).


ericniebler:
> Happy to file follow-up issues for these if maintainers agree they're separate

yes please. and thank you!

ericniebler (approved):

alwaysprince05:
Done — filed the three follow-up issues, thanks for the quick review!

- #2272 — [async_scope] `__future_opstate` touches incomplete `__future_state` members under extra type checking
- #2273 — [sequence_senders] `item_types_of_t` probes sequence item senders with the regular `__debug_sender`
- #2274 — [exec] `on`/`continues_on` hits `__not_a_scheduler` diagnostics under the debug environment

PR #2277, opened 2026-09-20
update the start scheduler before connecting scheduler sender in `continues_on`

### LINKED ISSUE #2268: continues_on: hop opstate not connected with secondary environment
Recent work has introduced the `get_start_scheduler` query.

Certain algorithms internally start an operation state from the completion of the predecessor. This is the case for instance for `let_value`, which correspondingly passes a secondary environment with the start scheduler set accordingly.

The `continues_on` algorithm is also an algorithm that starts an operation state from the completion of the predecessor, in this case to make the hop to the target scheduler:

https://github.com/NVIDIA/stdexec/blob/5f94dbac91de3c4869fe695b7fe4d0ed66c0612d/include/stdexec/__detail/__continues_on.hpp#L127

https://github.com/NVIDIA/stdexec/blob/5f94dbac91de3c4869fe695b7fe4d0ed66c0612d/include/stdexec/__detail/__continues_on.hpp#L385

However, in the case of `continues_on`, such a secondary environment is not passed:

https://github.com/NVIDIA/stdexec/blob/5f94dbac91de3c4869fe695b7fe4d0ed66c0612d/include/stdexec/__detail/__continues_on.hpp#L193

Because of this issue, compile-time queries such as `get_start_scheduler` and `get_completion_scheduler` may give the wrong answer. Indeed, in the hop opstate, the schedule sender is connected to an environment that indicates that the start scheduler is the start scheduler of the outer receiver, rather than the completion scheduler of the predecessor. A striking consequence is that if we hop onto an `inline_scheduler`, the query wrongly indicates that the completion scheduler is the start scheduler of the outer receiver.

A possible solution is probably to pass a secondary environment to the hop opstate along the lines of what has been done for `let_value` and recently for `starts_on`. I.e., on the quoted line 193, we probably need to pass the attributes of the predecessor through the secondary environment pattern; with a corresponding fix where the schedule sender is connected.

Tagging @ericniebler and @Cra3z as this seems related to your recent work on the start scheduler, the completion scheduler, and starts_on.

<details>
<summary>
reproducer
</summary>

```
#define THEN_STORE_THREAD_ID(p) stdexec::then([p] { *(p) = std::this_thread::get_id(); })

consteval bool test_is_forwarding_query() {
    return stdexec::forwarding_query(stdexec::get_start_scheduler);
}
static_assert(test_is_forwarding_query());

/**
 * @test Check that the start scheduler that @c stdexec::sync_wait sets in the receiver environment
 *       is a @c run_loop scheduler on the thread on which it starts the operation state.
 */
TEST(get_start_scheduler, sync_wait) {
    std::thread::id tid;

    auto sndr = stdexec::read_env(stdexec::get_start_scheduler) | stdexec::let_value([&](auto schd) {
                    static_assert(std::same_as<decltype(schd), stdexec::run_loop::scheduler>);
                    return stdexec::schedule(schd) | THEN_STORE_THREAD_ID(&tid);
                });

    stdexec::sync_wait(std::move(sndr)); // NOLINT(performance-move-const-arg)

    ASSERT_EQ(tid, std::this_thread::get_id());
}

/**
 * @test Check that the start scheduler that @c stdexec::let_value sets in the receiver environment
 *       of the sender returned by the closure is the completion scheduler of the predecessor.
 *
 * Indeed, @c stdexec::let_value starts the successor from the completion of the predecessor.
 *
 * See also:
 * - https://github.com/NVIDIA/stdexec/blob/5f94dbac91de3c4869fe695b7fe4d0ed66c0612d/include/stdexec/__detail/__let.hpp#L180
 * - https://github.com/NVIDIA/stdexec/blob/5f94dbac91de3c4869fe695b7fe4d0ed66c0612d/include/stdexec/__detail/__schedulers.hpp#L639-L655
 */
TEST(get_start_scheduler, let_value) {
    std::thread::id pool_tid, tid;

    experimental::execution::static_thread_pool pool{1};

    auto sndr = stdexec::schedule(pool.get_scheduler()) | THEN_STORE_THREAD_ID(&pool_tid) | stdexec::let_value([&]() {
                    return stdexec::read_env(stdexec::get_start_scheduler) | stdexec::let_value([&](auto schd) {
                               static_assert(std::same_as<decltype(schd), decltype(pool.get_scheduler())>);
                               return stdexec::schedule(schd) | THEN_STORE_THREAD_ID(&tid);
                           });
                });

    stdexec::sync_wait(std::move(sndr)); // NOLINT(performance-move-const-arg)

    ASSERT_EQ(tid, pool_tid);
    ASSERT_NE(tid, std::this_thread::get_id());
}

/**
 * @test Show that @c stdexec::continues_on onto an inline scheduler behaves as expected at run time. However, the compile-time
 *       scheduler queries are not correct: they indicate a hop onto the start scheduler set in the outer receiver environment
 *       rather than continuing on the completion scheduler of the predecessor.
 *
 * Indeed, although @c stdexec::continues_on starts an operation state from the completion of the predecessor, it does not set the start scheduler
 * accordingly through a secondary environment.
 *
 * See also:
 * - https://github.com/NVIDIA/stdexec/blob/5f94dbac91de3c4869fe695b7fe4d0ed66c0612d/include/stdexec/__detail/__continues_on.hpp#L193
 * - https://github.com/NVIDIA/stdexec/blob/5f94dbac91de3c4869fe695b7fe4d0ed66c0612d/include/stdexec/__detail/__continues_on.hpp#L127
 */
TEST(get_start_scheduler, continues_on) {
    std::thread::id pool_tid, tid;

    experimental::execution::static_thread_pool pool{1};

    auto sndr = stdexec::schedule(pool.get_scheduler()) | THEN_STORE_THREAD_ID(&pool_tid)
              | stdexec::continues_on(stdexec::inline_scheduler{}) | THEN_STORE_THREAD_ID(&tid);

    static_assert(
        std::same_as<
            stdexec::__completion_scheduler_of_t<
                stdexec::set_value_t,
                decltype(sndr),
                stdexec::prop<stdexec::get_start_scheduler_t, stdexec::run_loop::scheduler>
            >,
            stdexec::run_loop::scheduler // does not match run-time behavior; should be decltype(pool.get_scheduler())
        >);

    stdexec::sync_wait(std::move(sndr)); // NOLINT(performance-move-const-arg)

    ASSERT_EQ(tid, pool_tid); // run-time behavior is as expected
    ASSERT_NE(tid, std::this_thread::get_id());
}
```
</details>

Joint analysis with Claude (Anthropic).

ericniebler:
fixes #2268


include/stdexec/__detail/__continues_on.hpp
```diff
@@ -84,24 +93,24 @@ namespace STDEXEC
       }
 
       [[nodiscard]]
-      constexpr auto get_env() const noexcept -> env_of_t<_Receiver>
+      constexpr auto get_env() const noexcept -> __env2_t const &
       {
-        return STDEXEC::get_env(__state_->__rcvr_);
+        return __state_->__env2_;
```
maartenarnst: Here, the corresponding code in `__let.hpp` returns the join of the secondary environment and the outer receiver's environment.

This issue is probably important because `__mk_secondary_env_t` returns an environment that doesn't forward to the outer receiver's environment. The first branch that calls `__mk_sch_env` does pass the outer receiver's environment, but `__sched_env` only uses its type to determine the type of the domain.

The approach with the join in `__let.hpp` may itself have an issue as well. If we end up in one of the last two branches in the implementation of `__mk_secondary_env_t`, the returned environment contains only a domain or nothing, but the join then reintroduces the start scheduler from the outer receiver's environment.

A possible solution may be to carry out the join in the implementation of `__mk_secondary_env_t` and use something like the `__hide_scheduler` to hide the start scheduler.

include/stdexec/__detail/__continues_on.hpp
maartenarnst:
```suggestion
        requires __never_sends<_SetTag, schedule_result_t<_Scheduler>, __env2_t<_Env>...>
```

test/stdexec/queries/test_get_start_scheduler.cpp
```diff
@@ -0,0 +1,131 @@
+/*
+ * Copyright (c) 2022 ETH Zurich
```
maartenarnst: This is probably an artifact from a copy paste :). If you would like to include our institute, we're at the University of Liège. It's fine for us that a usual NVIDIA template is used.

include/stdexec/__detail/__continues_on.hpp
```diff
@@ -114,6 +123,15 @@ namespace STDEXEC
...
     {
      private:
+      template <class _Env>
+      using __env2_t = __secondary_env_t<_Sender, _Env, set_value_t>;
+
+      template <class _Env>
+      constexpr auto __mk_env2(_Env&& __env) const noexcept -> __env2_t<_Env>
+      {
+        return __mk_secondary_env_t<set_value_t>()(__sndr_, static_cast<_Env&&>(__env));
+      }
```
maartenarnst, line 137: Noting that the tag is hardcoded here to `set_value_t`, whereas the  call sites are templated on `_SetTag`? Could this lead to issues e.g. when the predecessor completes with error/stop?

test/stdexec/queries/test_get_start_scheduler.cpp
```diff
@@ -0,0 +1,131 @@
...
+  //! @test Check that the start scheduler that @c ex::let_value sets in the receiver
+  //! environment of the sender returned by the closure is the completion scheduler of the
+  //! predecessor.
+  //!
+  //! Indeed, @c ex::let_value starts the successor from the completion of the
+  //! predecessor.
+  //!
+  //! See also:
+  //! - https://github.com/NVIDIA/stdexec/blob/5f94dbac91de3c4869fe695b7fe4d0ed66c0612d/include/stdexec/__detail/__let.hpp#L180
+  //! - https://github.com/NVIDIA/stdexec/blob/5f94dbac91de3c4869fe695b7fe4d0ed66c0612d/include/stdexec/__detail/__schedulers.hpp#L639-L655
```
maartenarnst:
```suggestion
```

test/stdexec/queries/test_get_start_scheduler.cpp
```diff
@@ -0,0 +1,131 @@
...
+  }
+
+  //! @test Show that @c ex::continues_on onto an inline scheduler behaves as
+  //! expected at run time. However, the compile-time scheduler queries are not correct:
+  //! they indicate a hop onto the start scheduler set in the outer receiver environment
+  //! rather than continuing on the completion scheduler of the predecessor.
+  //!
+  //! Indeed, although @c ex::continues_on starts an operation state from the
+  //! completion of the predecessor, it does not set the start scheduler accordingly
+  //! through a secondary environment.
```
maartenarnst:
```suggestion
  //! @test Check that the start scheduler that @c ex::continues_on sets in the receiver
  //! environment of the schedule sender is the completion scheduler of the predecessor.
  //!
  //! Indeed, @c ex::continues_on starts an hop operation state from the
  //! completion of the predecessor.
```

test/stdexec/queries/test_get_start_scheduler.cpp
```diff
@@ -0,0 +1,131 @@
...
+  //!
+  //! See also:
+  //! - https://github.com/NVIDIA/stdexec/blob/5f94dbac91de3c4869fe695b7fe4d0ed66c0612d/include/stdexec/__detail/__continues_on.hpp#L193
+  //! - https://github.com/NVIDIA/stdexec/blob/5f94dbac91de3c4869fe695b7fe4d0ed66c0612d/include/stdexec/__detail/__continues_on.hpp#L127
```
maartenarnst:
```suggestion
```

test/stdexec/queries/test_get_start_scheduler.cpp
```diff
@@ -0,0 +1,131 @@
...
+
+    STATIC_CHECK(std::same_as<ex::__completion_scheduler_of_t<
+                                ex::set_value_t,
+                                decltype(sndr),
+                                ex::prop<ex::get_start_scheduler_t, ex::run_loop::scheduler>>,
+                              decltype(pool.get_scheduler())>);
+
+    ex::sync_wait(std::move(sndr));
+
+    CHECK(tid == pool_tid);  // run-time behavior is as expected
```
maartenarnst:
```suggestion
    CHECK(tid == pool_tid);
```

test/stdexec/queries/test_get_start_scheduler.cpp
```diff
@@ -0,0 +1,131 @@
...
+  //! - https://github.com/NVIDIA/stdexec/blob/5f94dbac91de3c4869fe695b7fe4d0ed66c0612d/include/stdexec/__detail/__schedulers.hpp#L639-L655
+  TEST_CASE("get_start_scheduler with let_value", "[sched_queries][get_start_scheduler]")
+  {
+    std::thread::id          pool_tid, tid;
+    exec::static_thread_pool pool{1};
+
+    auto sndr = ex::schedule(pool.get_scheduler())  //
+              | ::then_store_thread_id(pool_tid)    //
+              | ex::let_value(
+                  [&]() noexcept
```
maartenarnst, line 74: Just curious why you added `noexcept`?

include/stdexec/__detail/__continues_on.hpp
ericniebler: nice catch, thanks!

PR #2280, opened 2026-09-20
Fix completion signature detection with GCC 13.1 and 13.2

### LINKED ISSUE #2078: `sender_in` fails with GCC 13.1 due to `const completion_signatures` in `__constant_completion_signatures_v`
## Summary

A sender that compiles successfully with GCC 11.4 fails to compile with GCC 13.1 when passed to `exec::start_detached`.

The failure appears to come from the `sender_in` constraint. GCC 13.1 evaluates:

```cpp
__constant_completion_signatures_v<STDEXEC::get_completion_signatures<_Sender, _Env...>()>
```

as `false`, apparently because `decltype(_Completions)` is `const completion_signatures<...>`.

GCC 11 succeeds because stdexec has a GCC < 13 workaround that removes `const`:

```cpp
std::remove_const_t<decltype(_Completions)>
```

GCC 13 takes the other branch and does not remove const.

## Reproducer
```cpp
// reproducer.cpp
using S = decltype(stdexec::just());
static_assert(stdexec::sender_in<S, stdexec::__root_env>);
int main(){}

/* my usage:
#include <exec/start_detached.hpp>

int main(){
    stdexec::inline_scheduler sch;
    auto sender = stdexec::schedule(sch) | stdexec::then([]() noexcept {  });
    exec::start_detached(std::move(sender));
}
*/
```
```cmake
cmake_minimum_required(VERSION 3.21)
project(issue_stdexec 
    VERSION 0.0.1 
    LANGUAGES CXX)

include(CPM.cmake)
CPMAddPackage(
  NAME stdexec
  GITHUB_REPOSITORY NVIDIA/stdexec
  GIT_TAG 6d7ad68
)

add_executable(reproducer reproducer.cpp)
target_link_libraries(reproducer PUBLIC STDEXEC::stdexec)
```

### Environment
Linux GCC 11.4: compiles successfully
Linux GCC 13.1: fails
C++ mode: -std=gnu++20

## Possible Cause

The issue may be related to stdexec/__detail/__sender_concepts.hpp #2027 :

```cpp
#if STDEXEC_GCC() && STDEXEC_GCC_VERSION < 1300
  template <auto _Completions>
  inline constexpr bool __constant_completion_signatures_v =
    __valid_completion_signatures<std::remove_const_t<decltype(_Completions)>>;
#else
  template <auto _Completions>
  inline constexpr bool __constant_completion_signatures_v =
    __valid_completion_signatures<decltype(_Completions)>;
#endif
```

On GCC 13.1, the `remove_const_t` workaround still seems necessary.

MaxFreedomPollard:
GCC 13.1 and 13.2 still give the class-type completion-signatures template argument a const-qualified type, so they need the existing `remove_const_t` workaround too. Extend its upper bound to GCC 13.3, where the unqualified check works.

Fixes #2078.

Validation:
- The issue's `sender_in` reproducer fails on unchanged main with GCC 13.1 and 13.2. With this change, it compiles with `-std=c++20 -Wall -Wextra -Werror` and an `inline_scheduler` / `then` / `start_detached` pipeline runs successfully.
- GCC 13.3 compiles and runs the same program both before and after the change.
- The existing `test.stdexec` suite passes with Apple Clang: 622 test cases, 3,522 assertions.


ericniebler (approved):

ericniebler:
thanks!

